PR #11, opened 2021-10-15
update codebase

kvezo:
- update wallet integration
- update data fetch module
- remove redux store data


RoseRompkxm:
do we still need `saga` merchanism? seem likes we can delete them @zurab-kvezo

kvezo:
> do we still need `saga` merchanism? seem likes we can delete them @zurab-kvezo

I agree. we can create new issue for this.

RoseRompkxm:
image
seems like latest block is forgetten to be set @zurab-kvezo

kvezo:
> image seems like latest block is forgetten to be set @zurab-kvezo

you are right. i will add it.

PR #19, opened 2021-11-15
disable some untested wallets

RoseRompkxm:
disable some wallets that are not fully tested (Ledger & SafePal)


kkirka (commented):
1. Maybe it's just me but "Coming..." looks like some text didn't fit the screen. "Coming soon" might be better.
2. There's CSS for "beta" flag, but there's no beta flag near WalletConnect :thinking_face:
Screenshot 2021-11-15 at 12 38 11

kkirka:
And ideally the cursor should not be "pointer" on "coming soon" items

kkirka (approved):

PR #21, opened 2021-11-16
status update enhancement

RoseRompkxm:
## Problem
Currently, there are a lot of duplicated status updates that cause unnecessary network requests and UI updates, making user interaction clumsy and slow, thus rising some complaints from users like
image

it's actually not code error, just the UI jank

## Solution
The data fetching logic in `useVaiUser`, `useMarkets` & `useMarketInfo` hooks are extracted and hoisted to corresponding `Context` to reduce redundant status updates.


kkirka (approved):
I have reviewed the code but I have not tested it locally. The code looks good to me, but would be nice if someone could actually test it before we deploy to prod.

PR #40, opened 2021-12-29
fix voters list bug

RoseRompkxm:


src/components/Vote/VoteOverview/VoteCard.js
```diff
@@ -220,7 +228,11 @@ function VoteCard({
               </div>
             ))}
           </div>
-          {isViewAll && addressNumber > 3 && (
+          {/* view all is clicked and there are still empty lines to fill up */}
+          {!isViewAll && remainingToLoad > 0 && (
+            <Icon className="loading-icon" type="loading" />
+          )}
+          {isViewAll && addressNumber > 3 && remainingToLoad > 0 && (
```
kkirka: Why 3 and not 4 though? I see it works correctly, just curious

src/components/Vote/VotingWallet.js
```diff
@@ -165,10 +165,7 @@ function VotingWallet({
           <div className="flex align-center just-between">
             <div className="flex align-center">
               <img src={coinImg} alt="coin" />
-              <p className="content-value">
-                {getBefore(format(balance))}
-                <span>{getAfter(format(balance))}</span>
-              </p>
+              <p className="content-value">{format(balance)}</p>
```
kkirka, line 158: Do we still need `getBefore` and `getAfter` functions to be defined then?

src/utilities/time.js
```diff
@@ -0,0 +1,43 @@
+import moment from 'moment';
+
+export const FORMAT_STRING = 'lll';
+
+export const formatTimeInSeconds = sec => {
+  return `${moment(sec * 1000).format(FORMAT_STRING)}`;
```
kkirka, line 9: Now we sometimes get Jan, 01, 1970. I propose we have sth like:

```suggestion
  if (!sec) {
    return '—';
  }
  return `${moment(sec * 1000).format(FORMAT_STRING)}`;
```

kkirka (approved):

src/components/Vote/VotingWallet.js
RoseRompkxm: no need for that, i will delete these useless functions

src/utilities/time.js
RoseRompkxm: that's good suggestion, let me handle this

src/components/Vote/VoteOverview/VoteCard.js
RoseRompkxm: actually 'addressNumber > 3' is uselss now, i will delete it

PR #43, opened 2022-01-10
feat: ens name and avatar support

sasicodes:
Added ens name and avatar support is if the user has one.

image


sasicodes:
cc @RoseRompkxm @kkirka

PR #44, opened 2022-01-12
Use VenusLens in MarketContext

kkirka:
Problem: We perform _lots_ of requests to the node when we query supply/borrow/wallet balances of each user. These requests are done for each vToken, leading to performance issues.

Solution: Replace multiple requests with a single call to VenusLens.

NB: Initially I wanted to change the data structures we use (like "user asset list") and replace them with TypeScript classes. However, the effort required to adapt the existing components is way over the threshold, so I propose we do the "rehaul" iteratively. Migrating to VenusLens should be the first step.

### DIFF
--- a/src/context/MarketContext.js
+++ b/src/context/MarketContext.js
@@ -3,6 +3,7 @@ import BigNumber from 'bignumber.js';
 import { useWeb3React } from '@web3-react/core';
 import useRefresh from '../hooks/useRefresh';
 import { fetchMarkets } from '../utilities/api';
+import { indexBy } from '../utilities/common';
 import useWeb3 from '../hooks/useWeb3';
 import { useVaiUser } from '../hooks/useVaiUser';
 import { useComptroller, useVenusLens } from '../hooks/useContract';
@@ -71,13 +72,6 @@ const MarketContextProvider = ({ children }) => {
 
   useEffect(() => {
     let isMounted = true;
-
-    const indexBy = (indexFn, arr) => {
-      return arr.reduce((result, item) => {
-        result[indexFn(item)] = item;
-        return result;
-      }, {});
-    }
 
     const getXvsBalance = balances => {
       const vxvs = constants.CONTRACT_VBEP_ADDRESS['xvs'].address.toLowerCase();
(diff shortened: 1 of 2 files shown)


src/context/MarketContext.js
```diff
@@ -76,6 +71,20 @@ const MarketContextProvider = ({ children }) => {
...
   useEffect(() => {
     let isMounted = true;
+
+    const indexBy = (indexFn, arr) => {
+      return arr.reduce((result, item) => {
+        result[indexFn(item)] = item;
+        return result;
+      }, {});
+    }
+
```
RoseRompkxm: this method can be moved to the utility file

src/context/MarketContext.js
```diff
@@ -87,178 +96,143 @@ const MarketContextProvider = ({ children }) => {
...
+              supplyBalance,
+              borrowBalance,
+              isEnabled,
+              collateral,
+              percentOfLimit
+            };
+
+            return asset;
+          }
+        )
```
RoseRompkxm, line 182: just `return { .... }` will be good

src/context/MarketContext.js
```diff
@@ -87,178 +96,143 @@ const MarketContextProvider = ({ children }) => {
...
+          async asset => {
+            const getHypotheticalLiquidity = () => {
+              return comptrollerContract.methods
+                .getHypotheticalAccountLiquidity(
+                  account,
+                  asset.vtokenAddress,
+                  balances[asset.vtokenAddress.toLowerCase()].balanceOf,
+                  0
+                )
+                .call()
```
RoseRompkxm, line 198: let's create this batch liquidity query function in our next iteration :)

RoseRompkxm (approved):

src/context/MarketContext.js
kkirka: b7cd4500dccd10935ae75ecea28990aff8b15d83

src/context/MarketContext.js
kkirka: b7cd4500dccd10935ae75ecea28990aff8b15d83

src/context/MarketContext.js
kkirka: https://app.clickup.com/t/22bbtv6

RoseRompkxm (approved):

PR #50, opened 2022-02-02
docs: update readme, fix .env.template

baomastr:


README.md
```diff
@@ -8,7 +8,9 @@ npm run start
 
 ## Requirements and Configuration
 
-You’ll need to have Node 8.16.0 or Node 10.16.0 or later version on your local development machine
+You’ll need to have Node v14.19.0 on your local development machine
```
coreyar: We can set this as the engine in package.json so it will be checked automatically
https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines

kkirka: Do you guys think it makes sense to add `.nvmrc` as well?

coreyar: The docker file says node 12.19.0 so we should update this to be greater then 12

baomastr: gonna open new PR for node version updates. 
In current PR would be just package fixes

package.json
```diff
@@ -20,7 +20,7 @@
     "lodash": "^4.17.21",
     "markdown-it": "^12.0.0",
     "moment": "^2.24.0",
-    "node-sass": "^4.12.0",
+    "node-sass": "^4.14.0",
```
coreyar, line 23: node sass is [deprecated](https://sass-lang.com/blog/libsass-is-deprecated) so we should switch to dart sass

[Docs](https://create-react-app.dev/docs/adding-a-sass-stylesheet/)

PR #52, opened 2022-02-02
Remove sass

baomastr:
- Remove sass, used postcss and css modules instead
- fix css paths
- fix baseUrl


src/containers/App.js
```diff
@@ -21,7 +21,6 @@ import VoterLeaderboard from 'containers/Main/VoterLeaderboard';
 import Transaction from 'containers/Main/Transaction';
 import Theme from './Theme';
 
-import 'assets/styles/App.scss';
```
baomastr, line 24: it is already imported in index.css

jsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+  "compilerOptions": {
+    "baseUrl": "./src"
```
baomastr, line 3: fix for absolute paths in project.
NODE_PATH can be removed

PR #55, opened 2022-02-03
Migrate project to typescript

baomastr:


.env.template
```diff
@@ -1,2 +1 @@
-NODE_PATH=
-REACT_APP_CHAIN_ID=
\ No newline at end of file
+REACT_APP_CHAIN_ID=
```
coreyar: I dont think we need to change this file in this PR. I am also less inclined to remove NODE_PATH because it is used in the docker containers

Dockerfile
```diff
@@ -1,7 +1,6 @@
 FROM node:12.19.0-alpine3.12 as builder
 
 ENV NODE_ENV development
-ENV NODE_PATH=src/
```
coreyar: This shouldn't be part of this PR.

Dockerfile.local
```diff
@@ -2,10 +2,9 @@ FROM node:12.19.0-alpine3.12
 
 # must use development so that "npm run build" works
 ENV NODE_ENV development
-ENV NODE_PATH=src/
```
coreyar: This shouldn't be part of this PR.

Dockerfile
coreyar: I'm not clear on the purpose of setting this ENV_VAR. It is probably worth while checking in with devops to see why it was set and what it is used for

package.json
```diff
@@ -65,7 +66,9 @@
     "prettier": "^1.18.2",
     "stylelint": "^10.1.0",
     "stylelint-config-standard": "^18.3.0",
-    "stylelint-scss": "^3.9.2"
+    "stylelint-scss": "^3.9.2",
+    "ts-migrate": "^0.1.27",
```
coreyar: We can remove this as a dev dependency

src/components/Basic/BorrowModal.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from 'react';
+// @ts-expect-error ts-migrate(7016) FIXME: Could not find a declaration file for module 'styl... Remove this comment to see the full error message
```
coreyar, line 2: `yarn add -D @types/styled-components`

kkirka:
@baomastr FYI you can use the following technique to drop the "[fix absolute paths with jsConfig](https://github.com/VenusProtocol/venus-protocol-interface/pull/55/commits/afb322f0fa9b216a329aea22d552e16c373c44fe)" commit:

`git rebase -i 175f15d`

```
drop afb322f fix absolute paths with jsConfig
pick fa9dfc0 update config
pick b946086 Add tsFixMe aliases
pick 4530f0e Add tsConfig
pick 0b0d034 ts-migrate rename .
pick b11c133 ts-migrate migrate --aliases tsfixme .
pick a22bb1f fix last of errors manually
pick dd1184b fix missing declaration
pick 75271c9 set allowJs to false
```
(note the "drop" in the first line)

`git push --force-with-lease`

coreyar (approved):

PR #56, opened 2022-02-05
Add github action typecheck

coreyar:
Add an action to check types in CI using github actions

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "lint": "export NODE_OPTIONS=\"--max-old-space-size=4096\" && eslint .",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -8,7 +8,8 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "lint": "export NODE_OPTIONS=\"--max-old-space-size=4096\" && eslint .",
-    "lint:fix": "eslint . --fix"
+    "lint:fix": "eslint . --fix",
+    "typecheck": "tsc"
```
baomastr, line 12: How do you think, should we also add type check on prepush hook?

coreyar: Yes can add a prepush hook using husky, I can do that in a separate PR

baomastr (approved):

PR #59, opened 2022-02-06
Upgrade to react scripts 5

coreyar:
Problem:

Upgrading to react-scripts v4 introduced this [bug](https://github.com/facebook/create-react-app/issues/11773)
Pinning / force resolving the version of `react-error-overlay` didn't resolve. the issue

Solution:
Upgrade to react-scripts 5 which doesnt have this bug


baomastr:
there is tmp fix for overlay hiding. just get the iframe and hide/remove from DOM.
upgrade to react-scripts 5 may cause lot of other bugs

coreyar:
Can you provide more info on the other bugs? So far I have not encountered that on other projects

coreyar:
I saw an issue open on building and I'm getting a an issue resolving the eslint plugin when I try to build. I'll move this to draft for now in favor of hiding the iframe.

PR #67, opened 2022-02-07
Remove rule @typescript-eslint/naming-convention

baomastr:


src/utilities/contractHelpers.ts
```diff
@@ -27,7 +27,7 @@ import {
 import * as constants from './constants';
 
 const getContract = (abi: $TSFixMe, address: $TSFixMe, web3: $TSFixMe) => {
-  // eslint-disable-next-line no-underscore-dangle
+  // eslint-disable-next-line @typescript-eslint/naming-convention
```
coreyar: Rather then keep the issue ignored we can rename the variable

coreyar (approved):

PR #70, opened 2022-02-07
Add git hooks with husky and github action for linting

coreyar:
- Rationalizes prettier and eslint
- Adds precommit hooks for linting and formatting
- Adds github action for linting and formatting
- Runs updated config on all files


baomastr (approved):

package.json
```diff
@@ -7,8 +7,9 @@
     "build": "react-scripts --max_old_space_size=4096 build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
+    "pretty": "yarn prettier src/**/** public/**/*.html --check",
     "lint": "export NODE_OPTIONS=\"--max-old-space-size=4096\" && eslint .",
-    "lint:fix": "eslint . --fix",
+    "lint:fix": "yarn lint --fix && yarn pretty --write",
```
baomastr, line 12: what do you think about this method for local development?
https://gist.github.com/estorgio/e8bcaa8e87d0fcdcf85fdf598956e34c

coreyar: I am not a fan of automatically adding files. A developer may have edited files that don't belong in the commit. Is there anything else you see we are missing from that setup?

baomastr: this is needed to automatically comb the code in the files that were modified according to our codestyle

coreyar: The precommit hook currently doesn't modify files, it only checks that they pass linting/ formatting and aborts the commit if there is an error

PR #81, opened 2022-02-08
Fix rule @typescript-eslint/no-var-requires

baomastr:


baomastr:
this pr also has cherry-picked commit https://github.com/VenusProtocol/venus-protocol-interface/pull/77/commits/38baca0e9862deb179e396e3ec8a0c0e0d0d664f from https://github.com/VenusProtocol/venus-protocol-interface/pull/77

PR #86, opened 2022-02-08
Add types and remove comment for animated-number-react

coreyar:
Add custom type for animated-number-react and remove comment
[component source](https://github.com/Leocardoso94/animated-number-react/blob/master/src/lib/components/Animated-Number.js)


baomastr (approved):

src/components/Dashboard/animated-number-react.d.ts
```diff
@@ -0,0 +1,20 @@
+/// <reference types="react" />
```
baomastr, line 1: Should file be in directory types?

PR #91, opened 2022-02-09
Fix missing default import from ethers

coreyar:
Ethers doesn't have a default export

### DIFF
--- a/src/utilities/common.ts
+++ b/src/utilities/common.ts
@@ -1,7 +1,7 @@
 import * as constants from 'utilities/constants';
 import BigNumber from 'bignumber.js';
 import { getVaiTokenAddress } from './addressHelpers';
-import ethers from 'ethers';
+import { ethers } from 'ethers';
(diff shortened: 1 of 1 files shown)


src/utilities/common.ts
```diff
@@ -1,7 +1,7 @@
 import * as constants from 'utilities/constants';
 import BigNumber from 'bignumber.js';
 import { getVaiTokenAddress } from './addressHelpers';
-import ethers from 'ethers';
+import * as ethers from 'ethers';
```
baomastr: import {ethers} from 'ethers';

baomastr: it was already fixed earlier

coreyar: These are equivalent for this package but I am happy to use a named import

baomastr:
resolved in https://github.com/VenusProtocol/venus-protocol-interface/pull/94

PR #95, opened 2022-02-10
Add VRT conversion UI

RoseRompkxm:
VRT conversion enables users to swap their VRT tokens in a vesting way

Due to the possible chances of vesting calculation model, this feature will be overdue for some time, in order to avoid possible troublesome and tedious conflicts resolving brought by rebasing the master branch which will be continuously refactored, here we merge this feature into main branch and disable the entry for it


src/components/Basic/TabContainer.tsx
```diff
@@ -0,0 +1,80 @@
...
+    <TabContainerWrapper style={styles}>
+      <Tabs
+        activeKey={`${activeTabIndex}`}
+        renderTabBar={props =>
+        (
+          <div className="tab-header">
+            {titles.map((title, i) =>
+              (
+                <div
+                  key={i}
```
coreyar: We should avoid using indexes as keys because if there is another component doing this on the same page we'll have conflicting keys. I would use the index and the title which should make it unique to the page

src/containers/Main/VrtConversion.tsx
```diff
@@ -0,0 +1,208 @@
...
+                      }
+                    }}
+                    account={account || ''}
+                  />
+                  <Withdraw
+                    withdrawableAmount={new BigNumber(0)}
+                    handleClickWithdraw={() => {
+                      xvsVestingContract.methods.withdraw(account);
+                    }}
+                    account={account || ''}
```
coreyar, line 185: could we do something like `const { account = '' } = useWeb3React();` to avoid this operator?

Another option would be for useWeb3React to return the default

src/containers/Main/VrtConversion.tsx
```diff
@@ -0,0 +1,208 @@
...
+            <Col
+              xl={{ span: 8, offset: 8 }}
+              lg={{ span: 12, offset: 6 }}
+              md={{ span: 12, offset: 6 }}
+              sm={{ span: 24 }}
+              xs={{ span: 24 }}
+            >
+              <div className="container">
+                <TabContainer
+                  styles={{
```
coreyar: I think as a policy we should avoid inline styles because they can't be overridden

src/components/Basic/TabContainer.tsx
```diff
@@ -0,0 +1,80 @@
...
+  styles?: React.CSSProperties;
+  onChange?: (tabIndex: string) => void;
+  children: React.ReactElement[];
+  titles: string[];
+};
+
+export default ({ styles, onChange, children, titles }: TabContainerPropsType) => {
+  const [activeTabIndex, setActiveTabIndex] = useState(0);
+  return (
+    <TabContainerWrapper style={styles}>
```
coreyar: I think we should have a policy to avoid inline styles

src/containers/Main/VrtConversion.tsx
```diff
@@ -0,0 +1,208 @@
+/* eslint-disable no-useless-escape */
```
coreyar: I would either update the regex or ignore this inline

src/containers/Main/VrtConversion.tsx
```diff
@@ -0,0 +1,208 @@
...
+);
+const CONVERSION_RATIO_DECIMAL = new BigNumber(10).pow(18);
+
+export default function VrtConversion() {
+  // contract data
+  const [redeemableAmount, setRedeemableAmount] = useState(new BigNumber(0));
+  const [dailyUtilisation, setDailyUtilisation] = useState(new BigNumber(0));
+  const [vrtDailyLimit, setVrtDailyLimit] = useState(new BigNumber(0));
+  // eslint-disable-next-line no-unused-vars
+  const [numberOfDaysSinceStart, setMumberOfDaysSinceStart] = useState(
```
coreyar: It looks like we are setting this but not using it. Do we need to set it then?

src/components/Basic/TabContainer.tsx
RoseRompkxm: makes sense

src/containers/Main/VrtConversion.tsx
RoseRompkxm: yes, useWeb3React is used directly from an external library, we'd better add a wrapper to it, but this time I'd like to adopt option one and leave the second for later refactoring

src/containers/Main/VrtConversion.tsx
RoseRompkxm: what's the proposed way for users to add customized styles though? I think users already explicitly set the styles so they should have the highest priority

src/containers/Main/VrtConversion.tsx
RoseRompkxm: will fix

src/containers/Main/VrtConversion.tsx
coreyar: The preferred way would be to use a class name. This provides a couple advantages it keeps all styles. It keeps all styles organized together and it allows for the styles to be overridden easier.

For example, we might find ourselves in a situation where the parent wants to customize the styles of a child dynamically but if that style is passed in with the style attribute we won't be able to do that.

src/components/Basic/TabContainer.tsx
coreyar: @baomastr has done this with `react-uid` 
https://github.com/VenusProtocol/venus-protocol-interface/pull/96

src/containers/Main/VrtConversion.tsx
RoseRompkxm: do you mean something like `<TabContainer className={props.className} />` ?

RoseRompkxm: or we can just get rid of style customization here, because if we want to design more universal components, this should be more well designed

baomastr: event something like this: `<TabContainer className={classNames(styles.tabContainer, props.className)} />`

src/components/Basic/TabContainer.tsx
```diff
@@ -0,0 +1,83 @@
...
+          </div>
+        )
+        }
+        animated={false}
+        defaultActiveKey="0"
+        onChange={onChange}
+      >
+        {children.map((child, i) =>
+          (
+            <TabPane key={`${i}`} tab={titles[i]}>
```
baomastr: ```<TabPane key={`${uid(child)}`} tab={titles[i]}>```

PR #103, opened 2022-02-11
Set up precommit hook with lint-staged

baomastr:


baomastr:
fixed in https://github.com/VenusProtocol/venus-protocol-interface/pull/104

PR #108, opened 2022-02-14
fix eslint config

baomastr:


baomastr:
I think we don't eslint:recommended because main rules are handling by airbnb config

src/containers/App.tsx
```diff
@@ -17,11 +17,9 @@ import MarketDetail from 'containers/Main/MarketDetail';
 import VoteOverview from 'containers/Main/VoteOverview';
 import ProposerDetail from 'containers/Main/ProposerDetail';
 import VoterLeaderboard from 'containers/Main/VoterLeaderboard';
-import VrtConversion from 'containers/Main/VrtConversion';
 import Transaction from 'containers/Main/Transaction';
 import Theme from './Theme';
 
-import 'assets/styles/App.scss';
```
coreyar, line 24: This scss file is being used so we need to keep it

baomastr: but it's already imported in index.scss

coreyar: I see that now, imported in `src/index.tsx`

PR #117, opened 2022-02-21
feat(components): Add layout components

baomastr:


src/components/v2/Layout/Layout.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { ReactNode } from 'react';
```
coreyar, line 1: I think it is logical to call this file the index so we don't duplicate the directory name for the base file.

coreyar: We should add a storybook of this component as well.

baomastr: added index.ts for import without folder name

src/components/v2/Layout/Layout.tsx
```diff
@@ -0,0 +1,19 @@
...
+import Box from '@mui/material/Box';
+import { Sidebar } from './Sidebar';
+import { Header } from './Header';
+import { PageContainer } from './PageContainer';
+
+/* sidebar width */
+const drawerWidth = 180;
```
coreyar: Should this be defined as a theme constant and imported here?

src/components/v2/Layout/Sidebar.tsx
```diff
@@ -0,0 +1,105 @@
...
+    text: 'Dev',
+    icon: <XVSIcon width={20} height={20} />,
+  },
+];
+
+interface IProps {
+  drawerWidth: number;
+}
+
+export const Sidebar = ({ drawerWidth }: IProps) => {
```
coreyar: If the drawerWidth is part of the theme we don't need to pass it as a prop

src/components/v2/Layout/Layout.tsx
baomastr: this variable is used just in scope of layout, but if you think that we should define it inside theme, i would move it

src/components/v2/Layout/Sidebar.tsx
```diff
@@ -0,0 +1,105 @@
...
+  },
+  {
+    href: '/market',
+    text: 'Market',
+    icon: <XVSIcon width={20} height={20} />,
+  },
+  {
+    href: '/vault',
+    text: 'Vault',
+    icon: <XVSIcon width={20} height={20} />,
```
coreyar: We probably want to have these values be a constant in the theme so we have consistent icons across the app

src/components/v2/Layout/Sidebar.tsx
baomastr: okay. gonna define it inside theme scope

src/containers/App.tsx
```diff
@@ -75,6 +77,7 @@ class App extends React.Component {
                         {process.env.REACT_APP_CHAIN_ID === '97' && (
                           <Route exact path="/faucet" component={Faucet} />
                         )}
+                        {isDevEnvironment && <Route exact path="/dev" component={Dev} />}
```
coreyar: What is this page for?

baomastr: just for develop debugging

src/components/v2/Layout/Sidebar.tsx
baomastr: seems like this values may be constants just in scope of sidebar.
in other parts of app may be other sizes for icons

baomastr:
stories for layout components are ready @coreyar

src/containers/App.tsx
coreyar: Let's remove this in favor of using storybooks. Storybooks are perfect for this purpose

coreyar: And as a side effect we also use them for snapshotting so if we develop with them then there is no extra work to setup the snapshot

coreyar:
> stories for layout components are ready @coreyar

It looks like the responsiveness of the [Dashboard ](https://tf-venus-prod-vizreg.s3.amazonaws.com/fe58ede84b0eedb881695d3310d3eeef0212e970/index.html?id=changed-chrome_iphone7_Dashboard_Dashboard.png)changed. The main thing I'd call out is the capitalization changed on the connect button.

baomastr:
> > stories for layout components are ready @coreyar
> 
> It looks like the responsiveness of the [Dashboard ](https://tf-venus-prod-vizreg.s3.amazonaws.com/fe58ede84b0eedb881695d3310d3eeef0212e970/index.html?id=changed-chrome_iphone7_Dashboard_Dashboard.png)changed. The main thing I'd call out is the capitalization changed on the connect button.

yes, we don't need capitalization in buttons, like on current production version

baomastr:
I also set up the `layout: 'fullscreen'` in preview.js in case of paddings in body for storybooks preview.

src/containers/App.tsx
baomastr: removed

coreyar (approved):

PR #124, opened 2022-02-22
Add google analytics client

coreyar:
This PR adds a client for google analytics.

The client will send events if `gtag` is loaded otherwise it will log an error. It creates a closure around the setup and only exposes to call specific events.

I chose to create a `client` directory for this as a way to organize clients that will interact with external sources like apis, subgraph, ect.

### DIFF
--- a/src/clients/googleAnalytics.spec.ts
+++ b/src/clients/googleAnalytics.spec.ts
@@ -1,0 +1,30 @@
+import googleAnalytics from './googleAnalytics';
+
+describe('Google Analytics Client', () => {
+  let consoleSpy: jest.SpyInstance;
+  beforeAll(() => {
+    consoleSpy = jest.spyOn(console, 'warn');
+  });
+
+  afterAll(() => {
+    consoleSpy.mockRestore();
+  });
+
+  test('event is called when gtag exists on window', () => {
+    window.gtag = () => {};
+    const spy = jest.spyOn(window, 'gtag');
+    googleAnalytics.buttonPressed('connect');
+    expect(spy).toHaveBeenCalled();
+  });
+
+  test('calling event does not crash when gtag does not exit window', () => {
+    // @ts-expect-error gtag could be missing if script isn't loaded from document
+    window.gtag = undefined;
+    googleAnalytics.buttonPressed('connect');
+    expect(consoleSpy).toHaveBeenCalled();
+    expect(consoleSpy.mock.calls[0][1]).toStrictEqual([
+      'button_pressed',
+      { button_name: 'connect' },
+    ]);
+  });
+});
--- a/src/clients/googleAnalytics.ts
+++ b/src/clients/googleAnalytics.ts
@@ -1,0 +1,27 @@
+type EventParameters = Parameters<(eventName: string, parameters?: Record<string, string>) => void>;
+
+const googleAnalytics = () => {
+  let gtag: Gtag.Gtag;
+  const loadGTag = (callback: (...args: EventParameters) => void) => (...args: EventParameters) => {
+    gtag = window?.gtag;
+    if (!gtag && process.env.NODE_ENV !== 'production') {
+      console.warn('gtag not loaded. Google Analytics event not sent', args);
+    } else {
+      callback(...args);
+    }
+  };
+
+  const sendEvent = loadGTag((eventName: string, parameters?: Record<string, string>) => {
+    gtag('event', eventName, parameters);
+  });
+
+  const buttonPressed = (buttonName: string, parameters: Record<string, string> = {}) => {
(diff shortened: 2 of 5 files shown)


baomastr (approved):

src/clients/googleAnalytics.ts
```diff
@@ -0,0 +1,27 @@
+type EventParameters = Parameters<(eventName: string, parameters?: Record<string, string>) => void>;
+
+const googleAnalytics = () => {
+  let gtag: Gtag.Gtag;
+  const loadGTag = (callback: (...args: EventParameters) => void) => (...args: EventParameters) => {
+    gtag = window?.gtag;
+    if (!gtag && process.env.NODE_ENV !== 'production') {
+      console.warn('gtag not loaded. Google Analytics event not sent', args);
```
therealemjy: Wouldn't that crash in `production` then? As if `window.gtag` doesn't exist but the env is `production`, then this condition fails and the callback is called anyway.

src/clients/googleAnalytics.ts
```diff
@@ -0,0 +1,27 @@
...
+    gtag('event', eventName, parameters);
+  });
+
+  const buttonPressed = (buttonName: string, parameters: Record<string, string> = {}) => {
+    sendEvent('button_pressed', { button_name: buttonName, ...parameters });
+  };
+
+  return {
+    buttonPressed,
+  };
```
therealemjy, line 36: That looks very good, only exposing specific events is great in my opinion.

As the project grows and we add more events, we will have more and more functions that look like `buttonPressed`, some of which will share the exact same type definition apart from the name of the event being used.
In that case, we could instead expose the `sendEvent` function directly, but use the type definition to restrict its usage to only the events we want to use:

```typescript
// Define the types of each event and its parameters
export type AnalyticEventMap = {
  button_pressed: {
    button_name: string;
  };
};

// Extract the event name from the map
export type AnalyticEventName = keyof AnalyticEventMap;

const googleAnalytics = {
  // ...

  sendEvent: function<E extends AnalyticEventName>(
    eventName: E,
    parameters: AnalyticEventMap[E] extends undefined
      ? never
      : AnalyticEventMap[E]
  ) {
    gtag('event', eventName, parameters);
  }
};

export default googleAnalytics;
```

What do you think?

src/clients/googleAnalytics.ts
coreyar: yup updating

src/clients/googleAnalytics.ts
coreyar: Yeah, I like it

src/clients/googleAnalytics.ts
```diff
@@ -0,0 +1,29 @@
...
+  const loadGTag = (callback: (...args: EventParameters) => void) => (...args: EventParameters) => {
+    gtag = window?.gtag;
+    if (!gtag) {
+      if (process.env.NODE_ENV !== 'production') {
+        console.warn('gtag not loaded. Google Analytics event not sent', args);
+      }
+    } else {
+      callback(...args);
+    }
+  };
```
therealemjy, line 21: You can avoid nesting IF statements:
```typescript
if (gtag) {
  callback(...args);
} else if (process.env.NODE_ENV !== 'production') {
  console.warn('gtag not loaded. Google Analytics event not sent', args);
}
```

coreyar: I don't want to do that because I want to check if the tag exists or I want to call the callback. I only want to log errors in non-production environments. We could decide we want to handle the non existence of gtag another way but I though  would add this as a minimum

therealemjy: That's what this code does: it only logs in non-production environments, if `gtag` doesn't exist (otherwise it will execute the callback).

It's not a major thing, so if you prefer your syntax I'm happy to approve the PR. I thought I'd just highlight you can avoid the nesting 🙂

coreyar: you are right I misread it 🙃

therealemjy: Happens!

src/clients/googleAnalytics.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  test('event is called when gtag exists on window', () => {
+    window.gtag = () => {};
+    const spy = jest.spyOn(window, 'gtag');
+    googleAnalytics.buttonPressed('connect');
+    expect(spy).toHaveBeenCalled();
+  });
+
+  test('event calling event does not crash when gtag doesnot exit window', () => {
```
therealemjy: Small typo:
```
test('calling event does not crash when gtag does not exist on window', () => {
```

therealemjy (approved):
Nice!

PR #128, opened 2022-02-23
Add Icon component

therealemjy:
Preview:
Screenshot 2022-02-23 at 18 44 20

I only included the icons from the designs that use the SVG format, which means coin logos are excluded. For these, I'd say we can add them as PNGs inside the `src/assets/img/coins` folder (as it currently is in the project).


src/components/v2/Icon/icons/fee.svg
```diff
@@ -0,0 +1,4 @@
...
... 10.5653 11.088 11.004C10.6493 11.4333 10.0753 11.648 9.366 11.648ZM9.506 16.478H8.456L14.42 7.14H15.484L9.506 16.478ZM16.212 15.988C15.7733 16.4267 15.204 16.646 14.504 16.646C13.804 16.646 13.2347 16.4267 12.796 15.988C12.3573 15.5493 12.138 14.9987 12.138 14.336C12.138 13.664 12.3573 13.104 12.796 12.656C13.2347 12.208 13.804 11.984 14.504 11.984C15.204 11.984 15.7733 12.208 16.212 12.656C16.66 13.104 16.884 13.664 16.884 14.336C16.884 14.9987 16.66 15.5493 16.212 15.988ZM9.366 10.556C9.69267 10.556 9.95867 10.4487 10.164 10.234C10.3693 10.01 10.472 9.716 10.472 9.352C10.472 8.96933 10.3693 8.666 10.164 8.442C9.95867 8.218 9.69267 8.106 9.366 8.106C9.03933 8.106 8.77333 8.218 8.568 8.442C8.36267 8.666 8.26 8.96933 8.26 9.352C8.26 9.70667 8.36267 9.996 8.568 10.22C8.78267 10.444 9.04867 10.556 9.366 10.556ZM13.706 15.232C13.9113 15.4467 14.1773 15.554 14.504 15.554C14.8307 15.554 15.0967 15.4467 15.302 15.232C15.5167 15.008 15.624 14.7093 15.624 14.336C15.624 13.9533 15.5167 13.65 15.302 13.426C15.0967 13.202 14.8307 13.09 14.504 13.09C14.1773 13.09 13.9113 13.202 13.706 13.426C13.5007 13.65 13.398 13.9533 13.398 14.336C13.398 14.7093 13.5007 15.008 13.706 15.232Z" fill="white"/>
```
baomastr, line 3: shouldn't we use currentColor here as well?

src/components/v2/Icon/icons/sort.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="12" height="12" viewBox="0 0 12 12" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M2 7.71411L6.36358 11.9998L10.7272 7.71411H2Z" fill="white"/>
```
baomastr, line 2: and in all other icons

src/components/v2/Icon/index.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <>
+      {svgFileNames.map(svgFileName => (
+        <Icon name={svgFileName} key={svgFileName} />
+      ))}
+    </>
+  );
+};
+
+export const IconWithCustomColorAndSize = () => <Icon name="mask" size={32} color="#345345" />;
```
baomastr: it would be nice to add addon for selecting icons in storybook controls

coreyar (approved):

src/components/v2/Icon/icons/fee.svg
therealemjy: In this instance we can't, otherwise the "%" would be blended into the background:
Screenshot 2022-02-23 at 21 02 38

src/components/v2/Icon/icons/sort.svg
therealemjy: In this case, I figured that since we can't change the colour of the highlighted caret (on the `sortUp` and `sortDown` icons), then we might have to use these icons as is.
What do you think?

src/components/v2/Icon/index.stories.tsx
therealemjy: Sounds good to me, I'll add that in.

therealemjy: Done :)

baomastr (approved):

src/components/v2/Icon/icons/sort.svg
baomastr: yep. I think we can edit if it would be needed in future

PR #129, opened 2022-02-23
Add wallet connect event

coreyar:
This PR adds two events for connecting a wallet 
- When the connect wallet button is pressed and sends the connection type as a parameter
- When the connection errors, with the connector and failure reason sent as parameters

To make typing easier, a commit is included to rewrite the `ConnectorNames` as an enum and updates for related types which allowed for the removal of a couple type suppression comments

### DIFF
--- a/src/clients/googleAnalytics.ts
+++ b/src/clients/googleAnalytics.ts
@@ -1,10 +1,12 @@
+import { ConnectorNames } from 'utilities/connectors';
+
 type EventParameters = Parameters<(eventName: string, parameters?: Record<string, string>) => void>;
 
-export type ButtonEventMap = {
-  connect: Record<string, string>;
+export type EventMap = {
+  connect_wallet: { type: ConnectorNames; code?: string };
 };
 
-export type ButtonEventName = keyof ButtonEventMap;
+export type EventName = keyof EventMap;
 
 const googleAnalytics = () => {
   let gtag: Gtag.Gtag;
@@ -24,15 +26,23 @@ const googleAnalytics = () => {
     gtag('event', eventName, parameters);
   });
 
-  function buttonPressed<E extends ButtonEventName>(
+  function buttonPressed<E extends EventName>(
     buttonName: E,
-    parameters?: ButtonEventMap[E] extends undefined ? never : ButtonEventMap[E],
+    parameters?: EventMap[E] extends undefined ? never : EventMap[E],
   ): void {
     sendEvent('button_pressed', { button_name: buttonName, ...parameters });
   }
 
+  function error<E extends EventName>(
+    errorName: E,
+    parameters?: EventMap[E] extends undefined ? never : EventMap[E],
+  ): void {
+    sendEvent('error', { error_name: errorName, ...parameters });
+  }
+
   return {
     buttonPressed,
+    error,
   };
 };
 
--- a/src/utilities/connectors.ts
+++ b/src/utilities/connectors.ts
@@ -25,12 +25,12 @@ const ledger = new LedgerConnector({
   pollingInterval: POLLING_INTERVAL,
 });
 
-export const ConnectorNames = {
-  Injected: 'MetaMask',
-  WalletConnect: 'WalletConnect',
-  BSC: 'BSC',
-  Ledger: 'Ledger',
-};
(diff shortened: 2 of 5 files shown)


src/clients/googleAnalytics.ts
```diff
@@ -1,10 +1,12 @@
+import { ConnectorNames } from 'utilities/connectors';
+
 type EventParameters = Parameters<(eventName: string, parameters?: Record<string, string>) => void>;
 
-export type ButtonEventMap = {
-  connect: Record<string, string>;
+export type EventMap = {
+  connect_wallet: { type: ConnectorNames; code?: string };
```
therealemjy, line 6: Nice 👍

src/utilities/connectors.ts
```diff
@@ -25,12 +25,12 @@ const ledger = new LedgerConnector({
...
 });
 
-export const ConnectorNames = {
-  Injected: 'MetaMask',
-  WalletConnect: 'WalletConnect',
-  BSC: 'BSC',
-  Ledger: 'Ledger',
-};
+export enum ConnectorNames {
+  Injected = 'MetaMask',
```
therealemjy: I guess to be consistent with the rest of the enum we should update this to `MetaMask = 'MetaMask'`

src/clients/googleAnalytics.ts
```diff
@@ -24,15 +26,23 @@ const googleAnalytics = () => {
...
   ): void {
     sendEvent('button_pressed', { button_name: buttonName, ...parameters });
   }
 
+  function error<E extends EventName>(
+    errorName: E,
+    parameters?: EventMap[E] extends undefined ? never : EventMap[E],
+  ): void {
+    sendEvent('error', { error_name: errorName, ...parameters });
+  }
```
therealemjy, line 41: Won't we use something like Sentry to track errors?

coreyar: Yes, but I wasn't thinking about that in this way. Perhaps there is a better name for the information being captured here. In the `activate` function error callback they are not application errors that need to be solved. They are connection errors that could be the result of a user action (rejecting the metamask) or a UI flow (we tried to take an action before authorization). 

The events  sent mirror toasts being shown to the user so for the this reason I thought it would be nice to capture analytis on them.

I added a ticket to configure sentry.

src/utilities/connectors.ts
coreyar: Feels a bit like expanding the scope of the PR but I have no problem with that change.

src/clients/googleAnalytics.ts
therealemjy: Ah I see! That makes sense.
If we could detect that an error comes from the user rejecting the connection via MetaMask, then we could send the analytic event only in that case (and its name could be something like `metamask_user_rejection`). If the error comes from the UI flow, then I say it should be reported in Sentry as it's something we have control over and should fix ourselves.

Sounds good about the ticket, I'll pick it up then.

src/utilities/connectors.ts
therealemjy: Ok fair enough, in this case I'm cool with leaving this as it is and making the update later.

src/clients/googleAnalytics.ts
coreyar: RE: Metamask rejection that is getting sent as you described

RE: Other errors, I think we should have a discussion about this. Sentry is great for programming errors but reporting non programming errors (errors shown to the user as toasts or input validation errors for example) gets complicated because the solution will likely involve other disciplines. Product and Marketing don't need to look at Sentry and will have heavier influence in raising the issue and proposing a solution so I think it makes sense to surface this data to them.

src/utilities/connectors.ts
coreyar: I thought I pushed it actually

src/clients/googleAnalytics.ts
therealemjy: Ok I see your point, it's true Sentry should only be used for errors WE can fix basically (other than changing the UX if too many users are getting frustrated).
Let's go with your current solution then, it makes sense!

therealemjy (approved):

PR #135, opened 2022-02-24
Add vote events

coreyar:
Adds events for 
- Create Proposal
- Delegating Voting Powerso th
- Collect Venus & VAI

Locking tokens was requested in the ticket but excluded because it will be included in the staking event

Closes t/23e9q98

### DIFF
--- a/src/clients/googleAnalytics.ts
+++ b/src/clients/googleAnalytics.ts
@@ -7,6 +7,14 @@ export type EventMap = {
   mint_vai: { amount: string };
   repay_vai: { amount: string };
   collateral_toggle: { status: 'enter' | 'exit' | 'in_repayment'; market: string };
+  create_proposal: {
+    status:
+      | 'show_modal'
+      | 'proposal_in_progress'
+      | 'submit'
+      | 'submitted_with_invalid_parameters'
+      | 'proposal_submission_failed';
+  };
 };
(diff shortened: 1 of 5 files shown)


src/clients/googleAnalytics.ts
```diff
@@ -7,6 +7,16 @@ export type EventMap = {
...
+  create_proposal: {
+    status:
+      | 'show_modal'
+      | 'proposal_in_progress'
+      | 'submit'
+      | 'submitted_with_invalid_parameters'
+      | 'proposal_submission_failed';
+  };
+  delegate_voting: { type: 'delegate' | 'manual' };
+  collect_rewards: { tokens: string };
```
therealemjy, line 19: How about making that name a bit more explicit? `tokenAddresses`

therealemjy (approved):

src/clients/googleAnalytics.ts
coreyar: Yeah, that makes sense. I imagine this will change though because the token name is probably more useful but that isn't available here from the current data model

therealemjy: Got you. When that happens then we can rename it to `tokenNames` :)

therealemjy (approved):

PR #137, opened 2022-02-25
Structure components to be imported directly from components/index.ts

coreyar:
This is a proposal to export all components from `components/index.ts` and import them directly from `components`

It is nice to export all components from the index of the components dir because it makes importing them easy.
- You don't have to remember the dir
- They can all be imported on a single line

As a side effect  when we drop the old components dirs and the v2 dir we'll only have to update this file and remove `/v2`. The imports in the rest of the application won't change.

### DIFF
--- a/src/components/Basic/BorrowTabs/RepayBorrowTab.tsx
+++ b/src/components/Basic/BorrowTabs/RepayBorrowTab.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState, useCallback } from 'react';
-import { Button } from 'components/v2/Button';
+import { Button } from 'components';
 import NumberFormat from 'react-number-format';
 import { connectAccount } from 'core';
 import BigNumber from 'bignumber.js';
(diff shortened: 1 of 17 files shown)


therealemjy (approved):
Yep, good idea!

src/components/Basic/BorrowTabs/RepayBorrowTab.tsx
```diff
@@ -205,11 +205,7 @@ function RepayBorrowTab({ asset, changeTab, onCancel, setSetting }: Props & Disp
           <>
             <img src={asset.img} alt="asset" />
             <p className="center warning-label">
-              To Repay
-              {' '}
-              {asset.name}
-              {' '}
-              to the Venus Protocol, you need to enable it first.
+              To Repay {asset.name} to the Venus Protocol, you need to enable it first.
```
coreyar, line 208: It looks like we are still having issues with eslint / prettier. I am not sure why else this is changing.

src/components/Basic/DelegationVoting.tsx
```diff
@@ -97,8 +97,8 @@ const VotingWrapper = styled.div`
 `;
 
 interface Props extends RouteComponentProps {
-  isLoading: boolean,
-  onDelegate: (address: string) => void,
+  isLoading: boolean;
+  onDelegate: (address: string) => void;
```
coreyar, line 101: same here

PR #139, opened 2022-02-25
Add slider component

coreyar:
Adds a slider component based on designs.

I used the css prop because I thought it would get picked up by style-lint, the current style implementation for Dropdown isn't being picked up either.

I can update to match Dropdown but I wanted to flag the issue with `style-lint`

### DIFF
--- a/src/components/v2/Slider/Styles.ts
+++ b/src/components/v2/Slider/Styles.ts
@@ -1,0 +1,26 @@
+import { css } from '@emotion/react';
+import { useTheme } from '@mui/material';
+
+export const styles = ({ over }: { over: boolean }) => {
+  const theme = useTheme();
+  return css`
+    color: ${over ? theme.palette.error.slider : theme.palette.success.slider};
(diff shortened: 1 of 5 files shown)


coreyar:
We may also want to update this to be a controlled component

src/components/v2/Slider/Styles.ts
```diff
@@ -0,0 +1,26 @@
...
+      height: 8px;
+      color: ${theme.palette.background.black};
+    }
+    .MuiSlider-mark {
+      width: 4px;
+      height: 8px;
+      color: ${theme.palette.error.slider};
+    }
+  `;
+};
```
therealemjy, line 24: Nice. I think it looks clean this way.

therealemjy (approved):

src/components/v2/Slider/Styles.ts
therealemjy, line 24: Nice. I think it looks clean this way.

therealemjy (approved):

PR #140, opened 2022-02-25
Add paths to stories for better organization

coreyar:
We can path our stories so they are nested in the menu for better organization


therealemjy (approved):
Nice 👍

PR #146, opened 2022-03-01
Add support for React Query + Migrate Faucet container to React Query

therealemjy:
The `Faucet` container only makes one HTTP request (when requesting funds), so I figured this would be a good starting point to integrate React Query.

Let me know your thoughts!


kkirka:
Task linked: [CU-21pcr4f Migrate Faucet container to React Query](https://app.clickup.com/t/21pcr4f)

src/containers/Main/Faucet.tsx
```diff
@@ -92,48 +91,44 @@ const ButtonWrapper = styled.div`
...
-function Faucet({ form, getFromFaucet }: FaucetProps) {
+function Faucet({ form }: FaucetProps) {
   const { getFieldDecorator } = form;
-  const [isLoading, setIsLoading] = useState(false);
+
+  const {
+    mutate: requestFaucetFunds,
+    isLoading: isRequestFaucetFundsLoading,
+  } = useRequestFaucetFunds({
+    onSuccess: (_data, variables) => {
```
coreyar, line 15: What about defining this config in `hooks/useRequestFaucetFunds`?

src/containers/Main/Faucet.tsx
```diff
@@ -92,48 +91,44 @@ const ButtonWrapper = styled.div`
...
 
-function Faucet({ form, getFromFaucet }: FaucetProps) {
+function Faucet({ form }: FaucetProps) {
   const { getFieldDecorator } = form;
-  const [isLoading, setIsLoading] = useState(false);
+
+  const {
+    mutate: requestFaucetFunds,
+    isLoading: isRequestFaucetFundsLoading,
+  } = useRequestFaucetFunds({
```
coreyar, line 14: If we want pure presentational components, this should be in a logical component that passes the props `requestFaucetFunds` and `isRequestFaucetFundsLoading` to the presentational component

coreyar: Currently, we are mocking all requests when snapshotting stories. If we don't find the separation valuable we could continue with this strategy.

src/containers/Main/Faucet.tsx
therealemjy: I wouldn't. In order to make the hooks responsible for fetching data as flexible as possible, I'd only make them deal with the actual HTTP request and handling the data that comes from it.
Any behaviour that's specific to the component rendering the hook (in this case displaying a toast on success and failure) should be coded inside the hook call itself. This way if we want to reuse the hook elsewhere, but make the request fail silently or do something else on success, it's very easy to just write that logic in the relevant callbacks (`onSuccess`, `onError`, `onSettled` etc.).

src/containers/Main/Faucet.tsx
therealemjy: The way I see it, containers can't be pure; their sole purpose is to handle side-effect by passing down props to the component they render. If you're thinking about pages, I'd say they should be separate pure components that compose components together and form the UX using props they're given. Their respective containers would be in charge of handling side-effect (through hooks for the queries for example) and pass down the props needed to the page itself (`requestFaucetFunds` and `isRequestFaucetFundsLoading ` as you mentioned).

Regarding the snapshots, having this separation makes things easier actually. Instead of rendering the containers in Storybook, we can render the page components directly. This way we don't need to mock the requests, we can just pass down mocked properties to the component.

The containers can then be tested functionally.

What do you think of this approach?

coreyar: You are correct, I mistyped my comment (edited for how it should have read). Should we update this PR to follow that strategy then?

therealemjy: Sounds good to me, I'll make a separate `Faucet` component that gets imported and rendered by this container

src/containers/Main/Faucet.tsx
coreyar: That makes sense, although I would posit that a single request would get handled similarly throughout the app. I think it will be nicer with this defined in the "container". As it stands now it seems like a lot for just accessing data in a presentational component

therealemjy: My mistake was not refactoring the `Faucet` completely in this PR. It will make more sense once I'm done with the refactoring, then we can talk about this further!

therealemjy:
I updated this PR so the `Faucet` is divided into two components: a container and a pure component.
The `Faucet` will still need some refactoring (to use React Hook Form instead of antld for example), but the PR is already getting quite big and I feel this is out of its scope.

coreyar (approved):

PR #147, opened 2022-03-01
add Vault VRT & Vault components refactor

RoseRompkxm:
- divide components in `components/Vault` into `BasicVault`(in our case there will be VRT vault and VAI vault) & `VestingVault` (XVS vault)
- migrate icons used in the vault components to `v2/Icon`
- add VRT vault


therealemjy (approved):

src/components/Vault/BasicVault/VaiCard.tsx
```diff
@@ -95,68 +90,43 @@ function VaultCard({ settings }: VaultCardProps) {
...
+              .approve(
+                vaiVaultContract.options.address,
+                amt.toFixed(10),
+              )
+              .send({ from: account })}
+            onWithdraw={async (amt) => vaiVaultContract.methods
+              .withdraw(
+                amt.toFixed(0),
+              )
+              .send({ from: account })}
```
therealemjy: That's a very tiny thing, but you don't need to specify the `async` prefix here, since you're already returning promises.

src/components/Vault/BasicVault/VrtCard.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+interface VaultCardProps {
+  settings: Setting;
+}
+
+const DAYS_OF_YEAR = 365;
+const BLOCK_PER_MINUTE = 60 / 3;
+const BLOCK_PER_DAY = BLOCK_PER_MINUTE * 60 * 24;
+
+function VaultCard({ settings }: VaultCardProps) {
```
therealemjy: `settings` seems to be unused in the component, so I'd say remove it :)

src/components/Vault/BasicVault/VaiCard.tsx
```diff
@@ -95,68 +90,43 @@ function VaultCard({ settings }: VaultCardProps) {
...
+        apy={settings.vaiAPY || 0}
+        totalStakedAmount={new BigNumber(settings.vaultVaiStaked || 0)}
+        totalPendingRewards={totalPendingRewards.div(1e18)}
+        dailyEmission={dailyEmission}
+        onExpand={() => {
+          setExpanded(!expanded);
+        }}
+      />
       <div className="content-container">
         {expanded ? (
```
therealemjy:
```suggestion
        {expanded && (
```
Then see my next comment line 131...

src/components/Vault/BasicVault/VaiCard.tsx
```diff
@@ -95,68 +90,43 @@ function VaultCard({ settings }: VaultCardProps) {
...
+                amt.toFixed(10),
+              )
+              .send({ from: account })}
+            onWithdraw={async (amt) => vaiVaultContract.methods
+              .withdraw(
+                amt.toFixed(0),
+              )
+              .send({ from: account })}
           />
         ) : null}
```
therealemjy:
```suggestion
        )}
```
... this way you don't need to return `null` if `expanded` is false, it will have the same behaviour :)

src/components/Vault/BasicVault/VrtCard.tsx
```diff
@@ -0,0 +1,145 @@
...
+        apy={interestRatePerBlock.multipliedBy(DAYS_OF_YEAR * BLOCK_PER_DAY).div(1e18).multipliedBy(100).toFixed(2)}
+        totalStakedAmount={vaultVrtBalance.div(1e18)}
+        totalPendingRewards={vaultVrtBalance.div(1e18)}
+        dailyEmission={dailyEmission}
+        onExpand={() => {
+          setExpanded(!expanded);
+        }}
+      />
+      <div className="content-container">
+        {expanded ? (
```
therealemjy: Same here:
```suggestion
        {expanded && (
```

src/components/Vault/BasicVault/VrtCard.tsx
```diff
@@ -0,0 +1,145 @@
...
+                amt.toFixed(10),
+              )
+              .send({ from: account })}
+            onWithdraw={async (amt) => vrtVaultProxyContract.methods
+              .withdraw(
+                amt.toFixed(0),
+              )
+              .send({ from: account })}
+          />
+        ) : null}
```
therealemjy:
```suggestion
        )}
```

src/components/Vault/BasicVault/VrtCard.tsx
```diff
@@ -0,0 +1,145 @@
...
+      </div>
+    </VaultCardWrapper>
+  );
+}
+
+const mapStateToProps = ({ account }: State) => ({
+  settings: account.setting,
+});
+
+export default connect(mapStateToProps)(VaultCard);
```
therealemjy: Since we don't use `settings`, we can get rid of all that completely and export default component directly:
```suggestion
export default VaultCard;
```

therealemjy (changes requested):
Overall it looks good to me! Just a few things here and there.

The main thing however is your calls to contracts to fetch data. The way you're doing it isn't wrong, but it is a bit fragile: since you use `Promise.all` to send all the requests at once, if any of them fail they'll all fail and because you're not catching the errors the app will crash.
This is actually a great example when React Query would be very useful.
I've opened a PR to integrate it to the project: https://github.com/VenusProtocol/venus-protocol-interface/pull/146
Can you review it please? Once it's merged I'll be able to comment on your PR with more details on how you could rewrite the requests you're making to make them a bit more robust :)

src/components/Vault/BasicVault/VrtCard.tsx
RoseRompkxm: sure

src/components/Vault/BasicVault/VaiCard.tsx
RoseRompkxm: got it

src/components/Vault/BasicVault/VrtCard.tsx
RoseRompkxm: yep

src/components/Vault/BasicVault/VaiCard.tsx
RoseRompkxm: yep, sometimes just for hinting that this is a Promise returned, especially before TypeScript, but since we've integrated TS, let's remove this redundancy

RoseRompkxm: I'll try to refactor these requests based on it and see how it works :)

therealemjy: Got you! It gets compiled away anyway, so no biggie whether you keep it or not :)

therealemjy (approved):

PR #157, opened 2022-03-05
Add TextField component

therealemjy:
Some context on the PR:
- I've updated some of the background colors present in the palette, as I needed to use them on the `TextField` component
- I've added a `rightAdornment` property on the component so that any React element can be appended to it. In the designs we're currently appending a button, and so that's what I added to the story. The reason I didn't add a button directly to the component, although it is the only case covered by the designs, is so that it makes it easy to use any button and pass it any props wanted
- I set the type of the "size" property wrong on the `Icon` component when I created it. Since it is used in this PR I updated it to a `string` (which it should be, so that we can pass a value returned by the theme like "24px")
- I tried to use as many components from MUI as possible, but their `TextField` is too different from our designs and so it was easier to use a basic `input` tag with custom styles
- In the designs, any number contained in the input gets a comma inserted after every 3 digits. I didn't add this behavior to the component, as it complicates things a lot (we can no longer use a `number` input type but instead have to use a text input with a custom handler on value change) for what I feel doesn't add real UX benefit. I also checked on other exchanges and they all use the default number input behavior, which means users are used to that UX.

### Previews:

Default:
Screenshot 2022-03-05 at 15 40 07

With `label`:
Screenshot 2022-03-05 at 15 40 03

With `description`:
Screenshot 2022-03-05 at 15 40 12

With `hasError`:
Screenshot 2022-03-05 at 15 40 26

With `rightAdornment`:
Screenshot 2022-03-05 at 15 40 21

With `leftIconName`:
Screenshot 2022-03-05 at 15 40 17

### DIFF
--- a/src/components/v2/Button/index.tsx
+++ b/src/components/v2/Button/index.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { Button as MuiButton, ButtonProps } from '@mui/material';
 import { Icon } from 'components';
 // import SwapHorizIcon from '@mui/icons-material/loading';
-interface IButtonProps extends ButtonProps {
+export interface IButtonProps extends ButtonProps {
   className?: string;
   loading?: boolean;
   loadingIconSize?: number;
--- a/src/components/v2/Button/index.tsx
+++ b/src/components/v2/Button/index.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import MuiButton, { ButtonProps } from '@mui/material/Button';
-import { Icon } from 'components';
+import { Icon } from 'components/v2/Icon';
 
 export interface IButtonProps extends ButtonProps {
   className?: string;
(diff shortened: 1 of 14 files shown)


src/components/v2/Button/index.tsx
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
-import { Button as MuiButton, ButtonProps } from '@mui/material';
-import { Icon } from 'components';
-// import SwapHorizIcon from '@mui/icons-material/loading';
-interface IButtonProps extends ButtonProps {
+import MuiButton, { ButtonProps } from '@mui/material/Button';
+import { Icon } from 'components/v2/Icon';
```
therealemjy, line 3: This is to prevent a dependency cycle

coreyar (approved):

PR #158, opened 2022-03-07
Add v2/tooltip

baomastr:


src/components/v2/Tooltip/TooltipStyles.ts
```diff
@@ -0,0 +1,8 @@
+import { useTheme } from '@mui/material';
+
+export const useTooltipStyles = () => {
+  const theme = useTheme();
+  return {
+    /* TODO: custom styles */
+  };
+};
```
therealemjy: It looks like we can get rid of this file

coreyar: This should also be written using css object interpolation so it gets linted properly

src/theme/MuiThemeProvider/muiTheme.ts
```diff
@@ -192,5 +193,21 @@ export default createTheme({
         },
       },
     },
+    MuiTooltip: {
```
coreyar: Rather then override styles at the base level, we can override them with styles in the component directory

coreyar: Rather then override styles at the base level, we can override them with styles in the component directory

src/theme/MuiThemeProvider/muiTheme.ts
```diff
@@ -13,6 +13,8 @@ export const PALETTE = {
   background: {
     default: 'rgba(31, 32, 40, 1)',
     paper: 'rgba(40, 41, 49, 1)',
+    black: '#1F2028',
+    grey: '#383944',
```
therealemjy: These seem unused, so can be removed

therealemjy (commented):
There's some discrepancies with the designs:
Preview:
Screenshot 2022-03-21 at 16 28 28

Designs:
Screenshot 2022-03-21 at 16 29 44

The background color is incorrect, it should be `theme.palette.background.grey` and the shadow is missing.

coreyar:
@therealemjy I don't see the same color issue any more. I believe it was fixed in this commit [a14c8ed](https://github.com/VenusProtocol/venus-protocol-interface/pull/158/commits/a14c8ed9a105393d168cfd429da8bcc9d9a57a85)

therealemjy (approved):
👍

therealemjy (approved):

PR #159, opened 2022-03-07
Fix ignored typescript errors

coreyar:
Working through ignored typescript comments and replacing them with better types


therealemjy (approved):
Nice 👍

PR #160, opened 2022-03-07
Fix type error on size prop

coreyar:
Fixes type inconsistencies with the size prop, that got merged into dev recently


therealemjy (approved):
Oops, indeed! Thanks for fixing that.

PR #164, opened 2022-03-09
Update color palette for legacy components

baomastr:


src/containers/Layout/Sidebar.tsx
```diff
@@ -362,8 +376,82 @@ function Sidebar({ history, setSetting }: SidebarProps) {
...
+                x2="-5.57766"
+                y2="-2.83053"
+                gradientUnits="userSpaceOnUse"
+              >
+                <stop stopColor="#5433FF" />
+                <stop offset="0.5" stopColor="#20BDFF" />
+                <stop offset="1" stopColor="#5CFFA2" />
+              </linearGradient>
+            </defs>
+          </svg>
```
therealemjy: That's Venus' logo right? We should put this in an SVG file, which can then be imported as a react component:
```
import { ReactComponent as Logo } from 'assets/img/v2/venusLogo.svg';
```

Also, I don't know if you got a chance to see the conversation we had with @coreyar yesterday, but we've settled on adding a `v2` folder inside `assets/img` to separate the assets used by v2 from the current ones. We can put the SVG file of the new logo inside there.

src/theme/MuiThemeProvider/muiTheme.ts
```diff
@@ -84,42 +84,50 @@ export default createTheme({
...
-      lineHeight: '1.5',
+      // line-height has default value 1.5 for all font-sizes. can be defined in body
+      // lineHeight: '1.5',
       fontWeight: 400,
     },
     button: {
       fontSize: '14px',
-      lineHeight: '21px',
+      // line-height has default value 1.5 for all font-sizes. can be defined in body
+      // lineHeight: '1.5',
```
therealemjy: Let's move these typography changes to a separate PR, so that we don't try to do everything at once.

src/containers/Layout/Sidebar.tsx
baomastr: sure. forgot about it while fixed in online mode

src/theme/MuiThemeProvider/muiTheme.ts
baomastr: done

coreyar (approved):

PR #167, opened 2022-03-09
Fix VRT vault & VRT swap issues in test environment

RoseRompkxm:
- hide entries of VRT vault & VRT swap in the production environment
- update the latest contract addresses
- keep solving issues on this branch...

### DIFF
--- a/src/hooks/useContract.ts
+++ b/src/hooks/useContract.ts
@@ -13,8 +13,8 @@ import {
   getXvsVaultContract,
   getTokenContractByAddress,
   getGovernorBravoContract,
-  getVrtConverterContract,
-  getXvsVestingContract,
+  getVrtConverterProxyContract,
+  getXvsVestingProxyContract,
   getVrtTokenContract,
   getVrtVaultProxyContract,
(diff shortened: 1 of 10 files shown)


src/hooks/useContract.ts
```diff
@@ -87,14 +87,14 @@ export const useGovernorBravo = () => {
 };
 
 // VRT conversion
-export const useVrtConverter = () => {
+export const useVrtConverterProxy = () => {
   const web3 = useWeb3();
-  return useMemo(() => getVrtConverterContract(web3), [web3]);
+  return useMemo(() => getVrtConverterProxyContract(web3), [web3]);
```
coreyar, line 92: I'm curious why the name change? Will be change it back after testing?

coreyar (approved):

therealemjy (approved):
Nice :)

src/hooks/useContract.ts
RoseRompkxm: No, it won't change after testing, we are actually interacting with the proxy contract, I want it to be more precise

PR #168, opened 2022-03-09
Add table component

coreyar:
Add a basic table component with in place ordering that can be overridden with a callback.

There are outstanding questions about pagination, loading, empty and error states that haven't been included in designs yet.

Screen Shot 2022-03-09 at 14 37 21


src/components/v2/Table/styles.ts
```diff
@@ -0,0 +1,56 @@
+import { css } from '@emotion/react';
+import { useTheme } from '@mui/material';
+import { AnyStyledComponent } from 'styled-components';
```
therealemjy: This seems unused, so it could be removed.

src/components/v2/Table/styles.ts
```diff
@@ -0,0 +1,56 @@
...
+export const useStyles = () => {
+  const theme = useTheme();
+  return {
+    tableContainer: css`
+      background-color: transparent;
+      background-image: none;
+      box-shadow: none;
+    `,
+    table: css`
+      min-width: 650px;
```
therealemjy: This breaks the responsiveness unfortunately
Screenshot 2022-03-09 at 18 19 32

src/components/v2/Table/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const [rows, setRows] = useState(data);
+  const onRequestOrder = (property: typeof columns[number]['id']) => {
+    let newOrder: 'asc' | 'desc' = 'asc';
+    if (property === orderBy) {
+      newOrder = order === 'asc' ? 'desc' : 'asc';
+    }
+    setOrderBy(property);
+    setOrder(newOrder);
+    if (fetchData) {
+      fetchData({ order: { by: property, direction: newOrder } });
```
therealemjy: Why would the component refetch the data if it's just re-ordering it?
The table could be pure and just display data without handling any fetch request.

src/components/v2/Table/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { useStyles } from './styles';
+
+interface ITableRowProps {
+  key: string;
+  render: () => React.ReactNode | string;
+  value: string | number | boolean;
+}
+
+interface ITableProps {
+  title: string;
```
therealemjy, line 20: `title` might be misleading, given it's only used for accessibility purpose. How about just naming it `aria-label`, like the prop you're passing to the table?

src/components/v2/Table/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  data: ITableRowProps[][];
+  columns: { id: string; label: string; orderable: boolean }[];
```
therealemjy: Since you're using a `key` prop on the data, I'd suggest using `key` here as well instead of `id`, just to make it extra obvious that both are linked.

src/components/v2/Table/styles.ts
```diff
@@ -0,0 +1,56 @@
...
+        }
+      }
+    `,
+    tableSortLabelIcon: ({ active }: { active: boolean }) => css`
+      &.MuiTableSortLabel-icon {
+        fill: ${theme.palette.text.primary};
+      }
+      .Mui-active &.MuiTableSortLabel-icon {
+        fill: ${active ? theme.palette.success.slider : theme.palette.text.primary};
+        color: ${active ? theme.palette.success.slider : theme.palette.text.primary};
```
therealemjy, line 55: I know the v2 palette isn't merged with `develop` yet, but perhaps you could add a `TODO` comment to remember to change `theme.palette.success.slider` so it refers to the v2 palette?

src/components/v2/Table/styles.ts
```diff
@@ -0,0 +1,56 @@
...
+    tableSortLabel: css`
+      &.MuiTableSortLabel-root {
+        span {
+          color: ${theme.palette.text.secondary};
+        }
+      }
+      .MuiSvgIcon-root {
+        display: block;
+        margin-right: 8px;
+        margin-left: 8px;
```
therealemjy: Let's use the theme spacing:
```
margin-right: ${theme.spacing(1)};
margin-left: ${theme.spacing(1)};
```

src/components/v2/Table/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  fetchData?: (params: { order: { by: string; direction: 'asc' | 'desc' } }) => void;
+}
+
+export function Table({ columns, data, fetchData, title }: ITableProps) {
+  const styles = useStyles();
+  const [orderBy, setOrderBy] = useState<typeof columns[number]['id'] | undefined>(undefined);
+  const [order, setOrder] = useState<'asc' | 'desc' | undefined>('asc');
```
therealemjy: A small thing, but how about `orderDirection` to keep the same nomenclature you used for `orderBy`?

src/components/v2/Table/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const [rows, setRows] = useState(data);
```
therealemjy: Instead of storing the ordered rows in a state, you can memoize them and automatically update them when `order`, `orderBy` or the data itself change:

```typescript
const rows = React.useMemo(async () => {
  if (!orderBy) {
    return data;
  }

  const rowIndex = columns.findIndex(column => column.id === orderBy);
  
  return [...data].sort((a, b) => {        
    if (a[rowIndex].value < b[rowIndex].value) {
      return newOrder === 'asc' ? -1 : 1;
    }
    if (a[rowIndex].value > b[rowIndex].value) {
      return newOrder === 'asc' ? 1 : -1;
    }
    return 0;
  });
}, [data, orderBy, order]);
```

Then `onRequestOrder` would only change `orderBy` and `order`:

```typescript
const onRequestOrder = (property: string) => {
  const newOrder = property === orderBy && order === 'asc' ? 'desc' : 'asc';
  
  setOrderBy(property);
  setOrder(newOrder);
});
```

The reason behind wanting to avoid storing the ordered rows in a state is that if the `data` prop changes, the table won't reflect the update until the user presses on a header to change the order.

src/components/v2/Table/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+      setRows(newRows);
+    }
+  };
+
+  return (
+    <TableContainer css={styles.tableContainer} component={Paper}>
+      <TableMUI css={styles.table} aria-label={title}>
+        <Head columns={columns} orderBy={orderBy} order={order} onRequestOrder={onRequestOrder} />
+        <TableBody>
+          {rows.map((row: ITableRowProps[]) => (
```
therealemjy: Small thing: you don't need to specify the type of `row`. Since `rows` is already typed as `ITableRowProps[][]`, TypeScript will infer that the type of `row` is `ITableRowProps[]`.

src/components/v2/Table/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+            <TableRow key={`${row[0].key}`}>
```
therealemjy: Won't that output the same key for all the rows (since they all follow the same data structure)?

src/components/v2/Table/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+              {row.map(({ key, render }: ITableRowProps) => (
+                <TableCell key={key}>{render()}</TableCell>
```
therealemjy: Same thing here, won't the keys be repeated across cells of different rows?

therealemjy (commented):
It looks like the sorting icon and the text inside each header cell aren't vertically aligned with each other:
Screenshot 2022-03-09 at 19 01 06

src/components/v2/Table/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+import TableCell from '@mui/material/TableCell';
+import TableContainer from '@mui/material/TableContainer';
+import TableRow from '@mui/material/TableRow';
+import Paper from '@mui/material/Paper';
+import Head from './Head';
+import { useStyles } from './styles';
+
+interface ITableRowProps {
+  key: string;
+  render: () => React.ReactNode | string;
```
therealemjy, line 15: Nice 👍

src/components/v2/Table/index.tsx
coreyar: If the data is paginated it would need to be refetched

src/components/v2/Table/styles.ts
coreyar: Not sure how it passed linting with this 🤔

src/components/v2/Table/styles.ts
coreyar: I don't think this is related to the storybook decorator more than the minimum width on the table. I would think we would want a minimum width on the table to keep things spaced nicely and usable with a horizontal scroll. 

Or we need a mobile layout to display on smaller screens

src/components/v2/Table/index.tsx
coreyar: I added the title, it isn't clear from the component part of the designs that the table has a header but the current table component does and the table used on the page does so I just included it

src/components/v2/Table/index.tsx
coreyar: the first value of each row is different but I suppose that might not be guaranteed

src/components/v2/Table/index.tsx
coreyar: This is going to depend on how you format the data but we can add `uid` here for security

src/components/v2/Table/index.tsx
therealemjy: My bad, I didn't see you actually are displaying the title. Fine by me then

src/components/v2/Table/index.tsx
therealemjy: We can keep it in if you want, but if we are to handle pagination I'd say only then we should adapt the logic of the component. Currently I don't think `fetchData` will be used anywhere.

src/components/v2/Table/styles.ts
therealemjy: The thing is that this minimum width isn't based on the actual width of the content itself, since it's set as a fixed value. If we display a table that has only two columns for example, it would still be 650px wide which will make it huge.

Being able to apply a minimum width to the table is a nice idea I think, I would just make it a bit more modular. For example, we can make the table take the full available horizontal width, and add a `css` prop to the component which will apply to the container, so that a component/page rendering it can easily extend its style and so set a minimum width if needed.

src/components/v2/Table/styles.ts
therealemjy: Yeah ^^

therealemjy (approved):

therealemjy (approved):
Still looks good to me 👍

PR #169, opened 2022-03-10
Use exponential notation for large APYs

kkirka:
**Problem:** Currently we show the word "Infinity" when APYs are too large. This may happen when a new market is deployed, and the compound yield from XVS distribution is far above any sane values. This doesn't last long, but this is perfectly valid (e.g. if you supply 1 cent and you're the only supplier, you will receive the full distribution per day). Showing "Infinity" in this case is misleading.
    
**Solution:** Display large APYs in exponential notation (e.g., 1.22e+800%).


therealemjy (approved):
Nice

coreyar (approved):

PR #173, opened 2022-03-11
Update typography (font family + font styles)

therealemjy:


coreyar:
I'd expect the snapshots to change with the new fonts.

therealemjy:
Screenshot 2022-03-14 at 16 20 54

The screenshots from reg-viz show that the font is different, yet it's not considering this as a visual change. Perhaps the new font is too close to the old one to be considered a visual change?

I don't know how it works under the hood, if it alerts about ay pixel difference or if it has some tolerance.

therealemjy:
I added a story for the typography as we discussed in Slack, so now reg-viz is picking up the visual changes. It seems it only picks up changes coming from new stories. I've created a ticket to have a look at that: https://app.clickup.com/t/23ncru2

coreyar (approved):

PR #177, opened 2022-03-14
V2 button styles

coreyar:
- Style v2 buttons (Primary, secondary, text) with disabled and loading states
- Add stories for Primary, Secondary, Text and loading buttons
- Update existing buttons to use new prop api

### DIFF
--- a/src/components/Vote/Proposals.tsx
+++ b/src/components/Vote/Proposals.tsx
@@ -31,6 +31,8 @@ const ProposalsWrapper = styled.div`
       background-color: var(--color-gold);
       border: none;
       color: #fff;
+      width: 150px;
+      height: 40px;
       span {
         font-size: 14px;
         font-weight: 500;
@@ -245,9 +247,6 @@ function Proposals({
               title="You must have the voting power of at least 300K XVS to propose"
             >
               <Button
-                // @ts-expect-error ts-migrate(2322) FIXME: Type '{ children: (string | false | Element)[]; wi... Remove this comment to see the full error message
-                width={150}
-                height={40}
                 className="button create-proposal-btn"
                 onClick={handleShowProposalModal}
                 disabled={notProposable}
--- a/src/components/Vote/Proposals.tsx
+++ b/src/components/Vote/Proposals.tsx
@@ -2,13 +2,14 @@ import React, { useState, useEffect } from 'react';
 import styled from 'styled-components';
 import Web3 from 'web3';
 import BigNumber from 'bignumber.js';
-import { Pagination, Icon, Tooltip, Button } from 'antd';
+import { Pagination, Icon, Tooltip } from 'antd';
 import Proposal from 'components/Basic/Proposal';
 import ProposalModal from 'components/Vote/ProposalModal';
 import toast from 'components/Basic/Toast';
 import LoadingSpinner from 'components/Basic/LoadingSpinner';
 import arrowRightImg from 'assets/img/arrow-right.png';
 import { Card } from 'components/Basic/Card';
+import { PrimaryButton } from 'components';
 import { useWeb3React } from '@web3-react/core';
 import { Proposal as ProposalObject } from 'types';
 import { useToken, useGovernorBravo } from '../../hooks/useContract';
@@ -25,20 +26,6 @@ const ProposalsWrapper = styled.div`
       font-size: 17px;
       font-weight: 900;
       color: var(--color-text-main);
-    }
-    .create-proposal-btn {
-      border-radius: 5px;
-      background-color: var(--color-gold);
-      border: none;
-      color: #fff;
-      width: 150px;
-      height: 40px;
-      span {
-        font-size: 14px;
-        font-weight: 500;
-        color: var(--color-text-main);
-        text-transform: capitalize;
-      }
     }
   }
 
@@ -72,14 +59,7 @@ const ProposalsWrapper = styled.div`
     }
 
     .button {
-      width: 200px;
       flex-direction: row-reverse;
-      border: none;
-      span {
-        font-size: 16px;
-        font-weight: 900;
-        color: var(--color-text-main);
-      }
 
       img {
         width: 26px;
@@ -246,13 +226,14 @@ function Proposals({
               placement="top"
               title="You must have the voting power of at least 300K XVS to propose"
             >
-              <Button
-                className="button create-proposal-btn"
+              <PrimaryButton
+                className="button"
                 onClick={handleShowProposalModal}
                 disabled={notProposable}
+                loading={isLoading}
               >
-                {isLoading && <Icon type="loading" />} Create Proposal
-              </Button>
+                Create Proposal
+              </PrimaryButton>
             </Tooltip>
           )}
         </div>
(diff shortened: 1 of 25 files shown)


src/components/v2/Button/index.tsx
```diff
@@ -1,24 +1,42 @@
...
 import MuiButton, { ButtonProps } from '@mui/material/Button';
 import { Icon } from 'components/v2/Icon';
+import useStyles from './styles';
 
 export interface IButtonProps extends ButtonProps {
   className?: string;
   loading?: boolean;
-  loadingIconSize?: string;
-  loadingIconColor?: string;
+  loadingIcon?: JSX.Element;
```
therealemjy: How about removing this altogether? The loading icon will always be the same and have the same size in order to be consistent.

src/components/v2/Button/styles.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useTheme } from '@mui/material';
+
+export const styles = () => {
+  const theme = useTheme();
+  return css`
+    padding: ${theme.spacing(1.5)} ${theme.spacing(3)};
+    border-radius: ${theme.shape.borderRadius};
+    font-size: ${theme.typography.body1.fontSize};
+    color: ${theme.palette.text.primary};
+    min-width: 150px;
```
therealemjy: Looking at the designs, it seems like the buttons scale with the width of their content, so I'd just remove this line.
For instances where they take all the available horizontal space, we can either add a property to define that variant (`fullWidth` for example) or if that is used very rarely we could just consider that we'll pass styles to the component itself in order to make it expand.

src/components/v2/Button/styles.tsx
```diff
@@ -0,0 +1,41 @@
...
+    :active {
```
therealemjy, line 13: Don't you need to specify that it applies to the tag?
```suggestion
    &:active {
```

src/components/v2/Button/styles.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    &.MuiButton-text {
+      color: ${theme.palette.button.main};
+    }
+    &.MuiButton-text:hover {
+      color: ${theme.palette.button.dark};
+      background-color: transparent;
+    }
+    &.MuiButton-text:active {
+      color: ${theme.palette.button.light};
```
therealemjy, line 39: You might have to add the transparent background here as well. That's an edge case, but if you press on the button then slide your cursor outside the button the blue background shows:
Screenshot 2022-03-14 at 16 00 57

src/components/v2/Button/index.tsx
coreyar: yeah that works for me

src/components/v2/Button/styles.tsx
coreyar: I don't think that is the case here are some examples where the button doesn't take up all available horizontal space
Screen Shot 2022-03-15 at 16 57 42

src/components/v2/Button/index.tsx
coreyar: There are examples where this isn't the case though - https://github.com/VenusProtocol/venus-protocol-interface/pull/177/files#diff-487e3914f4237f6cbbf5afb459980429965b17a28184cbee3a7a8e9a458642a9R136

src/components/v2/Button/styles.tsx
coreyar: Either way is correct

src/components/v2/Button/styles.tsx
coreyar: Nice catch

src/components/v2/Button/index.tsx
coreyar: Actually after checking that out I think 28px is a good default size

therealemjy: Ideally we should use a value that can be divided by the base spacing. perhaps 24px or 32px?

src/components/v2/Button/styles.tsx
therealemjy: Yeah, as I mentioned it scales with the width of its content, so I'd just remove the line.
Then for instances where it does take all of the available horizontal space, we can add a `fullWidth` prop which will apply `width: 100%` to it.

src/components/v2/Button/styles.tsx
therealemjy: Nice, didn't know that

src/components/Vote/Proposals.tsx
```diff
@@ -31,6 +31,8 @@ const ProposalsWrapper = styled.div`
       background-color: var(--color-gold);
       border: none;
       color: #fff;
+      width: 150px;
+      height: 40px;
```
therealemjy: We should use the the `spacing` value from the theme to calculate those.

coreyar: I don't know that is particularly useful in this case. I understand spacing to be a multiplier on a value to create space between elements. The benefit is you could modify the base value and the app should nicely increase or decrease spaces between elements proportionally. I don't think we want to mix the size of elements with that

src/components/v2/Button/index.tsx
coreyar: I don't think we should mix base spacing with sizing elements

therealemjy: So you'd use the base spacing to express margins and paddings only?

src/components/Vote/Proposals.tsx
therealemjy: The way I've used the base spacing in the past was to make sure all the elements of the UI used a common base to define anything related to sizes and spacings, so that the UI flows nicely and you don't get odd values.
I don't think it can be used for responsiveness anyway, even if it is only used for spacings, as sizes would not update accordingly and it would quickly look odd.

src/components/v2/Button/index.tsx
coreyar: Yes, that is primarily how I would expect it to be used.

src/components/v2/Button/styles.tsx
coreyar: done

src/components/v2/Button/index.tsx
therealemjy: Ok, I'm onboard.

src/components/v2/Button/styles.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export const styles = ({ fullWidth }: { fullWidth: boolean }) => {
+  const theme = useTheme();
+  return css`
+    padding: ${theme.spacing(1.5)} ${theme.spacing(3)};
+    border-radius: ${theme.shape.borderRadius};
+    font-size: ${theme.typography.body1.fontSize};
+    color: ${theme.palette.text.primary};
+    text-transform: none;
+    ${fullWidth ? 'width: 100%;' : ''}
```
therealemjy, line 12:
```suggestion
    ${fullWidth && 'width: 100%;'}
```

therealemjy (approved):
Looks good :)
Left a very tiny comment, feel free to merge though.

PR #178, opened 2022-03-14
Move global styles to components

coreyar:
Move globally defined styles for components into the component directories to have a single source of truth for components/ custom styles

I left the input styles because there was a comment that it is addressing and issue with antd.


src/components/v2/Layout/PageContainer/PageContainerStyles.ts
```diff
@@ -1,11 +1,21 @@
...
-    minHeight: '100vh',
-  },
-  footer: { position: 'fixed', bottom: 0, left: 0, ...layoutOffset },
+export const useStyles = () => {
+  const theme = useTheme();
+  return {
+    main: css`
+      flex-grow: 1;
+      padding: 3;
+      padding-bottom: 6;
```
therealemjy: I know it's not your code, but since it's `v2` I'll still review it:
```suggestion
      padding: 3px 3px 6px 3px;
```

src/components/v2/Toggle/styles.ts
```diff
@@ -38,8 +39,29 @@ export const useStyles = () => {
...
+//       },
+//       '&+.MuiSwitch-track': {
+//         backgroundColor: defaultTheme.palette.background.default,
+//       },
+//       '&.Mui-checked+.MuiSwitch-track': {
+//         backgroundColor: defaultTheme.palette.background.default,
+//       },
+//     },
+//   },
+// },
```
therealemjy: I'd suggest just getting rid of that

therealemjy (approved):
Nice, that cleans up nicely. I had two minor comments only

PR #181, opened 2022-03-15
Update Web3 + refactor auth logic to allow users to log out

therealemjy:
As of today, Web3 does not allow users to disconnect their account from an app, they have to do that from their wallet interface directly. However, what we can do is mark them as disconnected on our app and remove their account details from the front end so that they no longer display.

I took this PR as an opportunity to refactor the auth flow, with some functions having a lot of nested `IF` statements. My aim was to make the logic easier to read and comprehend.

### DIFF
--- a/src/hooks/useAuth.ts
+++ b/src/hooks/useAuth.ts
@@ -10,42 +10,65 @@ import {
   WalletConnectConnector,
 } from '@web3-react/walletconnect-connector';
 import toast from 'components/Basic/Toast';
-import { connectorLocalStorageKey } from '../config';
+import { LS_KEY_IS_USER_LOGGED_IN } from '../config';
 import { connectorsByName, ConnectorNames } from '../utilities/connectors';
 import { setupNetwork } from '../utilities/wallet';
 
 const useAuth = () => {
   const { activate, deactivate } = useWeb3React();
   const login = useCallback(
-    (connectorID: ConnectorNames) => {
+    async (connectorID: ConnectorNames) => {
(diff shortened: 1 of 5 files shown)


src/hooks/useAuth.ts
```diff
@@ -10,42 +10,65 @@ import {
...
-                walletConnector.walletConnectProvider = undefined;
-              }
-              toast.error({ title: 'Please authorize to access your account' });
-            } else {
-              toast.error({ title: error.message });
-            }
-          }
+      if (!connector) {
+        toast.error({
+          title: 'An internal error occurred: wrong connector config. Please try again later',
```
coreyar: Would this happen if someone the user tries to connect with an unsupported wallet? I think that would be a nicer error message for the user and should also be good information to debug if it ever happened

therealemjy: I think that's the idea yeah, I'm not too sure to be honest. When I saw that, I actually thought to myself that this case should never happen as we are the ones defining what wallets can be connected or not, so the only way this issue would come up is if an incorrect `connectorID` is passed to the login function.

I agree we can use a nicer message then, maybe something like "An internal error occurred: unsupported wallet. Please try again later". The reason I'd still keep the "internal error" part in the message is because as I mentioned I think this case should never happen if our typing is correct.

coreyar: yeah which would make it an error to Sentry in addition to a nice user error message

coreyar (approved):

PR #185, opened 2022-03-16
[IGNORE] Update reg-viz config

therealemjy:
Ignore this PR, I'm using it to check why reg-viz doesn't pick up visual changes on existing components.


coreyar:
This was resolved by merging master into develop after the default branch was changed

coreyar:
This was resolved by merging master into develop after the default branch was changed

PR #188, opened 2022-03-17
Add support for Coinbase Wallet

therealemjy:


baomastr (approved):

src/utilities/connectors.ts
```diff
@@ -1,13 +1,13 @@
...
+import { WalletLinkConnector } from '@web3-react/walletlink-connector';
 import { BscConnector } from '@binance-chain/bsc-connector';
 import { LedgerConnector } from '@web3-react/ledger-connector';
 import getNodeUrl from './getRpcUrl';
 
 const POLLING_INTERVAL = 12000;
 const rpcUrl = getNodeUrl();
-// @ts-expect-error ts-migrate(2345) FIXME: Argument of type 'string | undefined' is not assig... Remove this comment to see the full error message
-const chainId = parseInt(process.env.REACT_APP_CHAIN_ID, 10);
+const chainId = parseInt(process.env.REACT_APP_CHAIN_ID || '', 10);
```
baomastr: is it necessary to add empty string?

therealemjy: Yeah, otherwise a TypeScript errors is thrown since `process.env.REACT_APP_CHAIN_ID` could be `undefined`, in which case the whole line would cause a JS error at runtime since `parseInt` can't be passed an `undefined` value.

baomastr: seems like unassigned chain id should occur an error. shouldn't we handle this case?

kkirka: Maybe default to 97?

therealemjy: Yeah assigning the `testnet` ID makes sense. I'll make the update 👍

coreyar (approved):

PR #190, opened 2022-03-17
Update Footer to correspond to design of landing page

therealemjy:
Desktop:
Screenshot 2022-03-18 at 08 34 39

Mobile:
Screenshot 2022-03-18 at 08 34 51


src/components/v2/Layout/Footer/Footer.tsx
```diff
@@ -0,0 +1,54 @@
...
+export interface IFooterProps {
+  currentBlockNumber: number;
+}
+
+export const Footer: React.FC<IFooterProps> = ({ currentBlockNumber }) => {
+  const styles = useStyles();
+
+  return (
+    <div css={styles.Container}>
+      <Typography
```
coreyar, line 25: It would be nice to have these setup as components and not use Typography directly. For example, all our a tags will likely have similar props

```
target="_blank"
rel="noreferrer"
component="a"
variant="small2"

File: src/components/v2/Layout/Footer/styles.ts [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+    align-items: center;
+
+    @media only screen and (max-width: 768px) {
+      height: auto;
+      padding: ${theme.spacing(3)} ${theme.spacing(2)};
+      flex-direction: column;
+    }
+  `;
+
+  const Status = css`
```
coreyar: In some places we write styles as an object here we are naming and exporting them directly with capital names. I think we should pick a style and stick to that style

src/containers/Layout/MainLayout.tsx
```diff
@@ -17,7 +18,7 @@ const MainLayoutWrapper = styled.div`
       padding-top: 20px;
       display: flex;
       flex-direction: column;
-      height: calc(100vh - 122px);
+      height: calc(100vh - 125px);
```
coreyar, line 21: Where does 125px come from? If it relates to another element it can be a constant

src/components/v2/Layout/Footer/Footer.tsx
therealemjy: They most probably won't actually, `small2` isn't a common variant for texts.
I don't think there's any advantage in making a separate component for this, it wouldn't be reused enough times to justify that.

src/components/v2/Layout/Footer/styles.ts
therealemjy: We aren't exporting the styles individually, check out the return statement. We return an object with named properties through a hook.

src/containers/Layout/MainLayout.tsx
therealemjy: That's temporary. Once we refactor the layout we shouldn't have to set a height at all, flex display should take care of that nicely.

src/components/v2/Layout/Footer/styles.ts
therealemjy: I've updated the file to look like what you did on your PR for the header. I'm happy to follow this style from now 👍

src/components/v2/Layout/Footer/styles.ts
```diff
@@ -0,0 +1,57 @@
...
+  return {
+    container: css`
+      height: 56px;
+      padding: 0 ${theme.spacing(2)};
+      background-color: var(--color-bg-main);
+      display: flex;
+      justify-content: flex-end;
+      align-items: center;
+
+      @media only screen and (max-width: 768px) {
```
baomastr, line 16: we should use our theme variables for media queries.
like `theme.breakpoints...`

coreyar (approved):

baomastr (approved):

PR #192, opened 2022-03-17
Add coin info and connect button to header

coreyar:
This updates the header to the current state of the designs. Interactions for CoinInfo are still being finalized

### DIFF
--- a/src/components/v2/Layout/Header/CoinInfo.tsx
+++ b/src/components/v2/Layout/Header/CoinInfo.tsx
@@ -1,0 +1,23 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+import BigNumber from 'bignumber.js';
+import commaNumber from 'comma-number';
+import Typography from '@mui/material/Typography';
+import { Paper } from '@mui/material';
+import { Icon } from '../../Icon';
+import { useStyles } from './styles';
+
+const format = commaNumber.bindWith(',', '.');
+
+export const CoinInfo = ({ balance, coin }: { balance: BigNumber; coin: 'xvs' | 'vai' }) => {
+  const styles = useStyles();
+
+  return (
+    <Paper css={styles.coinInfo}>
+      <Icon name={coin} size="20px" />
+      <Typography variant="small1" noWrap component="span">
+        {format(balance.dp(2, 1).toString(10))} {coin.toUpperCase()}
+      </Typography>
+    </Paper>
+  );
+};
--- a/src/components/v2/Layout/Header/CoinInfoContainers.tsx
+++ b/src/components/v2/Layout/Header/CoinInfoContainers.tsx
@@ -1,0 +1,14 @@
+import React from 'react';
+import { useVaiUser } from 'hooks/useVaiUser';
+import { useMarketsUser } from 'hooks/useMarketsUser';
+import { CoinInfo } from './CoinInfo';
+
+export const XvsCoinInfo = () => {
+  const { userVaiBalance } = useVaiUser();
+  return <CoinInfo balance={userVaiBalance} coin="vai" />;
+};
+
+export const VaiCoinInfo = () => {
+  const { userXvsBalance } = useMarketsUser();
+  return <CoinInfo balance={userXvsBalance} coin="xvs" />;
+};
--- a/src/components/v2/Layout/Header/styles.ts
+++ b/src/components/v2/Layout/Header/styles.ts
@@ -3,11 +3,48 @@ import { useTheme } from '@mui/material';
 
 export const useStyles = () => {
   const theme = useTheme();
-  return css`
-    background-color: ${theme.palette.background.default};
-    width: ${theme.shape.layoutOffset.width};
-    margin-left: ${theme.shape.layoutOffset.ml};
-    background-image: none;
-    box-shadow: none;
-  `;
+  return {
+    appBar: css`
+      background-image: none;
+      background-color: transparent;
+      box-shadow: none;
+    `,
+    toolbar: css`
+      justify-content: space-between;
+      display: flex;
+
+      /* @TODO Remove when the ConnectButton is refactored */
+      .connect-btn {
+        height: auto;
+        margin: ${theme.spacing(2)};
+        padding: ${theme.spacing(1.5)} ${theme.spacing(2)};
+        width: 165px;
+      }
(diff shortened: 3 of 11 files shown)


src/components/v2/Layout/Header/CoinInfo.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const XVSCoinInfo = () => {
+  const { userXVSBalance } = useMarketsUser();
+  return <CoinInfo balance={format(userXVSBalance.dp(2, 1).toString(10))} coin="xvs" />;
+};
+
+export const VAICoinInfo = () => {
+  const { userVaiBalance } = useVaiUser();
+  return <CoinInfo balance={format(userVaiBalance.dp(2, 1).toString(10))} coin="vai" />;
+};
```
therealemjy: These components should be kept pure, or they'll be considered containers.

We started discussing that but I don't think we came to an actual conclusion. What we could do is consider containers as components that handle the side effect and can also compose components themselves. What I mean by this is that containers could then not only handle data fetching, calls to contracts etc.. but also hold style and render multiple components/containers.

Doing that means we could have small containers, such as here for the header, that could be nested inside larger containers such as the layout here.
This way we also keep our `components` folder free of side effect.

src/components/v2/Layout/Header/styles.ts
```diff
@@ -3,11 +3,48 @@ import { useTheme } from '@mui/material';
...
+        display: none;
+      }
+    `,
+    rightItemPaper: css`
+      margin: ${theme.spacing(2)};
+      justify-content: center;
+      align-items: center;
+      padding: ${theme.spacing(1)};
+      display: inline-flex;
+      width: 10.3rem;
```
therealemjy: We should use pixels to express size (we don't want them to scale with the main font size of the app).

src/components/v2/Layout/Header/styles.ts
```diff
@@ -3,11 +3,48 @@ import { useTheme } from '@mui/material';
...
+    `,
+    coinInfo: css`
+      display: flex;
+      flex-direction: row;
+      align-items: center;
+      width: 10.3rem;
```
therealemjy: Same thing here about using pixels. Also, this should be a `min-width` instead of a fixed `width`, so that if a user have a large amount of coins their balance can still be displayed without it overflowing.

src/components/v2/Layout/Header/styles.ts
```diff
@@ -3,11 +3,48 @@ import { useTheme } from '@mui/material';
...
+      margin: ${theme.spacing(2)};
+      padding: ${theme.spacing(1.5)} ${theme.spacing(3)};
+      > * {
+        margin: ${theme.spacing(0.5)};
+      }
+      p {
+        color: ${theme.palette.text.primary};
+      }
+    `,
```
therealemjy: Some of the margins are incorrect (space between the button left border and the coin icon too big, space between the coin icon and the amount too small) and there's a box shadow being applied to the buttons that isn't present in the designs:
Screenshot 2022-03-23 at 09 32 30

Screenshot 2022-03-23 at 09 31 34

src/components/v2/Layout/Header/CoinInfo.tsx
```diff
@@ -0,0 +1,34 @@
...
+import React from 'react';
+import commaNumber from 'comma-number';
+import Typography from '@mui/material/Typography';
+import { Paper } from '@mui/material';
+import { useMarketsUser } from 'hooks/useMarketsUser';
+import { useVaiUser } from 'hooks/useVaiUser';
+import { Icon } from '../../Icon';
+import { useStyles } from './styles';
+
+const format = commaNumber.bindWith(',', '.');
```
therealemjy, line 10: I feel this will be used in a lot of places. How about making it a reusable util?

src/components/v2/Layout/Header/CoinInfo.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const CoinInfo = ({ balance, coin }: { balance: string; coin: 'xvs' | 'vai' }) => {
+  const styles = useStyles();
+
+  return (
+    <Paper css={styles.coinInfo}>
+      <Icon name={coin} size="20px" />
+      <Typography variant="small1" noWrap component="p">
```
therealemjy: Semantically speaking, this isn't a paragraph:
```suggestion
      <Typography variant="small1" noWrap component="span">
```

src/components/v2/Layout/Header/CoinInfo.tsx
coreyar: This was the idea that these could be containers, but I wasn't thinking of containers as necessarily only being at the root. I was thinking more to separate of concerns and make testing easier

therealemjy: Ok got you. Yeah I'm onboard with that.
In this case I say let's just move these to the `containers` folder and we're good to go.

src/components/v2/Layout/Header/CoinInfo.tsx
coreyar: I'll open a PR for this

src/components/v2/Layout/Header/CoinInfoContainers.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+export const XvsCoinInfo = () => {
+  const { userVaiBalance } = useVaiUser();
+  return <CoinInfo balance={userVaiBalance} coin="vai" />;
+};
+
+export const VaiCoinInfo = () => {
+  const { userXvsBalance } = useMarketsUser();
+  return <CoinInfo balance={userXvsBalance} coin="xvs" />;
+};
```
therealemjy, line 14: That looks good. But that's still an non-pure component inside the `components` folder which is meant to host only pure components 😂
Here's how I see this: if we want to keep following this logic of separating pure components from containers, then this file just needs to be moved to the `containers` folder and we're good to go.

If you want to have containers next to components directly, that's ok for me too, but in this case then we should get rid of the `containers` folder as then there's no more distinction made between pure components and containers.
That's the last thing for me.

It's up to you to decide on that, I'll go with the option your prefer. Everything else I see in this PR is great and I agree with 👍

therealemjy (approved):

baomastr (approved):

PR #195, opened 2022-03-18
Unify file structure and styles for v2 components

therealemjy:


src/components/v2/Slider/Styles.ts
```diff
@@ -1,7 +1,7 @@
 import { css } from '@emotion/react';
 import { useTheme } from '@mui/material';
 
-export const styles = ({ over }: { over: boolean }) => {
+export const useStyles = ({ over }: { over: boolean }) => {
```
coreyar, line 4: This file should be lowercase

coreyar: actually should be deleted

src/components/v2/Table/Head.tsx
```diff
@@ -6,7 +6,7 @@ import TableHead from '@mui/material/TableHead';
 import TableRow from '@mui/material/TableRow';
 import Box from '@mui/material/Box';
 import { visuallyHidden } from '@mui/utils';
-import { Icon } from '../Icon';
+import { Icon } from 'components/v2/Icon';
```
coreyar: I think it is better to avoid using v2 in imports because at some point we will remove it and if we export and import everything from `components/index` then we won't have to change any imports

In this case, as a relative import it also wouldn't change when we get rid of `v2`

coreyar (approved):

coreyar (approved):

PR #197, opened 2022-03-18
Display logos in a flex container

coreyar:
Display logos in a flex container so they are laied out horizontally


src/components/v2/Layout/Footer/styles.ts
```diff
@@ -32,6 +32,7 @@ export const useStyles = () => {
     `,
     links: css`
       color: ${theme.palette.text.primary};
+      display: flex;
```
therealemjy, line 35: Oops, I made some last minute update and removed this part visibly! Thanks man

therealemjy (approved):

PR #209, opened 2022-03-19
Clean up wallet integration

therealemjy:
The aim of this PR is to clean up the web3/wallet integration. The current codebase has files handling web3 stuff all over the place, so I figured it would be easier to have everything in one folder under `src/clients`.
Some of the files contained a lot of nested `IF` conditions, which in my opinions make things harder to understand, so I've done some minor refactors to unnest these logics in particular.
I haven't changed any of the features or UX, with the exception of some error messages to make them more explicit.

I ran some regression tests manually to check nothing broke, but we'll of course need to run a full set of regression tests before we deploy to prod.


src/clients/web3/getWeb3NoAccount.ts
```diff
@@ -1,18 +1,15 @@
...
 
-const RPC_URL = getNodeUrl();
+import { RPC_URL } from 'config';
 
 // @todo: we will get the "No 'Access-Control-Allow-Origin' header is present on the requested resource"
 // error on the frontend in testnet when the wallet is unconnected, because the bsc official testnet endpoints
 // don't support CORS request right now, and we didn't find any working testnet endpoints for
 // HTTPProvider or WebSocketProvider neither.
-const getWeb3NoAccount = () => {
+export default function getWeb3NoAccount() {
```
coreyar: I personally prefer arrow functions unless there is a reason to use the function declaration.
The main benefits I like from arrow functions are they are more concise and the lexical binding of `this` is more intuitive

src/clients/web3/useWeb3.ts
```diff
@@ -1,9 +1,9 @@
 import { useEffect, useState, useRef } from 'react';
 import Web3 from 'web3';
 import { useWeb3React } from '@web3-react/core';
-import { getWeb3NoAccount } from '../utilities/web3';
+import getWeb3NoAccount from './getWeb3NoAccount';
 
-const useWeb3 = () => {
+export default function useWeb3() {
```
coreyar: Same comment about arrow functions

src/components/Dashboard/Overview.tsx
```diff
@@ -310,7 +309,7 @@ function Overview({ settings, getMarketHistory }: OverviewProps) {
           </div>
           {/* <p className="value">{`$${
               (settings.marketType || 'supply') === 'supply'
-                ? new BigNumber(marketInfo.totalSupply || 0).div(new BigNumber(10).pow(vtokenDecimals)).dp(2, 1).toString(10)
+                ? new BigNumber(marketInfo.totalSupply || 0).div(new BigNumber(10).pow(VTOKEN_DECIMALS)).dp(2, 1).toString(10)
```
coreyar: Do we need to maintain commented code? Should we delete it?

src/index.tsx
```diff
@@ -2,23 +2,10 @@ import React from 'react';
...
-  <BrowserRouter>
-    <Web3ReactProvider getLibrary={getLibrary}>
-      <Web3ReactManager>
-        <App />
-      </Web3ReactManager>
-    </Web3ReactProvider>
-  </BrowserRouter>,
-  document.getElementById('root'),
-);
+ReactDOM.render(<App />, document.getElementById('root'));
```
coreyar, line 9: Nice

src/clients/web3/getWeb3NoAccount.ts
therealemjy: Yeah that works for me! I think I just saw one function being exported as default and just rolled with that.
I agree with your points, I prefer arrow functions too.
I'll update all the exports.

src/components/Dashboard/Overview.tsx
therealemjy: No, good point. I did a global search and replace and didn't see this affected code that was commented out. I'll remove it!

coreyar (approved):

PR #210, opened 2022-03-21
Add expandable drawer

coreyar:
Adds an expandable drawer for navigation visible on non-mobile devies

### DIFF
--- a/src/components/v2/Layout/Sidebar/index.tsx
+++ b/src/components/v2/Layout/Sidebar/index.tsx
@@ -1,78 +1,46 @@
 /** @jsxImportSource @emotion/react */
-import React from 'react';
+import React, { useState } from 'react';
 import { NavLink } from 'react-router-dom';
 import List from '@mui/material/List';
-import ListItem from '@mui/material/ListItem';
+import ListItemButton from '@mui/material/ListItemButton';
 import ListItemIcon from '@mui/material/ListItemIcon';
 import Drawer from '@mui/material/Drawer';
-import { uid } from 'react-uid';
 import Typography from '@mui/material/Typography';
-import { ReactComponent as XVSIcon } from 'assets/img/xvs.svg';
-import { ReactComponent as Logo } from 'assets/img/logo.svg';
+import { ReactComponent as LogoDesktop } from 'assets/img/v2/venusLogoWithText.svg';
+import { ReactComponent as LogoMobile } from 'assets/img/v2/venusLogoPure.svg';
+import { TextButton } from '../../Button';
 import { Toolbar } from '../Toolbar';
+import { Icon } from '../../Icon';
 import { useStyles } from './styles';
-
-const menuItems = [
-  {
-    href: '/dashboard',
-    text: 'Dashboard',
-    icon: <XVSIcon width={20} height={20} />,
-  },
-  {
-    href: '/vote',
-    text: 'Vote',
-    icon: <XVSIcon width={20} height={20} />,
-  },
-  {
-    href: '/xvs',
-    text: 'XVS',
-    icon: <XVSIcon width={20} height={20} />,
-  },
-  {
-    href: '/market',
-    text: 'Market',
-    icon: <XVSIcon width={20} height={20} />,
-  },
-  {
-    href: '/vault',
-    text: 'Vault',
-    icon: <XVSIcon width={20} height={20} />,
-  },
-  {
-    href: '/transaction',
-    text: 'Transaction history',
-    icon: <XVSIcon width={20} height={20} />,
-  },
-  {
-    href: '/dev',
-    text: 'Dev',
-    icon: <XVSIcon width={20} height={20} />,
-  },
-];
+import { menuItems } from '../constants';
 
 export const Sidebar = () => {
-  const styles = useStyles();
+  const [expanded, setExpanded] = useState(true);
+  const styles = useStyles({ expanded });
   return (
-    <Drawer css={styles.drawer} variant="permanent" anchor="left">
+    <Drawer variant="permanent" anchor="left" css={styles.drawer}>
       <Toolbar css={styles.toolbar}>
-        <NavLink to="/">
-          <Logo />
-        </NavLink>
+        <TextButton onClick={() => setExpanded(!expanded)}>
+          <LogoDesktop css={styles.logo} />
+          <LogoMobile css={styles.logoMobile} />
+        </TextButton>
       </Toolbar>
 
       <List css={styles.list}>
         {menuItems.map(({ href, icon, text }) => (
-          <ListItem
-            key={uid(text)}
-            button
-            component={NavLink}
-            to={href}
-            activeStyle={styles.activeMenuItem}
-            css={styles.listItem}
-          >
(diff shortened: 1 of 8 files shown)


src/components/v2/Layout/Sidebar/Sidebar.tsx
```diff
@@ -1,77 +1,46 @@
...
-    <Drawer variant="permanent" anchor="left">
-      <Toolbar sx={styles.toolbar}>
-        <NavLink to="/">
-          <Logo />
-        </NavLink>
+    <Drawer variant="permanent" anchor="left" css={styles.drawer}>
+      <Toolbar css={styles.toolbar}>
+        <TextButton onClick={() => setExpanded(!expanded)}>
+          <LogoDesktop css={styles.logo} height="33" />
+          <LogoMobile css={styles.logoMobile} height="33" />
```
therealemjy: The heights should be assigned through css, so we only have one source of truth for styles.
I know the `Icon` component uses a `size` property, that's because when I built it I had Styled Components in mind that let you assign props directly from the stylesheet (so that only one source of truth is kept). We could change that if we want too.

src/components/v2/Layout/Sidebar/styles.ts
```diff
@@ -0,0 +1,86 @@
...
+      ${theme.breakpoints.up('md')} {
+        display: block;
+      }
+    `,
+    toolbar: css`
+      display: flex;
+      justify-content: center;
+      padding-top: ${theme.spacing(4)};
+      padding-bottom: ${theme.spacing(3)};
+      ${theme.breakpoints.up('md')} {
```
therealemjy: Same thing about the app being desktop first

src/components/v2/Layout/Sidebar/styles.ts
```diff
@@ -0,0 +1,86 @@
...
+        min-height: 0;
+      }
+      ${theme.breakpoints.up('sm')} {
```
therealemjy: Same thing about the app being desktop first

src/components/v2/Layout/Sidebar/styles.ts
```diff
@@ -0,0 +1,86 @@
+import { css } from '@emotion/react';
+import { useTheme } from '@mui/material';
+
+export const useStyles = ({ expanded }: { expanded: boolean }) => {
+  const theme = useTheme();
+  return {
+    drawer: css`
+      display: none;
+      ${theme.breakpoints.up('md')} {
```
therealemjy: Since we're building the website as desktop first app, we need to use theme.breakpoints.down instead.

src/components/v2/Layout/Sidebar/styles.ts
```diff
@@ -0,0 +1,86 @@
...
+        padding-top: ${theme.spacing(3)};
+        padding-bottom: ${theme.spacing(3)};
+        display: inline-flex;
+      }
+
+      .active-menu-item {
+        padding-left: ${theme.spacing(1.5)};
+        .left-border {
+          border: 2px solid ${theme.palette.interactive.primary};
+          border-radius: 0 ${theme.shape.borderRadius}px ${theme.shape.borderRadius}px 0;
```
therealemjy, line 64: Since we have a lot of different border radiuses used in the new design, I wouldn't use the value set in the theme. I actually think we should get rid of it completely.

src/components/v2/Layout/Sidebar/Sidebar.tsx
```diff
@@ -1,77 +1,46 @@
...
-            activeStyle={styles.activeMenuItem}
-            sx={styles.listItem}
-          >
-            <ListItemIcon sx={styles.listItemIcon}>{icon}</ListItemIcon>
-            <Typography variant="caption">{text}</Typography>
-          </ListItem>
+          <ListItemButton key={text} component="li" css={styles.listItem} disableRipple>
+            <NavLink to={href} activeClassName="active-menu-item">
+              <ListItemIcon css={styles.listItemIcon}>
+                <Icon name={icon} size="20" />
```
therealemjy: The size needs to be assigned a unit: `size="20px"`

src/components/v2/Layout/Sidebar/styles.ts
coreyar: I wouldn't get rid of it, I would add them all to the theme personally

therealemjy: Ok, can you elaborate? That's a small thing so I'm cool with that, but it feels like the theme is there to contain values that can be updated dynamically, for example if we want to support a lightmode. Border radiuses won't ever be dynamically changed.

baomastr: Agree, that border radius should be described with theme variables.

therealemjy: Why though?

src/components/v2/Layout/Sidebar/index.tsx
```diff
@@ -1,78 +1,46 @@
...
-            button
-            component={NavLink}
-            to={href}
-            activeStyle={styles.activeMenuItem}
-            css={styles.listItem}
-          >
-            <ListItemIcon css={styles.listItemIcon}>{icon}</ListItemIcon>
-            <Typography variant="caption">{text}</Typography>
-          </ListItem>
+          <ListItemButton key={text} component="li" css={styles.listItem} disableRipple>
```
baomastr, line 31: if we don't need ripple effect in project, we'd better add disableRipple prop as default for button on createTheme level.
so that we just assign it globally and shouldn't add to every component locally

src/theme/MuiThemeProvider/muiTheme.ts
```diff
@@ -111,24 +111,6 @@ export default createTheme({
     },
   },
   components: {
-    MuiButton: {
-      defaultProps: {
-        variant: 'contained',
```
baomastr, line 116: I'd assign `defaultProps: { disableRipple: true }` here if we don't use this effect

src/components/v2/Layout/Sidebar/styles.ts
baomastr: cause it can easily been changed in case of theme v3, for example

src/components/v2/Layout/Toolbar/styles.ts
```diff
@@ -6,8 +6,9 @@ export const useStyles = () => {
   return css`
     padding-top: ${theme.spacing(3)};
     padding-bottom: ${theme.spacing(3)};
-    [${theme.breakpoints.up('md')}]: {
-      min-height: 96;
+    min-height: 96;
+    [${theme.breakpoints.down('md')}]: {
```
baomastr, line 10: in order of mobile first i'd use only `breakpoints.up` (but it's just recommendation)

baomastr (approved):

src/theme/MuiThemeProvider/muiTheme.ts
coreyar: We aren't going to use default theme styles because we aren't going to ever use the components directly and may eventually drop MUI

PR #214, opened 2022-03-22
Add tabs component

therealemjy:
I also had to refactor the `Button` component so that we can support the 4th variant.

### DIFF
--- a/src/components/v2/Button/index.tsx
+++ b/src/components/v2/Button/index.tsx
@@ -1,42 +1,43 @@
 /** @jsxImportSource @emotion/react */
 import React from 'react';
-import MuiButton, { ButtonProps } from '@mui/material/Button';
-import { Icon } from '../Icon';
-import useStyles from './styles';
+import Typography from '@mui/material/Typography';
 
-export interface IButtonProps extends ButtonProps {
-  className?: string;
+import { Icon } from 'components/v2/Icon';
+import useStyles from './styles';
+import { Variant } from './types';
+
+export interface IButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+  label: string;
   loading?: boolean;
-  fillWidth?: boolean;
+  fullWidth?: boolean;
+  small?: boolean;
+  variant?: Variant;
 }
 
 export const Button = ({
   className,
   loading,
   disabled,
-  startIcon = null,
   fullWidth = false,
-  ...restProps
+  small = false,
+  variant = 'primary',
+  label,
+  ...otherProps
 }: IButtonProps) => {
-  const styles = useStyles({ fullWidth });
+  const styles = useStyles({ fullWidth, variant, small });
+
   return (
-    <MuiButton
-      css={styles}
-      className={className}
-      startIcon={loading ? <Icon name="loading" size="28px" /> : startIcon}
-      disabled={loading || disabled}
-      disableRipple
-      {...restProps}
-    />
+    <button css={styles.button} disabled={loading || disabled} type="button" {...otherProps}>
+      {loading && <Icon name="loading" size="28px" css={styles.loadingIcon} />}
+
+      <Typography css={styles.label} component="span" variant={small ? 'small1' : 'body1'}>
+        {label}
+      </Typography>
+    </button>
   );
 };
 
-export const PrimaryButton = (props: IButtonProps) => (
-  <Button variant="contained" color="button" {...props} />
-);
-
-export const SecondaryButton = (props: IButtonProps) => (
-  <Button variant="outlined" color="button" {...props} />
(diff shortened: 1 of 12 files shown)


src/components/v2/Button/index.tsx
```diff
@@ -1,42 +1,48 @@
...
-      css={styles}
+    <button
+      css={styles.button}
       className={className}
-      startIcon={loading ? <Icon name="loading" size="28px" /> : startIcon}
       disabled={loading || disabled}
-      disableRipple
-      {...restProps}
-    />
+      type="button"
```
coreyar, line 33: It is probably that we will want to set this to "submit" for certain use cases

coreyar: or is the idea to override this with `otherProps`

src/components/v2/TextField/TextField.tsx
```diff
@@ -16,7 +15,7 @@ export interface ITextFieldProps extends Omit<InputHTMLAttributes<HTMLInputEleme
 }
 
 export const TextField: React.FC<ITextFieldProps> = ({
-  css,
+  className,
```
coreyar, line 18: Since we are doing all of our styles in js the css prop makes more sense here. With a `className` you would need to use it within a `css` prop on a parent component, whereas by passing in a `css` object you can style this component directly.

src/components/v2/Tabs/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  onChange: (newIndex: number) => void;
+  className?: string;
+  fullWidth?: boolean;
+}
+
+export const Tabs = ({
+  tabTitles,
+  activeTabIndex,
+  onChange,
+  className,
```
coreyar, line 19: Rather then use a className we can pass in an array of `css` objects

src/components/v2/Tabs/index.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Tabs } from '.';
+
+export default {
+  title: 'Components/Tabs',
+  component: Tabs,
+  decorators: [withThemeProvider, withCenterStory({ width: 350 })],
+} as ComponentMeta<typeof Tabs>;
+
+export const Default = () => (
+  <State initial={{ activeTabIndex: 0 }}>
```
coreyar, line 14: Nice, we have a `withState` decorator but it currently is customized for a checkbox. Would be nice to have a versitile decorator for adding state to stories

src/components/v2/Button/index.tsx
therealemjy: Yeah in this case it can be overwritten through the props we pass down. I originally wanted to simply assign a default value to `type`, but TypeScript complained it has to assigned as a string on the tag itself.

src/components/v2/TextField/TextField.tsx
therealemjy: So I thought the same as you, but it turns out even when you pass a `css` prop to the component, emotion will just pass down a class instead, which is why we need to pass down `className` in order to make the component's style customizable through the `css` prop.

src/components/v2/Tabs/index.tsx
therealemjy: Same thing as previous comment

src/components/v2/Tabs/index.stories.tsx
therealemjy: Yeah I saw that, but I didn't know if you wanted it to be specific to form components so I just imported this library in dev

src/components/v2/TextField/TextField.tsx
coreyar: Per the [documentation](https://emotion.sh/docs/composition) we can do this `css={[styles.button, css]}`
This isn't working? 

Emotion should take these styles add a `className` and  render a style tag on the dom. The `className` that emotion generates should be unique which is nice because it can help avoid styles being unintentionally applied to other elements

src/components/v2/Tabs/index.stories.tsx
coreyar: We can keep it as is, I meant this comment mostly as food for thought

src/components/v2/TextField/TextField.tsx
therealemjy: Unfortunately it doesn't :/
The `css` property comes out as `undefined` on the button when it's being passed down.

What you describe in your second paragraph is what's happening though. All the styles are merged together into a single class which gets applied to the component:

Screenshot 2022-03-23 at 10 20 18

Screenshot 2022-03-23 at 10 20 14

baomastr (approved):

PR #221, opened 2022-03-24
Re export configured comma number and format helper

coreyar:
- Re exports configured  `comma-number` library configure separator and decimal characters
- Adds a format function to handle setting decimal places and formatting to a string
- Removes local type for library since my PR was to DefinitelyTyped as merge and installs that type


therealemjy (approved):
Nice clean up 👍

baomastr (approved):

src/utilities/common.ts
```diff
@@ -127,6 +127,10 @@ export const formatApy = (apy?: BigNumber | string | number): string => {
   return `${apyBN.toExponential(2, 1)}%`;
 };
 
+export const boundCommaNumber = commaNumber.bindWith(',', '.');
+export const format = (bigNumber: BigNumber, dp = 2) =>
```
therealemjy, line 131: Sorry I just caught this now, but I'd suggest renaming this to something more explicit like `formatToCommaNumber`, as otherwise I feel it isn't clear what this actually does

src/components/Vault/BasicVault/CardHeader.tsx
```diff
@@ -50,7 +48,8 @@ const CardHeader = ({
         <Col className="col-item" lg={{ span: 4 }} md={{ span: 6 }} xs={{ span: 12 }}>
           <div className="title">Reward Pool</div>
           <div className="content">
-            {commaFormatter(totalPendingRewards.dp(4, 1).toFixed())} {rewardToken}
+            {formatCommaThousandsPeriodDecimal(totalPendingRewards.dp(4, 1).toFixed())}{' '}
```
baomastr, line 51: shouldn't we have a separate util for this case

baomastr: for wrapping like `toSomeFormat(totalPendingRewards)`

baomastr (approved):

therealemjy (approved):

src/components/Vault/BasicVault/CardHeader.tsx
coreyar: We could. The requirements for formatting are a varied once you get outside of setting decimals places and casting to a string. We'd probably need several functions to handle those cases. For this PR I focused on wrapping our use of `comma-number`

PR #222, opened 2022-03-24
Add collapsable mobile menu

coreyar:
- Adds a collapsable mobile menu
- Moves CoinInfo components to their own directory


src/components/v2/Layout/Header/index.tsx
```diff
@@ -21,8 +21,8 @@ export const Header = ({ pageTitle }: IHeaderProps) => {
           {pageTitle}
         </Typography>
         <Box flexDirection="row" display="flex" flex={1} justifyContent="right">
-          <XvsCoinInfo css={styles.rightItemPaper} />
-          <VaiCoinInfo css={styles.rightItemPaper} />
+          <XvsCoinInfo css={styles.rightItemPaper} className="coinInfo" />
+          <VaiCoinInfo css={styles.rightItemPaper} className="coinInfo" />
```
therealemjy, line 25: I can't find the `coinInfo` class being defined anywhere, can it be removed or am I missing something?

therealemjy (commented):
Looks good!
I spotted a few CSS issues:

The logo isn't vertically aligned with the burger and close icons:
Screen Shot 2022-03-29 at 10 46 12 AM
Screen Shot 2022-03-29 at 10 43 28 AM


The left margin should be removed from the VAI coin info bullet, while the right margin should be removed from the XVS coin info bullet:
Screen Shot 2022-03-29 at 10 43 11 AM


The spacings on the list items are a bit off:
Screen Shot 2022-03-29 at 10 48 18 AM
Screen Shot 2022-03-29 at 10 47 57 AM

src/components/v2/Layout/Header/index.tsx
coreyar: yeah this was the issue we talked about [here](https://github.com/VenusProtocol/venus-protocol-interface/pull/214#pullrequestreview-917593692) with emotion needing class to grab to process the css prop :/ I agree it is a bit confusing

coreyar: In the case of the sidebar I moved the styles to the parent in order to use the first and last  child selectors but we will still likely see classNames appear like this unless we choose a different solution

therealemjy: But what I mean is that the `coinInfo` class isn't defined anywhere in CSS, so it seems it's doing nothing here.
I'm not sure I understand the issue to be honest, since if you want to pass specific styling to these components you could pass it through the `css` property.

baomastr: rule is in CoinInfo/styles.ts:

https://github.com/VenusProtocol/venus-protocol-interface/pull/222/files#diff-015adddde90a6f30d0d7382ba71e87cd9f0b4e1de5ec709b4fb847474f095f69R7

so refers to wrong place

therealemjy: That references a property of an object, it wouldn't translate to the class name `coinInfo` (it would be formatted into a unique name).

coreyar: It is required for emotion to process the css prop. I linked the thread where we discussed this previously

therealemjy: Ok, I thought this discussion was about just passing `className` (as in the prop, not an actual string) but fair enough!

therealemjy (approved):

baomastr:
image

it would be great to resize a clickable link area inside of sidebar menu

coreyar:
@baomastr Ok I'll check that as I integrate everything into the layout

PR #223, opened 2022-03-24
Refactor Auth/Wallet Connect modal

therealemjy:
The goal of this PR is to refactor the modal that lets users connect their wallet with our dApp. Here's the breakdown:
- created an `AuthContext`, which is also responsible for rendering the auth modal itself (through its provider), so that it can be opened from anywhere within the app by simply importing and calling the `openAuthModal` from `AuthContext`. The context also exports the account address of the user (if they're connected) and the functions to handle logging in and logging out. The goal with this context is to have one central place containing all the logic related to auth.|
- refactored `ConnectedButton` so it is no longer responsible for rendering the auth modal
- updated the key used to detect if a user previously connected their wallet so it now stores the name of the connector. This is then used to both detect whether a user has connected their wallet,and display which wallet is connected in the modal:
Screenshot 2022-03-25 at 21 52 55

- updated some of the paddings used on the `Modal` component to match the designs
- created a `truncateAddress` utility to truncate addresses into a smaller format


.eslintrc.js
```diff
@@ -47,6 +47,7 @@ module.exports = {
       },
     ],
     // Custom
+    'import/no-named-as-default': 0,
```
therealemjy, line 50: @coreyar mentioned wanting to remove this rule and I second that, so here it is.

src/components/v2/AuthModal/WalletList/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+      {UPCOMING_WALLETS.map(({ name, Logo }) => (
+        <div css={styles.getListItem({ isActionable: false })} key={`upcoming-wallet-${name}`}>
+          <Logo css={styles.walletLogo} />
+
+          <Typography css={styles.walletName} component="span">
+            {name}
+          </Typography>
+
+          <Typography css={styles.comingSoonText} component="span">
```
coreyar, line 47: are these so different from out font styles that we can't use a variant?

src/components/v2/AuthModal/styles.ts
```diff
@@ -0,0 +1,10 @@
+import { css } from '@emotion/react';
```
coreyar: We could remove this until it is needed

src/components/v2/Layout/Header/ConnectButton.tsx
```diff
@@ -0,0 +1,30 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+
+import { AuthContext } from 'context/AuthContext';
+import { truncateAddress } from 'utilities/truncateAddress';
+import { SecondaryButton, IButtonProps } from '../../Button';
+
+export interface IConnectButton extends IButtonProps {
+  openAuthModal: () => void;
```
coreyar: Naming this prop `onClick` is more consistent and opens the possibility for this component to have a different interaction besides opening a modal

src/components/v2/AuthModal/WalletList/index.tsx
coreyar: If we aren't using a variant, we could use the tag directly

src/components/v2/AuthModal/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+export interface IAuthModalProps {
+  isOpen: boolean;
+  onClose: IModalProps['handleClose'];
+  onLogin: IWalletListProps['onLogin'];
+  onLogOut: IAccountDetailsProps['onLogOut'];
+  onCopyAccountAddress: IAccountDetailsProps['onCopyAccountAddress'];
+  account?: IAccountDetailsProps['account'];
+}
+
+export const AuthModal: React.FC<IAuthModalProps> = ({
```
coreyar, line 16: Since this is being rendered out side the hierarchy of the route we could even use a portal

src/components/v2/AuthModal/AccountDetails/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+            <Typography component="span" css={styles.accountAddress}>
+              {account.address}
+            </Typography>
+
+            {/* Only displayed on mobile */}
+            <Typography component="span" css={styles.accountAddressMobile}>
+              {truncatedAccountAddress}
+            </Typography>
+
+            <button
```
coreyar, line 59: What is missing that our button component doesn't handle this? This would be "action button", although I've been using the `TextButton` in this situation

src/components/v2/AuthModal/WalletList/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  onLogin: (connector: Connector) => void;
+}
+
+export const WalletList: React.FC<IWalletListProps> = ({ onLogin }) => {
+  const styles = useStyles();
+
+  return (
+    <div css={styles.container}>
+      {WALLETS.map(({ name, connector, Logo }) => (
+        <button
```
coreyar, line 21: What is missing that our button component misses this use case?

src/components/v2/AuthModal/WalletList/index.tsx
therealemjy: We do use a variant, in this case the default one: `body1`. However, this needs to be rendered as a `span` as `Typography` renders a `p` tag by default and this isn't a paragraph.

I know you don't like using `Typography`, but I believe that's how it is intended to be used. The only css applied to it via `styles.comingSoonText` is styling that isn't present in the variant (namely `color` and `text-align`).

src/components/v2/AuthModal/styles.ts
therealemjy: Yeah definitely, the file can be removed. Good catch!

src/components/v2/Layout/Header/ConnectButton.tsx
therealemjy: You're right 👍

src/components/v2/AuthModal/index.tsx
therealemjy: Using a portal there is a valid point, but wouldn't we want the `Modal` component to be responsible of that?

src/components/v2/AuthModal/AccountDetails/index.tsx
therealemjy: It is very different from our `button` component. The `button` tag is used semantically here, but UI wise this is what it looks like:
Screen Shot 2022-03-28 at 4 47 10 PM

src/components/v2/AuthModal/WalletList/index.tsx
therealemjy: Our button component would never cover this case. The `button` tag is used semantically here, this is what it looks like in the UI:
Screen Shot 2022-03-28 at 4 47 59 PM

src/components/v2/AuthModal/WalletList/index.tsx
coreyar: I don't have any problem using `Typography` and understand its purpose. I would wrap our default variants in their own components to make them less verbose.

therealemjy: Yeah that works for me 👍

src/components/v2/AuthModal/index.tsx
coreyar: yes, I think that makes sense

src/components/v2/AuthModal/WalletList/index.tsx
coreyar: It could be written with the two icons and text positioned but I agree it would require additional styles. The advantage to using our button component is interactive styles would be consistently applied

coreyar (approved):

src/components/v2/AuthModal/WalletList/index.tsx
therealemjy: The problem is that in this case they don't share any styling, including the `hover` and `active` styles.
For other interactive styles, like the `outline` property, these should be defined globally for the whole project in my opinion.

PR #225, opened 2022-03-24
Replace process.env.REACT_APP_CHAIN_ID with CHAIN_ID from config

therealemjy:
Now that we've introduced a `CHAIN_ID` constant in the config, we can reference it from everywhere in the app without using the environment variable directly.


src/containers/App.tsx
```diff
@@ -79,10 +80,10 @@ class App extends React.Component {
                             <Route exact path="/vote/leaderboard" component={VoterLeaderboard} />
                             <Route exact path="/vote/proposal/:id" component={VoteOverview} />
                             <Route exact path="/vote/address/:address" component={ProposerDetail} />
-                            {process.env.REACT_APP_CHAIN_ID === '97' && (
+                            {CHAIN_ID === BscChainId.TESTNET && (
```
baomastr: maybe it would be useful to add util like `isBscChain`?

src/containers/Layout/Sidebar.tsx
```diff
@@ -553,7 +554,7 @@ function Sidebar({ history, setSetting }: SidebarProps) {
               Forum
             </Label>
           </Option>
-          {process.env.REACT_APP_CHAIN_ID === '97' && (
+          {CHAIN_ID === BscChainId.TESTNET && (
```
baomastr: so it could be used overall

baomastr (approved):

coreyar (approved):

coreyar:
Looks like a race condition with the storybook is showing up here for the vault page

src/containers/App.tsx
therealemjy: Yeah good point!
I've added a `isOnTestnet` variable to the config (`isBscChain` would be always `true`, since both the mainnet and testnet networks are BSC).

PR #227, opened 2022-03-25
Add modal wrapper component

baomastr:


src/components/v2/Modal/Modal.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { ReactElement } from 'react';
+import Backdrop from '@mui/material/Backdrop';
+import { Button, Modal as MUIModal, ModalProps } from '@mui/material';
+import Fade from '@mui/material/Fade';
+import Box from '@mui/material/Box';
+import { useModalStyles } from './ModalStyles';
+
+export interface IModalProps extends ModalProps {
+  children: any;
```
therealemjy: You don't need to specify this prop, since you're extending `React.FC` it will already be included

src/components/v2/Modal/Modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+  titleComponent,
+}) => {
+  const s = useModalStyles({ hasTitleComponent: Boolean(titleComponent) });
+  const closeModal = () => {
+    handleClose();
+  };
+  return (
+    <MUIModal
+      open={isOpened}
+      onClose={closeModal}
```
therealemjy: You can pass `handleCloseModal` directly

src/components/v2/Modal/Modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+                  fill="white"
+                />
+                <rect
+                  x="1.41431"
+                  width="18"
+                  height="2"
+                  transform="rotate(45 1.41431 0)"
+                  fill="white"
+                />
+              </svg>
```
therealemjy: I'd add this as an icon and use the `Icon` component

src/components/v2/Modal/Modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+import Fade from '@mui/material/Fade';
+import Box from '@mui/material/Box';
+import { useModalStyles } from './ModalStyles';
+
+export interface IModalProps extends ModalProps {
+  children: any;
+  className?: string;
+  isOpened: boolean;
+  handleClose: () => void;
+  titleComponent?: ReactElement;
```
therealemjy: Nice. Although I'd just rename this prop to `title`, since it isn't actually a component (otherwise someone might expect they need to pass an actual component rather than just DOM).

src/components/v2/Modal/Modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+      closeAfterTransition
+      BackdropComponent={Backdrop}
+      BackdropProps={{
+        timeout: 500,
+      }}
+    >
+      <Fade in={isOpened}>
+        <Box sx={s.box}>
+          <Box sx={s.titleWrapper}>
+            <p style={s.titleComponent}>{titleComponent}</p>
```
therealemjy: Let's render the title directly without any wrapper it in a paragraph, since sometimes we'll want to display more than text, like for the supply modal:
Screenshot 2022-03-25 at 10 14 40

src/components/v2/Modal/ModalStyles.ts
```diff
@@ -0,0 +1,54 @@
...
+  const theme = useTheme();
+  return {
+    box: {
+      position: 'absolute' as const,
+      top: '50%',
+      left: '50%',
+      transform: 'translate(-50%, -50%)',
+      width: 400,
+      bgColor: 'background.paper',
+      border: '2px solid #000',
```
therealemjy: There's no border in the design, so we should get rid of that.

src/components/v2/Modal/ModalStyles.ts
```diff
@@ -0,0 +1,54 @@
...
+      minWidth: iconCloseSize,
+      marginLeft: 'auto',
+      padding: 0,
+    },
+    contentWrapper: {
+      padding: '24px',
+      paddingTop: 0,
+    },
+  };
+};
```
therealemjy: This file will need to be re-written to use css interpolation.

Also, margins and paddings need to be expressed using the theme `spacing`.

src/components/v2/Modal/ModalStyles.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const useModalStyles = ({ hasTitleComponent }: { hasTitleComponent: boolean }) => {
+  const theme = useTheme();
+  return {
+    box: {
+      position: 'absolute' as const,
+      top: '50%',
+      left: '50%',
+      transform: 'translate(-50%, -50%)',
+      width: 400,
```
therealemjy: Having a fixed width will break the responsiveness

src/components/v2/Modal/Modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+      onClose={closeModal}
+      onBackdropClick={closeModal}
+      closeAfterTransition
+      BackdropComponent={Backdrop}
+      BackdropProps={{
+        timeout: 500,
+      }}
+    >
+      <Fade in={isOpened}>
+        <Box sx={s.box}>
```
therealemjy: I'd pass `className` to this tag, so that its width could be modified through CSS for example.

src/components/v2/Modal/ModalStyles.ts
```diff
@@ -0,0 +1,54 @@
...
+      top: 0,
+      position: 'absolute',
+      height: iconCloseSize,
+      width: iconCloseSize,
+      minWidth: iconCloseSize,
+      marginLeft: 'auto',
+      padding: 0,
+    },
+    contentWrapper: {
+      padding: '24px',
```
therealemjy: I think we should have an optional prop like `noHorizontalPadding` that can be passed to the component to remove the left and right paddings, otherwise we'll have issues building this modal:
Screenshot 2022-03-25 at 10 33 01

therealemjy (changes requested):

src/components/v2/Modal/Modal.tsx
baomastr: title prop has conflict with ModalProps. so i defined titleComponent here. renamed to `modalTitle`

src/components/v2/Modal/Modal.tsx
baomastr: changed paragraph to Box component. cause wrapper is necessary for styling

src/components/v2/Modal/Modal.tsx
therealemjy: You're not expanding the modal props onto the modal component, so you can either get rid of the extend altogether or omit the `title` prop from it:
```typescript
export interface IModalProps extends Omit<ModalProps, 'title> {
  ...
}
```

src/components/v2/Modal/ModalStyles.ts
```diff
@@ -0,0 +1,61 @@
...
+      padding: 0,
+      borderRadius: theme.spacing(3),
+      paddingTop: theme.spacing(3),
+    },
+    titleWrapper: {
+      position: 'relative',
+      paddingLeft: theme.spacing(3),
+      paddingRight: theme.spacing(3),
+      paddingBottom: hasTitleComponent ? theme.spacing(3) : 0,
+      borderBottom: hasTitleComponent ? `1px solid ${theme.palette.secondary.dark}` : 0,
```
therealemjy:
```suggestion
      borderBottom: hasTitleComponent ? `1px solid ${theme.palette.secondary.light}` : 0,
```

Screenshot 2022-03-25 at 13 02 00

src/components/v2/Modal/ModalStyles.ts
```diff
@@ -0,0 +1,61 @@
...
+      width: iconCloseSize,
+      minWidth: iconCloseSize,
+      marginLeft: 'auto',
+      padding: 0,
+    },
+    contentWrapper: {
+      padding: theme.spacing(3),
+      paddingLeft: noHorizontalPadding ? 0 : theme.spacing(3),
+      paddingRight: noHorizontalPadding ? 0 : theme.spacing(3),
+      paddingTop: 0,
```
therealemjy:
```suggestion
      padding: theme.spacing(0, noHorizontalPadding ? 0 : 5, 5, noHorizontalPadding ? 0 : 5),
```

Screenshot 2022-03-25 at 13 06 41

baomastr: for me, it's easier to read current padding props

therealemjy (approved):

PR #228, opened 2022-03-25
Add warning that UST cannot be deposited on binance

coreyar:
Users are trying to deposit UST on binance but it isn't supported. Add warning that this is not possibl
Screen Shot 2022-03-25 at 11 05 04
e

### DIFF
--- a/src/containers/Layout/MainLayout.tsx
+++ b/src/containers/Layout/MainLayout.tsx
@@ -1,5 +1,7 @@
 import React from 'react';
+import Typography from '@mui/material/Typography';
 import styled, { withTheme } from 'styled-components';
+
 import Sidebar from 'containers/Layout/Sidebar';
 import Header from 'containers/Layout/Header';
 import { Footer } from 'components/v2/Layout/Footer';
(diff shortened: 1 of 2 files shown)


src/containers/Layout/MainLayout.tsx
```diff
@@ -61,6 +74,15 @@ function MainLayout({ title = '', isHeader, children }: Props) {
 
   return (
     <MainLayoutWrapper>
+      <div className="ust-warning">
+        <p>
+          <span role="img" aria-label="warning">
+            ⚠️
+          </span>{' '}
+          Wormhole UST on Venus is not recognized by CEXs. Please convert it to WrappedUST before
+          using.
```
kkirka:
```suggestion
          Wormhole UST on Venus is <strong>not</strong> recognized by CEXs. Please swap it to WrappedUST
          before depositing on a CEX.
```

therealemjy: Just pushed an update for this:
Screenshot 2022-03-25 at 17 52 53

src/containers/Layout/MainLayout.tsx
```diff
@@ -12,13 +14,13 @@ const MainLayoutWrapper = styled.div`
   background-color: var(--color-bg-main);
 
   .main {
-    height: 100vh;
+    height: calc(100vh - 29px);
```
therealemjy, line 17: Unfortunately this is needed because of how the current layout uses fixed heights... Thankfully when we introduce the new layout we won't have to do stuff like this 😅

src/containers/Layout/Sidebar.tsx
```diff
@@ -23,7 +23,7 @@ import { useComptroller, useVaiToken } from '../../hooks/useContract';
 import { getVaiVaultAddress } from '../../utilities/addressHelpers';
 
 const SidebarWrapper = styled.div`
-  height: 100vh;
+  height: calc(100vh - 29px);
```
therealemjy, line 26: Same thing here about fixed heights being used..

therealemjy (approved):

PR #229, opened 2022-03-28
 Fix modal with scrollable content

baomastr:
Fix for scrollable modal

### DIFF
--- a/src/components/v2/Modal/styles.ts
+++ b/src/components/v2/Modal/styles.ts
@@ -18,19 +18,23 @@ export const useModalStyles = ({
       top: 50%;
       left: 50%;
       transform: translate(-50%, -50%);
-      width: 80%;
+      width: calc(100% - ${theme.spacing(4)});
       max-width: 544px;
       border-radius: ${theme.spacing(3)};
-      padding-top: ${theme.spacing(3)};
       background-color: ${theme.palette.background.paper};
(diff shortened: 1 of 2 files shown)


src/components/v2/Modal/styles.ts
```diff
@@ -21,16 +21,20 @@ export const useModalStyles = ({
       width: 80%;
       max-width: 544px;
       border-radius: ${theme.spacing(3)};
-      padding-top: ${theme.spacing(3)};
       background-color: ${theme.palette.background.paper};
+      overflow: scroll;
+      max-height: 100%;
```
therealemjy: Nice one 👍

I have two suggestions:
```suggestion
      overflow: auto;
      max-height: max-height: calc(100% - ${theme.spacing(4)});
```

And I'd also change the width property:
```css
width: calc(100% - ${theme.spacing(4)});
```

Doing that means that on mobile on the spacing between the modal and the borders of the screens will always be 16px, which is the spacing used in the designs for the cards:
Screen Shot 2022-03-28 at 10 44 22 AM

This is what the modal looks like on mobile with these css properties:
Screen Shot 2022-03-28 at 10 39 00 AM

therealemjy (commented):
Really nice fix 👍
I just had one comment

therealemjy (approved):

PR #230, opened 2022-03-28
Add my account component

baomastr:
image

### DIFF
--- a/src/components/v2/Toggle/index.tsx
+++ b/src/components/v2/Toggle/index.tsx
@@ -1,19 +1,23 @@
 /** @jsxImportSource @emotion/react */
 import React from 'react';
 import Switch from '@mui/material/Switch';
+import { SwitchBaseProps } from '@mui/material/internal/SwitchBase';
+
 import { useStyles } from './styles';
 
 export interface IToggleProps {
-  onChange: (event: React.ChangeEvent<HTMLInputElement>, checked: boolean) => void;
+  onChange: SwitchBaseProps['onChange'];
   value: boolean;
+  className?: string;
 }
 
 const label = { inputProps: { 'aria-label': 'Switch' } };
 
-export const Toggle = ({ onChange, value }: IToggleProps) => {
+export const Toggle = ({ onChange, value, className }: IToggleProps) => {
   const styles = useStyles();
   return (
     <Switch
+      className={className}
       css={styles}
       focusVisibleClassName=".Mui-focusVisible"
       disableRipple
--- a/src/components/v2/Toggle/styles.ts
+++ b/src/components/v2/Toggle/styles.ts
@@ -1,5 +1,5 @@
-import { useTheme } from '@mui/material';
 import { css } from '@emotion/react';
+import { useTheme } from '@mui/material';
 
 export const useStyles = () => {
   const theme = useTheme();
--- a/src/components/v2/Slider/index.tsx
+++ b/src/components/v2/Slider/index.tsx
@@ -3,7 +3,7 @@ import React, { useState } from 'react';
 import MaterialSlider from '@mui/material/Slider';
 import { useStyles } from './styles';
 
-interface ISliderProps {
+export interface ISliderProps {
   defaultValue: number;
   mark: number;
   step: number;
@@ -11,6 +11,7 @@ interface ISliderProps {
   min: number;
   max: number;
   onChange: (value: number | number[]) => void;
+  isDisabled?: boolean;
 }
 
 export const Slider = ({
@@ -21,6 +22,7 @@ export const Slider = ({
   min,
   max,
   onChange,
+  isDisabled,
 }: ISliderProps) => {
   const marks = mark ? [{ value: mark }] : undefined;
   const [over, setOver] = useState(defaultValue > mark);
@@ -43,6 +45,7 @@ export const Slider = ({
       min={min}
       max={max}
       size="medium"
+      disabled={isDisabled}
     />
   );
 };
--- a/src/components/v2/Slider/styles.ts
+++ b/src/components/v2/Slider/styles.ts
@@ -8,6 +8,9 @@ export const useStyles = ({ over }: { over: boolean }) => {
     background-color: ${theme.palette.background.default};
     height: 8px;
     padding: 0;
+    &.Mui-disabled {
+      color: ${over ? theme.palette.interactive.error50 : theme.palette.interactive.success};
+    }
     .MuiSlider-track {
       height: 8px;
     }
--- a/src/components/v2/Tooltip/Tooltip.tsx
+++ b/src/components/v2/Tooltip/Tooltip.tsx
@@ -11,7 +11,7 @@ export const Tooltip = ({ children, ...rest }: ITooltipProps) => {
     <>
       <Global styles={styles} />
       <MuiTooltip arrow placement="top" {...rest}>
-        {children}
+        <span>{children}</span>
       </MuiTooltip>
     </>
   );
--- a/src/pages/Dashboard/Markets/BorrowMarket.tsx
+++ b/src/pages/Dashboard/Markets/BorrowMarket.tsx
@@ -2,7 +2,7 @@
 import React from 'react';
 import {
   formatCoinsToReadableValue,
-  formatDollarsToReadableValue,
+  formatCentsToReadableValue,
   formatApy,
 } from 'utilities/common';
 import { Asset, TokenSymbol } from 'types';
@@ -57,7 +57,7 @@ export const BorrowMarketUi: React.FC<IBorrowMarketUiProps> = ({
     },
     {
       key: 'liquidity',
-      render: () => formatDollarsToReadableValue(asset.liquidity),
+      render: () => formatCentsToReadableValue(asset.liquidity),
       value: asset.liquidity.toString(),
     },
   ]);
--- a/src/pages/Dashboard/MyAccount/index.stories.tsx
+++ b/src/pages/Dashboard/MyAccount/index.stories.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import { ComponentMeta } from '@storybook/react';
+import { withThemeProvider, withCenterStory } from 'stories/decorators';
+import { MyAccount } from '.';
+
+export default {
+  title: 'Components/Dashboard/MyAccount',
+  component: MyAccount,
+  decorators: [withThemeProvider, withCenterStory({ width: 650 })],
+} as ComponentMeta<typeof MyAccount>;
+
+export const Default = () => (
+  <MyAccount
+    netApyPercentage={2493}
+    dailyEarningsCents={14829}
+    supplyBalanceCents={122889}
+    borrowBalanceCents={30243}
+    borrowLimitCents={7373}
+    borrowLimitUsedPercentage={41}
+    safeLimitPercentage={41}
+  />
+);
--- a/src/pages/Dashboard/MyAccount/index.tsx
+++ b/src/pages/Dashboard/MyAccount/index.tsx
@@ -1,0 +1,120 @@
+/** @jsxImportSource @emotion/react */
+import React, { useState } from 'react';
+import Box from '@mui/material/Box';
+import Paper from '@mui/material/Paper';
+import Typography from '@mui/material/Typography';
+
+import { formatCentsToReadableValue } from 'utilities/common';
+import { Slider } from '../../../components/v2/Slider';
+import { useMyAccountStyles as useStyles } from './styles';
+import { IToggleProps, Toggle } from '../../../components/v2/Toggle';
+import { Icon } from '../../../components/v2/Icon';
+import { Tooltip } from '../../../components/v2/Tooltip';
+
+interface IMyAccountProps {
+  netApyPercentage: number;
+  dailyEarningsCents: number;
+  supplyBalanceCents: number;
+  borrowBalanceCents: number;
+  borrowLimitCents: number;
+  safeLimitPercentage: number;
+  borrowLimitUsedPercentage: number;
+}
+
+export const MyAccount = ({
+  netApyPercentage,
+  dailyEarningsCents,
+  supplyBalanceCents,
+  borrowBalanceCents,
+  borrowLimitCents,
+  safeLimitPercentage,
+}: IMyAccountProps) => {
+  const [isToggleSwitched, setToggleSwitched] = useState(true);
+  const handleSwitch: IToggleProps['onChange'] = (event, checked) => {
+    setToggleSwitched(checked);
+  };
+  const styles = useStyles();
+  return (
+    <Box css={styles.root} component={Paper}>
+      <div css={styles.row}>
+        <Typography variant="h4">My account</Typography>
+        <Typography component="span" variant="small2" css={styles.apyWithXvs}>
+          <Tooltip css={styles.tooltip} title="tooltip content">
+            <Icon css={styles.iconInfo} name="info" />
+          </Tooltip>
+          <Typography color="text.primary" variant="small1">
+            APY with XVS
+          </Typography>
+          <Toggle css={styles.toggle} value={isToggleSwitched} onChange={handleSwitch} />
+        </Typography>
+      </div>
+      <div>
+        <Typography component="p" variant="small2" css={styles.netApyLabel}>
+          Net APY
+          <Tooltip css={styles.tooltip} title="tooltip content">
+            <Icon css={styles.iconInfo} name="info" />
+          </Tooltip>
+        </Typography>
+        <Typography variant="h1" color="interactive.success">
+          {netApyPercentage}%
+        </Typography>
+      </div>
+      <ul css={styles.list}>
+        <Typography component="li" variant="h4" css={styles.item}>
+          <Typography component="p" variant="small2" css={styles.labelListItem}>
+            Daily earnings
+          </Typography>
+          {formatCentsToReadableValue(dailyEarningsCents)}
+        </Typography>
+        <Typography component="li" variant="h4" css={styles.item}>
+          <Typography component="p" variant="small2" css={styles.labelListItem}>
+            Supply balance
+          </Typography>
+          {formatCentsToReadableValue(supplyBalanceCents)}
+        </Typography>
+        <Typography component="li" variant="h4" css={styles.item}>
+          <Typography component="p" variant="small2" css={styles.labelListItem}>
+            Borrow balance
+          </Typography>
+          {formatCentsToReadableValue(borrowBalanceCents)}
+        </Typography>
+      </ul>
+      <div css={styles.row}>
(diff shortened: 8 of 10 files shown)


src/components/v2/MyAccount/MyAccount.tsx
```diff
@@ -0,0 +1,80 @@
...
+      </div>
+      <div>
+        <p css={[styles.label, styles.netApyLabel]}>
+          Net APY <Icon css={styles.iconInfo} name="info" />
+        </p>
+        <span css={styles.netApy}>{netApy}</span>
+      </div>
+      <ul css={styles.list}>
+        <li css={styles.item}>
+          <p css={[styles.label, styles.labelListItem]}>Daily earnings</p>
```
baomastr: how do you think, guys @coreyar @therealemjy, should we move texts in separate json config for translating feature or just resources organizing?

src/components/v2/MyAccount/MyAccount.tsx
```diff
@@ -0,0 +1,106 @@
...
+  apyWithXvs: 'APY with XVS',
+  apyWithXvsTooltip: 'tooltip content',
+  netApy: 'Net APY',
+  dailyEarnings: 'Daily earnings',
+  supplyBalance: 'Supply balance',
+  borrowBalance: 'Borrow balance',
+  borrowLimit: 'Borrow limit:',
+  safeLimit: 'Your safe limit:',
+};
+
```
therealemjy: This is confusing in my opinion. Since we've said we wouldn't handle internationalization for now, I say let's just have the texts hardcoded in the DOM directly.

src/components/v2/MyAccount/MyAccount.tsx
```diff
@@ -0,0 +1,106 @@
...
+interface IMyAccountProps {
+  netApy: string;
+  dailyEarnings: string;
+  supplyBalance: string;
+  borrowBalance: string;
+  borrowLimit: string;
+  borrowLimitSum: string;
+  safeLimit: string;
+  sliderProps: ISliderProps;
```
therealemjy: We'll want to pass the actual values to the component, so `BigNumber` values instead of strings.
These values will then be used to render the slider, so we shouldn't need to pass `sliderProps` at all.

src/components/v2/MyAccount/MyAccount.tsx
```diff
@@ -0,0 +1,106 @@
...
+  dailyEarnings,
+  supplyBalance,
+  borrowBalance,
+  borrowLimit,
+  borrowLimitSum,
+  safeLimit,
+  sliderProps,
+}: IMyAccountProps) => {
+  const [isToggleSwitched, setToggleSwitched] = useState(true);
+  const handleSwitch = (event: React.ChangeEvent<HTMLInputElement>, checked: boolean) => {
```
therealemjy: I'd suggest importing the types of the `Toggle` component and assigning the type of its `onChange` property to `handleSwitch`.

src/components/v2/MyAccount/MyAccount.tsx
```diff
@@ -0,0 +1,106 @@
...
+          {borrowBalance}
+        </li>
+      </ul>
+      <div css={[styles.row, styles.borrowLimit]}>
+        <div>
+          <p css={[styles.label, styles.inline]}>{myAccountResources.borrowLimit}</p> {borrowLimit}
+        </div>
+        {borrowLimitSum}
+      </div>
+      <Slider {...sliderProps} />
```
therealemjy: It seems the green bar inside the slider can be moved by clicking anywhere on the slider, but it should be fixed.

src/components/v2/MyAccount/MyAccount.tsx
```diff
@@ -0,0 +1,106 @@
...
+    setToggleSwitched(checked);
+  };
+  const styles = useStyles();
+  return (
+    <Box css={styles.root} component={Paper}>
+      <div css={styles.row}>
+        {myAccountResources.myAccountTitle}
+        <span css={styles.apyWithXvs}>
+          <Tooltip css={styles.tooltip} title={myAccountResources.apyWithXvsTooltip}>
+            <span>
```
therealemjy: Just curious; what do you need the `span` there for?

src/components/v2/Toggle/index.tsx
```diff
@@ -1,19 +1,23 @@
+/** @jsxImportSource @emotion/react */
 import React from 'react';
 import Switch from '@mui/material/Switch';
+import { SerializedStyles } from '@emotion/react';
 import { useStyles } from './styles';
 
 export interface IToggleProps {
   onChange: (event: React.ChangeEvent<HTMLInputElement>, checked: boolean) => void;
   value: boolean;
+  cssStyles?: SerializedStyles;
```
therealemjy:
```suggestion
  className?: string;
```

You'll then need to apply the `className` prop to the `Switch` component tag.

It is counterintuitive, but passing the `className` means components can pass a `css` prop directly to `Toggle` and it will be applied to it. You can check the `Tabs` component (`src/components/v2/Tabs/index.tsx`) to see an example.

src/components/v2/MyAccount/MyAccountStyles.ts
```diff
@@ -0,0 +1,105 @@
...
+      display: flex;
+      justify-content: flex-end;
+      align-items: center;
+    `,
+    safeLimit: css`
+      margin-left: ${theme.spacing(1)};
+      margin-right: ${theme.spacing(1)};
+    `,
+  };
+};
```
therealemjy: Let's follow the codebase's best practices and rename this file to `styles.ts`

src/components/v2/MyAccount/index.ts
```diff
@@ -0,0 +1 @@
+export * from './MyAccount';
```
therealemjy: Let's follow the codebase's best practices and move the component inside this file directly and delete `MyAccount.tsx`.

src/components/v2/MyAccount/MyAccountStyles.ts
```diff
@@ -0,0 +1,105 @@
...
+      padding: ${theme.spacing(3)};
+    `,
+    row: css`
+      width: 100%;
+      display: flex;
+      justify-content: space-between;
+      align-items: center;
+    `,
+    apyWithXvs: css`
+      font-size: 14px;
```
therealemjy: Why not use the `Typography` component with our theme variant to render this text?

src/components/v2/MyAccount/MyAccountStyles.ts
```diff
@@ -0,0 +1,105 @@
...
+    iconInfo: css`
+      margin-left: ${theme.spacing(1)};
+      margin-right: ${theme.spacing(1)};
+    `,
+    toggle: css`
+      margin-left: ${theme.spacing(1)};
+    `,
+    label: css`
+      font-size: 14px;
+      letter-spacing: 0.16px;
```
therealemjy: Same as previous comment about using the `Typography` component.

src/components/v2/MyAccount/MyAccountStyles.ts
```diff
@@ -0,0 +1,105 @@
...
+        border-left: none;
+        padding-left: 0;
+      }
+      :last-child {
+        border-right: none;
+        padding-right: 0;
+      }
+    `,
+    borrowLimit: css`
+      font-size: 14px;
```
therealemjy: Same as previous comment about using the `Typography` component.

src/components/v2/MyAccount/MyAccountStyles.ts
```diff
@@ -0,0 +1,105 @@
...
+    `,
+    bottom: css`
+      font-size: 14px;
```
therealemjy: Same as previous comment about using the `Typography` component.

src/components/v2/MyAccount/MyAccountStyles.ts
```diff
@@ -0,0 +1,105 @@
+import { css } from '@emotion/react';
+import { useTheme } from '@mui/material';
+
+export const useMyAccountStyles = () => {
+  const theme = useTheme();
+  return {
+    root: css`
+      font-size: 20px;
```
therealemjy: Since only the title uses a font size of 20px, I'd suggest not applying it to the whole tag but instead wrap the title inside a `Typography` component and apply the corresponding variant to it (`h4` I believe).

src/components/v2/MyAccount/MyAccount.tsx
baomastr: > Tooltip is not working properly because the child of a Material-UI Tooltip must be able to hold a ref.

https://stackoverflow.com/questions/57527896/material-ui-tooltip-doesnt-display-on-custom-component-despite-spreading-props


going to move wrapper for children directly into tooltip component

src/components/v2/MyAccount/MyAccount.tsx
baomastr: could you explain in more details?  
you suggest to pass callback for state management like setToggleSwitched into Toggle onChange? or I missed how should we manage component state

src/components/v2/MyAccount/MyAccount.tsx
baomastr: added disabled prop

src/components/v2/Toggle/index.tsx
baomastr: but what's wrong with passing css as SerializedStyles?

src/components/v2/MyAccount/MyAccount.tsx
baomastr: shouldn't we in order of dumb components calculate inside of parent hook?

src/components/v2/MyAccount/MyAccount.tsx
therealemjy: I see! Thanks for the info :)

src/components/v2/MyAccount/MyAccount.tsx
therealemjy: For sure. Your implementation is correct, I just mentioned you could reuse the type of `IToggleProps`:
```
const handleSwitch: IToggleProps['onChange'] = (event, checked) => {
  ...
}
```

This way if the type of the `onChange` property of the `Toggle` component changes, `MyAccount` will use the updated type automatically.

src/components/v2/MyAccount/MyAccount.tsx
therealemjy: Calculations are fine inside a dumb component. We call them dumb because they work as a black box: if you always pass the same props to a dumb component, it will always render the same thing.
The container will be in charge of fetching this data and handling any other side effect needed.

src/components/v2/MyAccount/MyAccount.tsx
baomastr: oh. misunderstood. sure. gonna add typing

src/components/v2/MyAccount/MyAccount.tsx
baomastr: requested calculations description from Sean.
I can make it in separate task in order to merge this one. what do you think of this?

therealemjy: I saw that, but I'm not sure what calculations you're talking about. Your current implementation is very close to what I think we should do, except I wouldn't use strings to express the values.
I mentioned we should use `BigNumber` in my first component because I'm used to it with cryptos, but I just realized that we're only displaying values in dollars on this component so using numbers only would work.

My point is we should pass the values directly, then pass these to the slider (rather than having `sliderProps`.
Something like this for example:
```
netApyPercentage: number;
dailyEarningsCents: number;
supplyBalanceCents: number;
borrowBalanceCents: number;
borrowLimitCents: number;
safeLimitPercentage: number;
borrowLimitUsedPercentage: number;
```

Then you don't need to pass `sliderProps`, as you've got all the values you need for it already.

baomastr: yep. but there should be extra wrappers for mapping to format that awaits the frontend like `${formatToDollars(dailyEarningsCents)}`

baomastr: okay. I'll use numbers format for props in this component. but still need calculations for passing into the slider component. so can't do it without formulas from thread in slack.
so we can merge this one in order to add calculations within the next task

added TODO here

therealemjy: About the `formatToDollars`, you can create this util in this PR :) It will be relevant to a lot of other parts of the app too.

I'm sorry but what calculations are you thinking about?
The slider only displays the current borrow balance, the safe limit and the borrow limit. You can then use these values to define the props that get passed to the slider:
```
const min = 0;
const max = borrowLimitCents;
const mark = Math.round(borrowLimitCents * safeLimitPercentage / 100);

<Slider min={min} max={max} mark={mark}  />
```

Is there something missing here?

baomastr: there can be also a react component, I think:

https://github.com/VenusProtocol/venus-protocol-interface/pull/244/files#diff-6fcef2bd92cbf1c6b895efaa8c8eda1c978cd54f7a5f4808037c2ba75d0658f7R25

so we just pass into dumb component already prepared strings/components. and it allows us to make more clear atoms and molecules. all calculations and formatting will be made in parent container.

therealemjy: Why would `borrowBalance` be a component? We don't need to make the `MyAccount` component ultra modular, it only needs to have just enough flexibility to render exactly like in the designs.

And about the calculations, again there's no calculations there. We only pass the values to the component, which then renders them.
If you're referring to how we're fetching and calculating the values that are passed to `MyAccount`, that will indeed be done in its parent container, but you can avoid having `sliderProps` as a prop on `MyAccount`.

baomastr: yeah. you are right about modularity. 
I thought about separate layers for ui and for side effects in order to split business logic from ui

src/components/v2/Toggle/index.tsx
therealemjy: Nothing's wrong about it, you can just benefit from this feature without needing a custom `cssStyles` prop, since `className` is already a React prop

src/components/v2/Toggle/index.tsx
```diff
@@ -1,19 +1,24 @@
...
 
 const label = { inputProps: { 'aria-label': 'Switch' } };
 
-export const Toggle = ({ onChange, value }: IToggleProps) => {
+export const Toggle = ({ onChange, value, cssStyles }: IToggleProps) => {
   const styles = useStyles();
   return (
     <Switch
       sx={styles}
+      css={cssStyles}
```
coreyar: We could clean this up by moving the sx styles to the css prop styles so styles are applied uniformly

coreyar: We could clean this up by moving the sx styles to the css prop styles so styles are applied uniformly

src/components/v2/Toggle/index.tsx
coreyar: I think either approach is valid, but we should be consistent. Currently we have been passing the className and not the css object

coreyar: I think either approach is valid, but we should be consistent. Currently we have been passing the className and not the css object

src/components/v2/MyAccount/MyAccount.tsx
baomastr: fixed UI props

src/components/v2/MyAccount/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { useMyAccountStyles as useStyles } from './styles';
+import { IToggleProps, Toggle } from '../Toggle';
+import { Icon } from '../Icon';
+import { Tooltip } from '../Tooltip';
+
+function toPercentageString(n: number) {
+  return `${n}%`;
+}
+function toUsdString(n: number) {
+  return `$${n}`;
```
therealemjy: This util should also divide the value passed by 100, to convert it into dollars (since you're passing values in cents to it).

src/components/v2/MyAccount/index.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+  decorators: [withThemeProvider, withCenterStory({ width: 'calc(100 - 32px)' })],
+} as ComponentMeta<typeof MyAccount>;
+
+export const MyAccountDefault = () => (
+  <MyAccount
+    netApyPercentage={24.93}
+    dailyEarningsCents={14.829}
+    supplyBalanceCents={122.889}
+    borrowBalanceCents={30.243}
+    borrowLimitCents={73.73}
```
therealemjy: The values are in cents, not in dollars:
```suggestion
    dailyEarningsCents={14829}
    supplyBalanceCents={122889}
    borrowBalanceCents={30243}
    borrowLimitCents={7373}
```

therealemjy (commented):
Just left two small comments, we're very close to being ready to merge this in.

src/components/v2/Toggle/index.tsx
```diff
@@ -1,19 +1,23 @@
...
 
 const label = { inputProps: { 'aria-label': 'Switch' } };
 
-export const Toggle = ({ onChange, value }: IToggleProps) => {
+export const Toggle = ({ onChange, value, parentStyles }: IToggleProps) => {
   const styles = useStyles();
   return (
     <Switch
-      sx={styles}
+      css={[styles, parentStyles]}
```
therealemjy: I think you're getting confused with the styling.
You don't need to pass any css prop in order to extend the styles of the component. Instead, you can just pass the `className` prop to the component. This will have the exact same result:

```typescript
export interface IToggleProps {
  onChange: SwitchBaseProps['onChange'];
  value: boolean;
  className?: string;
}

export const Toggle = ({ onChange, value, className }: IToggleProps) => {
  ...

  return (
    <Switch 
      css={styles}
      className={className}
      ...
```

Doing that, your component can still get its styles extended, except it can be done through the already supported `css` prop:

```typescript
<Toggle css={yourCustomStyle} />
```

I understand it's confusing, it's an Emotion thing :/

therealemjy (approved):

src/components/v2/Toggle/styles.ts
```diff
@@ -1,8 +1,9 @@
+import { css } from '@emotion/react';
 import { useTheme } from '@mui/material';
 
 export const useStyles = () => {
   const theme = useTheme();
-  return {
+  return css({
```
coreyar: Rather then wrap this in css we should rewrite it to be an interpolated string, that way it is lintable and more readable

coreyar: I have done this here - https://github.com/VenusProtocol/venus-protocol-interface/pull/265

therealemjy: I've approved and merged your PR

PR #235, opened 2022-03-29
Fix: layout height with banner

baomastr:
Fixed layout height depending on new banner


src/containers/Layout/MainLayout.tsx
```diff
@@ -15,13 +15,13 @@ const MainLayoutWrapper = styled.div`
   background-color: var(--color-bg-main);
 
   .main {
-    height: calc(100vh - 29px);
+    height: calc(100vh - 56px);
```
coreyar, line 18: Ideally this would be a variable so that if the high changes again we can avoid issues like this. I can handle this when adding the banner to the new layout

coreyar (approved):

PR #236, opened 2022-03-29
Build "Mint VAI" section

therealemjy:
- Create `TokenTextField` component, which is a `TextField` component that specifically handles token amounts. It uses `BigNumber` values, expressed in wei so they can be used with contracts without the need for any conversion. This component is very common in our designs and since it's a very important piece of the app I've covered it with tests
Screen Shot 2022-03-29 at 10 15 43 PM

- Add support for [react-testing-library](https://testing-library.com/)
- Create `renderComponent` test utility which renders and wraps any DOM passed to it within the same providers and containers as for the app, so that the test environment is similar to our app environment
- Update `withCenterStory` decorator to make it responsive
- And environments variables to GitHub actions at workflow level
- Create `MintRepayVai` component
Screen Shot 2022-03-29 at 10 15 35 PM

Note that the "Repay VAI" tab is missing. I'm still waiting on the designs to be complete in order to work on it, but since this PR is already getting quite big I'll anyway have to do that in a separate PR.


src/components/v2/MintRepayVai/Mint/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+    if (value) {
+      await onSubmit(value);
+
+      // Reset value
+      setValue('');
+    }
+  };
+
+  return (
+    <form onSubmit={handleSubmit}>
```
coreyar: We should create a form component that standardizes our form behaviors and layout such as validation, loading, and form level errors

src/components/v2/MintRepayVai/Mint/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+            Available VAI Limit
+          </Typography>
+        </div>
+
+        <Typography component="span" css={styles.infoValue} variant="small1">
+          {readableLimitVai} VAI
+        </Typography>
+      </div>
+
+      <div css={styles.getRow({ isLast: true })}>
```
coreyar: nit: Prefixing style functions is a new convention. I could see us not having any preference about this or preferring to use or not use it.

therealemjy: Yeah, I've been doing that to differentiate functions from interpolated css strings. To me, it doesn't make a huge difference whether we all follow this convention or not (I think it's pretty clear to devs what this does either way), but if you want to follow another pattern I'm happy to oblige.
Just let me know what you prefer.

src/components/v2/MintRepayVai/Mint/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+      </SecondaryButton>
+    </form>
+  );
+};
+
+export const Mint: React.FC = () => {
+  // TODO: fetch actual data
+  const isUserLoggedIn = true;
+  const limitWei = new BigNumber('100.12').multipliedBy(new BigNumber(10).pow(18));
+  const mintFeePercentage = 2.14;
```
coreyar, line 132: Should this be a constant?

src/components/v2/MintRepayVai/Mint/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+      >
+        Mint VAI
+      </SecondaryButton>
+    </form>
+  );
+};
+
+export const Mint: React.FC = () => {
+  // TODO: fetch actual data
+  const isUserLoggedIn = true;
```
coreyar, line 130: In other parts of the app we use `useWeb3Account` to grab the account and check if the user is loggedin. We can remove this todo if we check this way.

src/components/v2/MintRepayVai/Mint/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+export interface IMintUiProps {
+  disabled: boolean;
+  isSubmitting: boolean;
+  onSubmit: (value: BigNumber) => Promise<void>;
+  limitWei: BigNumber;
+  mintFeePercentage: number;
+}
+
+// TODO: Move to dashboard component/container once created
+export const MintUi: React.FC<IMintUiProps> = ({
```
coreyar, line 47: This could be a nice naming convention to follow if the presentational component and container are in the same file and we want the container to have the final component name

src/components/v2/MintRepayVai/Mint/index.tsx
therealemjy: No, it will be fetched from the contract too. Check out line 108:
```// TODO: fetch actual data```

File: src/components/v2/MintRepayVai/Mint/index.tsx (hunk 4 above)
Comment (therealemjy) [reply src/components/v2/MintRepayVai/Mint/index.tsx:130]: Yeah that's how I plan on doing it here.
This is the container for `MintUi`, which is out of scope for this task. It will be worked on next when I get to this ticket: https://app.clickup.com/t/24qtzxz

File: src/components/v2/MintRepayVai/Mint/index.tsx (hunk 5 above)
Comment (therealemjy) [reply src/components/v2/MintRepayVai/Mint/index.tsx:47]: I hesitated on that to be honest. If you like it, I'm cool with following that pattern.

File: src/components/v2/MintRepayVai/Mint/index.tsx (hunk 1 above)
Comment (therealemjy) [reply src/components/v2/MintRepayVai/Mint/index.tsx]: Yeah agreed! I actually wanted to use react-hook-form for this, but they improve performances by using uncontrolled components only. Unfortunately in our case that won't work since we need to block the user input when they attempt to enter a value that's too low or too high.

You suggested using Formik when we had that discussion, I argued React Hook Form was better suited but since it turns out it isn't I go back on my position and suggest we use Formik. What do you think?

File: src/components/v2/MintRepayVai/Mint/styles.ts [hunk 6]
```diff
@@ -0,0 +1,35 @@
...
+      justify-content: space-between;
+      margin-bottom: ${theme.spacing(isLast ? 3 : 2)};
+    `,
+    infoColumn: css`
+      display: flex;
+      align-items: center;
+      padding-right: ${theme.spacing(2)};
+    `,
+    coinIcon: css`
+      width: 20px;
```
coreyar: We should add this as a value under shape. We'll have different icon sizes but there should be some standardization between large, medium, and small icons for example.

src/components/v2/MintRepayVai/styles.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const useStyles = () => {
+  const theme = useTheme();
+
+  return {
+    theme,
+    container: css`
+      width: 100%;
+      background-color: ${theme.palette.background.paper};
+      border-radius: 24px;
```
coreyar: this could be grabbed as a value under shape in the theme

src/components/v2/TextField/TextField.tsx
```diff
@@ -0,0 +1,58 @@
+/** @jsxImportSource @emotion/react */
```
coreyar: I think this should be removed because the updated component is at the index

src/components/v2/TextField/index.tsx
```diff
@@ -16,18 +16,41 @@ export interface ITextFieldProps extends Omit<InputHTMLAttributes<HTMLInputEleme
...
+  max,
+  min,
+  type,
   ...inputProps
 }) => {
   const styles = useStyles();
 
+  const handleChange: InputHTMLAttributes<HTMLInputElement>['onChange'] = e => {
+    // Prevent value from being updated if it does not follow the rules
+    const followsMaxRule =
```
coreyar, line 35: I don't think we should put validation in the TextComponent. I think we should put it in customized components for specific validation and text types. Ideally validation wouldn't happen in the Text component at all but would happen at the form level using a library like `yup`

src/components/v2/TextField/TextField.tsx
therealemjy: Yeah, that's good catch on your end because this file shouldn't exist at all ahah
https://github.com/VenusProtocol/venus-protocol-interface/pull/234

I might not have rebased with develop, I'll do that. Thanks!

src/components/v2/TextField/index.tsx
therealemjy: So in this case it's a bit different, because we're actually blocking the user's action when these rules aren't followed.
As far as I can see, `yup` wouldn't let us do that as it would validates the input after it's been entered already.
I agree validation should be outside the `TextField` component, however I was under the impression this doesn't quite fall in that category since it is blocking a user action. I'm happy to make any modifications we see fit.
What's your opinion on that?

coreyar: ok I understand, in that case if we have start to see lots of diffenent logic and styles for number and string inputs we might want to break them out but in that context this makes sense

src/components/v2/MintRepayVai/Mint/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  disabled,
+  limitWei,
+  mintFeePercentage,
+  isSubmitting,
+  onSubmit,
+}) => {
+  const styles = useStyles();
+  const [value, setValue] = React.useState<BigNumber | ''>('');
```
baomastr: I prefer `undefined` for empty value

src/components/v2/MintRepayVai/Mint/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+        <Typography component="span" css={styles.infoValue} variant="small1">
+          {readableFeeVai} VAI ({mintFeePercentage}%)
+        </Typography>
+      </div>
+
+      <SecondaryButton
+        css={styles.submitButton}
+        type="submit"
+        loading={isSubmitting}
+        disabled={disabled || !isValueValid}
```
baomastr: do we have any message that highlights the reason of disabling the form submit?

src/components/v2/MintRepayVai/Mint/index.tsx
baomastr: maybe we can have any hook or util to validate user login like `isLoggedIn()`?

src/components/v2/MintRepayVai/Mint/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const mintFeePercentage = 2.14;
+
+  const onSubmit: IMintUiProps['onSubmit'] = async value => {
+    // TODO: call contract
+    console.log('Amount to mint:', value.toString());
+  };
+
+  return (
+    <MintUi
+      disabled={!isUserLoggedIn}
```
baomastr, line 142: so we can do it directly inside of UI component with `isLoggedIn()`

src/components/v2/MintRepayVai/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+export interface IMintRepayVaiProps {
+  className?: string;
+}
+
+// TODO: Move to dashboard component/container once created
+export const MintRepayVai: React.FC<IMintRepayVaiProps> = ({ className }) => {
+  const styles = useStyles();
+  const [activeTabIndex, setActiveTabIndex] = React.useState(0);
+
+  const isOnMobile = useMediaQuery(styles.theme.breakpoints.down('sm'));
```
baomastr: it would be great to have utils for `isDown('sm')` etc

baomastr (approved):

src/components/v2/MintRepayVai/Mint/index.tsx
therealemjy: If we want yeah. Although to be honest we already have quite a few ways of checking that: `useWeb3Account`, `useAuth` or the `AuthContext` all return the account (if the user is connected with one).

src/components/v2/MintRepayVai/Mint/index.tsx
therealemjy: That would break the input, as it will be marked as uncontrolled by React when its value is `undefined`

src/components/v2/MintRepayVai/Mint/index.tsx
therealemjy: Not in the designs no. It's pretty obvious though, as it's only invalid while there's no value in the input or if it's 0

src/components/v2/MintRepayVai/Mint/index.tsx
therealemjy: It shouldn't be in the UI. `MintUi` needs to be dumb so no side-effect there.

src/components/v2/MintRepayVai/index.tsx
therealemjy: Why not use the theme directly? I'm cool with having utils, but here it feels it would only be an alias for `styles.theme.breakpoints.down`

src/components/v2/MintRepayVai/Mint/index.tsx
coreyar: I liked using Formik, but am happy with whatever works for us.

therealemjy: I think it will be the best option

therealemjy: Done

src/components/v2/MintRepayVai/index.tsx
baomastr: just in reason of comfortable short named function. 
code looks more clear with `isSmDown()` then `useMediaQuery(styles.theme.breakpoints.down('sm'))`

therealemjy: Got you. Let's do it then :) 
I'll update the PR.

therealemjy: Done :)

therealemjy:
@coreyar @baomastr the design team updated the designs slightly by adding the token icon inside the input, so I've added all the token icons to the `Icon` component to support this and since this will be needed in the future anyway.

coreyar (approved):

PR #245, opened 2022-03-30
Implement v2 layout in application pages

coreyar:


therealemjy:
Nice, that's a big piece of work.

It looks like the horizontal scrolling issue you talked about yesterday is there:
Screenshot 2022-04-01 at 11 15 18

The footer is also going under the sidebar:
Screenshot 2022-04-01 at 11 14 33

I also spotted a few issues with the sidebar.
The tablet responsive design is missing:
Screenshot 2022-04-01 at 11 16 36

Some of the spacings of the list items are off, and the active state is different from the one in the designs:
Screenshot 2022-04-01 at 11 17 24

Designs:
Screenshot 2022-04-01 at 11 20 31

coreyar:
@therealemjy I actually don't see the first two issues you see, so I'll have to do some investigating. The collapsed menu is there but it isn't responsive, I understood it to be interactive. So if you click the header it expands or collapses. I can double check with Pavel. 

I checked the spacings but I can double check them on the menu and good catch on the active color I'll push that update

coreyar:
Ahh I see looks like your comments are from the storybook layouts. It took a big of work to iron out the placement and I was mostly focused on the app but good catch that the storybooks aren't aligned

src/components/v2/Layout/Sidebar/styles.ts
```diff
@@ -47,31 +63,62 @@ export const useStyles = ({ expanded }: { expanded: boolean }) => {
...
+        justify-content: start;
+        ${theme.breakpoints.down('lg')} {
+          padding: ${theme.spacing(2)};
+          justify-content: center;
+        }
+        width: 100%;
+      }
+      ${theme.breakpoints.down('lg')} {
+        :hover {
+          border-radius: ${theme.shape.borderRadius.large}px;
```
therealemjy: The border radius used in the design is `16px` which we currently don't have in the theme.
Perhaps we could add a `medium` size that corresponds to a 16 pixel border-radius?

therealemjy (commented):
Looks good.
The tablet sidebar is perfect, as well as the mobile menu. Nice one 👍

I'm still seeing some small issues with the desktop view.

- The footer is still taking all of the horizontal space available instead of just the page width. Since we're not displaying any content on the left hand side of it, it's ok if it takes to the full width, but it needs to go below the sidebar
Screenshot 2022-04-04 at 19 05 49
Screenshot 2022-04-04 at 19 05 52

- The horizontal spacing for the header and the page container are set at 24 pixels, instead of 32 like in the designs
Screenshot 2022-04-04 at 19 06 14
Screenshot 2022-04-04 at 19 06 26
Screenshot 2022-04-04 at 19 06 21

- The VAI and XVS buttons don't have any border radius applied to them anymore, and the spacing between them is incorrect (32px instead of 12px). The spacing between these buttons and the connect button is also incorrect (32px instead of 24px).
If you feel this is out of scope, that's cool with me, but we should create a ticket to have a look at this before we release then
Screenshot 2022-04-04 at 19 06 47

coreyar:
> * The footer is also going under the sidebar:
Based on this comment I understood the request to be to put the footer above the sidebar

> * The footer is still taking all of the horizontal space available instead of just the page width. Since we're not displaying any content on the left hand side of it, it's ok if it takes to the full width, but it needs to go below the sidebar
I am happy to put it back below but I just want to make sure we are aligned

therealemjy (approved):

PR #252, opened 2022-04-01
Create BorrowMarket component

therealemjy:
- Create `BorrowMarket` component
Screenshot 2022-04-04 at 10 38 19

- create `convertCentsToDollars` util to convert a value expressed in cents into dollars (1001 => 10.01)
- create `formatDollarsToReadableValue` util to format a value expressed in cents into a readable format (1001 => $10.01)
- fix some styling issues with the `Table` component
- fix non-orderable columns of the `Head` component of the `Table` component
- create `Token` component
Screenshot 2022-04-04 at 10 48 35


src/components/v2/BorrowMarket/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+          orderDirection: 'asc',
+        }}
+      />
+    </div>
+  );
+};
+
+const BorrowMarket: React.FC = () => {
+  // TODO: fetch actual data
+  const assetDataFromContract = [
```
coreyar, line 79: I would make this accessible to be used in tests and story books by exporting it from the module scope, placing it in its own file or in the storybook.

src/components/v2/BorrowMarket/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+    {
+      key: 'liquidity',
+      render: () => formatDollarsToReadableValue(asset.liquidityCents),
+      value: asset.liquidityCents.toString(),
+    },
+  ]);
+
+  return (
+    <div className={className} css={styles.container}>
+      {/* TODO: add loading state */}
```
coreyar: Lets add loading and errors to the Table component

coreyar (approved):

src/components/v2/BorrowMarket/index.tsx
therealemjy: Good idea. I'll add a TODO for this.

src/components/v2/BorrowMarket/index.tsx
therealemjy: Yeah that's a good idea.

therealemjy: On second thought, I just realized this will be specific to the contract itself. I don't know what the data will actually look like yet, I'll have more information on that once I get to connect the container to the contract.
When I get to it, I'll extract the test data so it can be reused as you suggested

PR #255, opened 2022-04-02
Type token contract address constant

coreyar:
Add type to token constants so that they can be indexed with type safety.

### DIFF
--- a/src/constants/contracts.ts
+++ b/src/constants/contracts.ts
@@ -52,12 +52,11 @@ import vtrx from 'assets/img/coins/vtrx.png';
 import vust from 'assets/img/coins/vust.png';
 import vluna from 'assets/img/coins/vluna.png';
 
+import { TokenSymbol } from 'types';
 import { isOnTestnet } from 'config';
-import contracts from './contracts';
+import contracts from 'utilities/contracts';
 
-export const CONTRACT_XVS_TOKEN_ADDRESS = isOnTestnet
-  ? contracts.TEST.TOKEN.xvs
-  : contracts.MAIN.TOKEN.xvs;
+export const XVS_TOKEN = isOnTestnet ? contracts.TEST.TOKEN.xvs : contracts.MAIN.TOKEN.xvs;
 
 // Contract ABI
 export const CONTRACT_TOKEN_ADDRESS = isOnTestnet
@@ -446,7 +445,19 @@ export const CONTRACT_TOKEN_ADDRESS = isOnTestnet
       },
     };
 
-export const CONTRACT_VBEP_ADDRESS = isOnTestnet
+interface IToken {
+  id: TokenSymbol;
+  symbol: Uppercase<TokenSymbol>;
+  decimals: number;
+  address: string;
+  asset: string;
+  vasset: string;
+}
+
+export const getToken = (key: TokenSymbol): IToken =>
+  CONTRACT_TOKEN_ADDRESS[key as keyof typeof CONTRACT_TOKEN_ADDRESS] as IToken;
+
+export const VBEP_TOKENS = isOnTestnet
   ? {
       sxp: {
         id: 'sxp',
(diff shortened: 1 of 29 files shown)


src/utilities/constants.ts
```diff
@@ -60,7 +60,17 @@ export const CONTRACT_XVS_TOKEN_ADDRESS = isOnTestnet
...
+export const CONTRACT_TOKEN_ADDRESS: Record<
+  string,
+  {
+    id: string;
+    symbol: string;
+    address: string | undefined;
+    decimals: number;
+    asset: string;
+    vasset?: string;
+  }
```
therealemjy: Actually, this type is partially incorrect: the key of the record isn't a `string`, but a set of possible string values. This is important because we'll some accuracy by typing it as a `string` only. Perhaps we can extract all the token names into a `string` and use this value as key instead?

```typescript
type TokenSymbol = 'sxp' | 'xvs' | 'bnb'; // TODO: add all token symbols

type Token = {
  id: TokenSymbol,
  symbol: Uppercase<TokenSymbol>,
  decimals: number,
  address: string,
  asset: string,
  vasset?: string,
};

export const CONTRACT_TOKEN_ADDRESS: Record<TokenSymbol, Token> = {...};
```

src/constants/contracts.ts
```diff
@@ -446,6 +447,18 @@ export const CONTRACT_TOKEN_ADDRESS = isOnTestnet
...
+interface IContractTokenAddress {
+  id: TokenSymbol;
+  symbol: Uppercase<TokenSymbol>;
+  decimals: number;
+  address: string;
+  asset: string;
+  vasset: string;
+}
+
+export const getContractTokenAddress = (key: TokenSymbol): IContractTokenAddress =>
```
therealemjy: The name is a bit misleading to me. We're not just getting the address, we're getting the whole asset.
I think it should just be `getToken` or something like this.

src/constants/contracts.ts
```diff
@@ -446,6 +447,18 @@ export const CONTRACT_TOKEN_ADDRESS = isOnTestnet
       },
     };
 
+interface IContractTokenAddress {
```
therealemjy: I think that name is confusing. This isn't just the address, but the whole token. Perhaps we can simply name it `Token`.

src/constants/contracts.ts
```diff
@@ -671,3 +684,12 @@ export const CONTRACT_VBEP_ADDRESS = isOnTestnet
...
       },
     };
+
+interface IContractVbepAddress {
+  id: TokenSymbol;
+  symbol: `v${Uppercase<TokenSymbol>}`;
+  address: string;
+}
+
+export const getContractVbepAddress = (key: TokenSymbol): IContractVbepAddress =>
```
therealemjy: Same thing about the name

src/constants/contracts.ts
```diff
@@ -671,3 +684,12 @@ export const CONTRACT_VBEP_ADDRESS = isOnTestnet
         address: contracts.MAIN.VBEP.luna,
       },
     };
+
+interface IContractVbepAddress {
```
therealemjy: Same think here about the name, perhaps `VBepToken`?

src/constants/contracts.ts
```diff
@@ -446,6 +447,18 @@ export const CONTRACT_TOKEN_ADDRESS = isOnTestnet
...
+  decimals: number;
+  address: string;
+  asset: string;
+  vasset: string;
+}
+
+export const getContractTokenAddress = (key: TokenSymbol): IContractTokenAddress =>
+  CONTRACT_TOKEN_ADDRESS[key as keyof typeof CONTRACT_TOKEN_ADDRESS] as IContractTokenAddress;
+
 export const CONTRACT_VBEP_ADDRESS = isOnTestnet
```
therealemjy: Same thing here about the name

src/constants/contracts.ts
```diff
@@ -52,8 +52,9 @@ import vtrx from 'assets/img/coins/vtrx.png';
 import vust from 'assets/img/coins/vust.png';
 import vluna from 'assets/img/coins/vluna.png';
 
+import { TokenSymbol } from 'types';
 import { isOnTestnet } from 'config';
-import contracts from './contracts';
+import contracts from 'utilities/contracts';
 
 export const CONTRACT_XVS_TOKEN_ADDRESS = isOnTestnet
```
therealemjy: I think that name is confusing. It isn't just the address but the actual token. Perhaps `XVS_TOKEN`.

therealemjy (commented):
Looks good, just a few comments on naming.

baomastr (approved):

therealemjy (approved):

PR #259, opened 2022-04-04
Create AmountForm

therealemjy:
- Create `AmountForm` component for single `BigNumber` value forms, with validation built-in
- Decouple `Mint` subcomponent of `MintRepayVai` and use `AmountForm`


src/components/v2/AmountForm/validateForm.ts
```diff
@@ -0,0 +1,15 @@
+import { FormValues } from './types';
+
+const validateForm = (values: FormValues) => {
+  const errors: {
+    amount?: string;
+  } = {};
+
+  if (!values.amount || values.amount.isEqualTo(0)) {
+    errors.amount = 'required';
```
coreyar: I'd like to use `yup` for validation. What are your thoughts?

therealemjy: I wanted to originally, but in this case I feel it's adding more complexity than anything, because we'd have to create a custom `yup` type to handle `BigNumber` values, which means also having to extend yup's Typescript definition in order to support our custom type. This feels like a lot of overhead for a use case that's only relevant to this specific form.

As far as I can see, the only other type of form we have in the app is the form for proposing a vote. For this one I think using `yup` would make sense since the values used will be supported out of the box.

src/components/v2/MintRepayVai/Mint/index.tsx
```diff
@@ -58,17 +34,10 @@ export const MintUi: React.FC<IMintUiProps> = ({
...
-  };
-
   return (
-    // TODO: add validation schema
-    <Formik initialValues={initialValues} onSubmit={handleSubmit}>
-      {({ values, setFieldValue, handleBlur, isSubmitting, isValid }) => (
-        <Form>
+    <AmountForm onSubmit={onSubmit}>
+      {({ values, setFieldValue, handleBlur, isSubmitting, isValid, touched }) => (
+        <>
```
baomastr, line 40: is this fragment needed?

baomastr (approved):

src/components/v2/MintRepayVai/Mint/index.tsx
therealemjy: Yeah, the render function only allows one JSX element to be returned.

src/components/v2/AmountForm/validateForm.ts
coreyar: I think the complexity difference is marginal but there are advantages to using yup which is why I think it would be nice to use yup here although it may be only used in one place for now. Yup gives us a really clean way to define validation in a single place included consistent error messages. It is likely that we will have to extend yup in the future so setting up how to extend it is worth the time. Once we have an example on how to extend yup it should be very easy for anyone to extend it from that point on

coreyar (approved):

PR #262, opened 2022-04-04
Move mint/ repay vai to Dashboard page dir

coreyar:
Organizes component specific to Dashboard within the Dashboard


src/pages/Dashboard/MintRepayVai/index.tsx
```diff
@@ -2,8 +2,8 @@
 import React from 'react';
 
 import { useIsSmDown } from 'hooks/responsive';
+import { Tabs } from 'components';
 import { Mint } from './Mint';
-import { Tabs } from '../Tabs';
```
therealemjy, line 6: I'd keep the relative import, or we'll end up with dependency cycles down the line

coreyar: We should not have dependency cycles out side of the components dir and our components dir and project is structure for imports to use aliases and work this way. If we are brining the MintRepayVai component into the `components` dir we would have issues but at that point we should move into the components dir and use relative imports.

From the pages dir absolute import of components is convenient and safe

therealemjy (approved):

PR #263, opened 2022-04-05
Create SuccessfulTransactionMessage component

therealemjy:
- add `check` icon to `Icon` component
- update list of component imports at `src/components/index.ts`
- extract `BscLink` components from `AuthModal`
- create `SuccessfulTransactionMessage` component
- create `generateBscScanAddressUrl` util to generate a URL on BSC scan from an address

Screenshot 2022-04-05 at 09 49 36


src/components/v2/BscLink/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export interface IBscLinkProps {
+  hash: string;
+  className?: string;
+}
+
+export const BscLink: React.FC<IBscLinkProps> = ({ hash, className }) => {
+  const styles = useStyles();
+
+  const bscScanUrl = `https://${isOnTestnet ? 'testnet.' : ''}bscscan.com/address/${hash}`;
```
baomastr: can be a good util that wraps urls. wouldn't it used in other places?

baomastr (approved):

src/components/v2/BscLink/index.tsx
therealemjy: Yeah good point. It could be reused in the footer already. I'll make the change 👍

therealemjy: Done

src/components/v2/Layout/Footer/index.tsx
```diff
@@ -37,7 +38,12 @@ export const Footer: React.FC<IFooterProps> = ({ currentBlockNumber }) => {
       </Typography>
 
       <div css={styles.links}>
-        <a css={styles.link} href={ETHERSCAN_XVS_URL} target="_blank" rel="noreferrer">
+        <a
```
coreyar, line 41: Should we use typography here?

coreyar (approved):

src/components/v2/Layout/Footer/index.tsx
therealemjy: We shouldn't no, the anchor tag is used as a wrapper only here, we don't display any text inside it.

PR #264, opened 2022-04-05
Wire up supply table

coreyar:
This PR adds side effects to fetch assets and enable them as collateral. It also update the table component to accept on option onClick prop for rows and adds a hover effect.


src/pages/Dashboard/Markets/SupplyMarket.tsx
```diff
@@ -76,23 +95,63 @@ export const SupplyMarketUi: React.FC<ISupplyMarketUiProps> = ({
...
+          `There was a problem disabeling collateral ${asset.name}. Please try again.`,
+        );
+      }
+    } else {
+      throw new ToastError(
+        'Collateral Required',
+        'Please repay all borrowed assets or set other assets as collateral.',
+      );
+    }
+  };
```
therealemjy: We should use React Query for this as we discussed

src/utilities/errors.ts
```diff
@@ -0,0 +1,11 @@
...
+  title: string;
+
+  description: string;
+
+  constructor(title: string, description: string) {
+    super(title);
+    this.title = title;
+    this.description = description;
+  }
+}
```
therealemjy, line 11: 👍

src/components/v2/Table/index.tsx
```diff
@@ -80,7 +88,7 @@ export function Table({ columns, data, title, minWidth, initialOrder }: ITablePr
 
         <TableBody>
           {rows.map(row => (
-            <TableRow key={uid(row)}>
+            <TableRow hover key={uid(row)} onClick={() => rowOnClick && rowOnClick(row)}>
```
therealemjy: I just saw we use `uid(row)` as key, but it will yield a different value every time a prop of `row` changes. We had that issue with the `Vaults` page.

src/clients/api/queries/getAssetsInAccount.ts
```diff
@@ -0,0 +1,18 @@
...
+export interface IGetAssetsInAccountInput {
+  comptrollerContract: Contract; // @TODO: use contract type (through Typechain?)
+  account: string | null | undefined;
+}
+
+export type IGetAssetsInAccountOutput = number;
+
+const getAssetsInAccount = async ({ comptrollerContract, account }: IGetAssetsInAccountInput) => {
+  let response = [];
+  if (account) {
```
therealemjy: I think `account` should be mandatory here.
The validation should happen beforehand, so that we don't have false positive since the empty array doesn't necessarily indicate the user isn't connected.

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export interface IGetHypotheticalAccountLiquidityInput {
+  comptrollerContract: Contract; // @TODO: use contract type (through Typechain?)
+  account: string | null | undefined;
+  asset: Asset;
+  balanceOf: string;
+  borrowAmount?: number | BigNumber;
+}
+
+export type IGetHypotheticalAccountLiquidityOutput = number;
```
therealemjy: `GetHypotheticalAccountLiquidityInput` and `GetHypotheticalAccountLiquidity` (to make them as types, not interfaces).

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
```diff
@@ -0,0 +1,34 @@
...
+  balanceOf,
+  borrowAmount = 0,
+}: IGetHypotheticalAccountLiquidityInput) => {
+  const a = asset;
+  if (account) {
+    const response = await comptrollerContract.methods
+      .getHypotheticalAccountLiquidity(account, asset.vtokenAddress, balanceOf, borrowAmount)
+      .call();
+    if (response) {
+      a.hypotheticalLiquidity = response;
```
therealemjy: I avoid mutating parameters, this can lead to a lot of issues that are hard to debug. An easy solution is to make a shallow copy of `asset`.
```typescript
const a = {...asset};
```

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const getHypotheticalAccountLiquidity = async ({
+  comptrollerContract,
+  account,
+  asset,
+  balanceOf,
+  borrowAmount = 0,
+}: IGetHypotheticalAccountLiquidityInput) => {
+  const a = asset;
+  if (account) {
```
therealemjy: Same thing here, I'd make `account` mandatory.

src/clients/api/queries/getMarkets.ts
```diff
@@ -0,0 +1,30 @@
...
+import { VBEP_TOKENS } from 'constants/contracts';
+import { Market } from 'types';
+
+// interface IGetMarketsResponse {
+//   dailyVenus: string;
+//   markets: Market[];
+//   request: { addresses: string[] };
+//   venusRate: string;
+// }
+
```
therealemjy: We can remove that.

src/clients/api/queries/getVTokenBalancesAll.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export interface IGetVTokenBalancesAllOutput {
+  balanceOf: string;
+  balanceOfUnderlying: string;
+  borrowBalanceCurrent: string;
+  tokenAllowance: string;
+  tokenBalance: string;
+  vToken: string;
+}
```
therealemjy: Shall we use types instead of interfaces here too? This way we're consistent with the other queries.

src/clients/api/queries/getVTokenBalancesAll.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const getVTokenBalancesAll = async ({
+  venusLensContract,
+  vtAddresses,
+  account,
+}: IGetVTokenBalancesAllInput) => {
+  let response = [];
+  if (account) {
```
therealemjy: Same thing with the `account` (I'd make it mandatory).

src/hooks/useUserMarketInfo.ts
```diff
@@ -0,0 +1,181 @@
...
+  const { data: vTokenBalancesAccount } = useQuery(
+    createGetVTokenBalancesAllKey({ vtAddresses, account }),
+    () => getVTokenBalancesAll({ venusLensContract, vtAddresses, account }),
+    { placeholderData: [] },
+  );
+  const { data: vTokenBalancesTreasury } = useQuery(
+    createGetVTokenBalancesAllKey({ vtAddresses, account: TREASURY_ADDRESS }),
+    () => getVTokenBalancesAll({ venusLensContract, vtAddresses, account: TREASURY_ADDRESS }),
+    { placeholderData: [] },
+  );
```
therealemjy: Let's move each of them in their own file, this way they are reusable ;) 

Also this way each of them can import the contracts they need (they're memoized anyway so this won't add any significant processing) and so `useUserMarketInfo` doesn't have to handle that.

src/constants/functionKey.ts
```diff
@@ -0,0 +1,27 @@
...
+  // Mutations
+}
+
+export const createGetHypotheticalLiquidityKey = ({
+  account,
+  token,
+}: {
+  account: string | null | undefined;
+  token: string;
+}) => `${FunctionKey.GET_HYPOTHETICAL_LIQUIDITY}-${account}-${token}`;
```
therealemjy: Just a heads up that arrays are supported by React Query. That can make things a bit simpler later if we want to retrieve the results from the cache:
```
[FunctionKey.GET_HYPOTHETICAL_LIQUIDITY, token]
```

I also didn't include the account, because the cache is relevant for the connected user only anyway and this query would get wiped out on log out.

src/constants/functionKey.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const createGetVTokenBalancesAllKey = ({
+  account,
+  vtAddresses,
+}: {
+  account: string | null | undefined;
+  vtAddresses: string[];
+}) => `${FunctionKey.GET_VTOKEN_BALANCES_ACCOUNT}-${account}-${JSON.stringify(vtAddresses)}`;
```
therealemjy: Same thing here with the array (and not including `account`) :)

src/hooks/useUserMarketInfo.ts
```diff
@@ -0,0 +1,181 @@
...
+      queryKey: createGetHypotheticalLiquidityKey({ account, token: asset.name }),
+      queryFn: () =>
+        getHypotheticalAccountLiquidity({
+          comptrollerContract,
+          account,
+          asset,
+          balanceOf: balances[asset.vtokenAddress.toLowerCase()]?.balanceOf,
+        }),
+    })),
+  );
```
therealemjy: Same here, let's move this into its own file

src/pages/Dashboard/Markets/CollateralConfirmModal.tsx
```diff
@@ -0,0 +1,35 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+import { Icon, Modal, IModalProps } from 'components';
+import { Asset } from 'types';
+import { ReactComponent as LogoWithText } from 'assets/img/v2/venusLogoWithText.svg';
+import Typography from '@mui/material/Typography';
+import { useStyles } from './styles';
+
+export interface IConfirmCollateralModalProps {
+  asset: Asset | undefined;
```
therealemjy, line 10: Very small thing, feel free to ignore it:
```suggestion
  asset?: Asset;
```

src/pages/Dashboard/Markets/SupplyMarket.tsx
```diff
@@ -76,23 +95,63 @@ export const SupplyMarketUi: React.FC<ISupplyMarketUiProps> = ({
...
+    } else if (!asset || !asset.borrowBalance.isZero()) {
+      throw new ToastError(
+        'Collateral Required',
+        'Please repay all borrowed assets or set other assets as collateral.',
+      );
+    } else if (!asset.collateral) {
+      try {
+        await comptrollerContract.methods
+          .enterMarkets([asset.vtokenAddress])
+          .send({ from: account });
```
therealemjy: We should use React Query for this too, with a `useMutation` hook.

src/pages/Dashboard/Markets/SupplyMarket.tsx
```diff
@@ -76,23 +95,63 @@ export const SupplyMarketUi: React.FC<ISupplyMarketUiProps> = ({
...
+      } catch (error) {
+        console.log('enter markets error :>> ', error);
+        throw new ToastError(
+          'Collateral Enable Error',
+          `There was a problem enabeling collateral ${asset.name}. Please try again.`,
+        );
+      }
+    } else if (+asset.hypotheticalLiquidity['1'] > 0 || +asset.hypotheticalLiquidity['2'] === 0) {
+      try {
+        await comptrollerContract.methods.exitMarket(asset.vtokenAddress).send({ from: account });
```
therealemjy: Same thing here: `useMutation`

src/components/v2/Table/index.tsx
```diff
@@ -80,7 +88,7 @@ export function Table({ columns, data, title, minWidth, initialOrder }: ITablePr
 
         <TableBody>
           {rows.map(row => (
-            <TableRow key={uid(row)}>
+            <TableRow hover key={uid(row.length)} onClick={() => rowOnClick && rowOnClick(row)}>
```
therealemjy: Unfortunately that doesn't work either, because the rows all have the same length, resulting in them having the same key:
Screenshot 2022-04-08 at 08 48 09

A way to fix that can be to change the data model so that each row is an object that contains a `key` and the cells.
If you want to keep using an array for cells, you can simply change the type of table rows to something like that:
```typescript
type Row = {
  key: string;
  cells: ITableRowProps[];
}
```

Side note: `ITableRowProps` should probably have another name, since it's the type of a cell rather than a row.

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
coreyar: Interface is the correct option here. We are describing the shape of data/ object. Types are good for aliasing types of data (unions, primatives, other types ect.)

src/clients/api/queries/getMarkets.ts
coreyar: I think this would be nice to keep but I can remove it. It describes the shape of the data being returned but I didn't want to clutter eslint  with an unused variable name. I don't feel strongly about keeping it but I do think it provides useful information about what is going on here

src/clients/api/queries/getAssetsInAccount.ts
coreyar: I agree that this would be nice but the requirements for hooks require that the number of hooks remains the same on every render which means `useUserMarketInfo` with the hooks we want must be called when account is null or undefined because we can't conditionally call them if account is defined or not.

coreyar: We can't default to an empty string here because the calls will throw unnecessary invalid address errors

therealemjy: You can conditionally call a query, using the `enabled` prop: https://react-query.tanstack.com/guides/dependent-queries

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
therealemjy: Right, but the second type isn't an interface. You posted this exact comment on my PR actually ahah

coreyar: 🤦 I read that wrong

src/clients/api/queries/getVTokenBalancesAll.ts
coreyar: Ahh this is where I mean to post my other comment
> Interface is the correct option here. We are describing the shape of data/ object. Types are good for aliasing types of data (unions, primatives, other types ect.)

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
coreyar: Actually I think it was meant as a response here - https://github.com/VenusProtocol/venus-protocol-interface/pull/264#discussion_r845817589

src/clients/api/mutations/exitMarket.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type ExitMarketOutput = void;
+
+const exitMarket = async ({
+  comptrollerContract,
+  account,
+  vtokenAddress,
+}: IExitMarketInput): Promise<ExitMarketOutput> => {
+  await comptrollerContract.methods.exitMarket(vtokenAddress).send({ from: account });
+};
```
therealemjy: Small thing:
```suggestion
}: IExitMarketInput): Promise<ExitMarketOutput> => comptrollerContract.methods.exitMarket(vtokenAddress).send({ from: account });
```

src/clients/api/mutations/enterMarkets.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type EnterMarketsOutput = void;
+
+const enterMarkets = async ({
+  comptrollerContract,
+  account,
+  vtokenAddresses,
+}: IEnterMarketsInput): Promise<EnterMarketsOutput> => {
+  await comptrollerContract.methods.enterMarkets(vtokenAddresses).send({ from: account });
+};
```
therealemjy: Small thing:
```suggestion
}: IEnterMarketsInput): Promise<EnterMarketsOutput> => comptrollerContract.methods.enterMarkets(vtokenAddresses).send({ from: account });
```

src/clients/api/queries/getAssetsInAccount.ts
```diff
@@ -0,0 +1,15 @@
...
+  comptrollerContract: Contract; // @TODO: use contract type (through Typechain?)
+  account: string | undefined | null;
+}
+
+export type IGetAssetsInAccountOutput = number;
+
+const getAssetsInAccount = async ({ comptrollerContract, account }: IGetAssetsInAccountInput) => {
+  const response = await comptrollerContract.methods.getAssetsIn(account?.toLowerCase()).call();
+  return response;
+};
```
therealemjy:
```suggestion
const getAssetsInAccount = async ({ comptrollerContract, account }: IGetAssetsInAccountInput): Promise<IGetAssetsInAccountOutput> => comptrollerContract.methods.getAssetsIn(account?.toLowerCase()).call();
```

src/clients/api/mutations/exitMarket.ts
```diff
@@ -0,0 +1,19 @@
+import { Contract } from 'web3-eth-contract';
+
+export interface IExitMarketInput {
+  comptrollerContract: Contract; // @TODO: use contract type (through Typechain?)
+  account: string | undefined | null;
+  vtokenAddress: string;
+}
+
+export type ExitMarketOutput = void;
```
therealemjy, line 9: Right, so that's what I meant earlier. Should stick to using types instead of interfaces for these?
The only reason I ask this is so we're consistent with our naming. When we import these types from elsewhere, it isn't intuitive in my opinion to have to guess whether they are prefixed with and "I" or not.

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export type GetHypotheticalAccountLiquidityOutput = number;
+
+const getHypotheticalAccountLiquidity = async ({
+  comptrollerContract,
+  account,
+  asset,
+  balanceOf,
+  borrowAmount = 0,
+}: IGetHypotheticalAccountLiquidityInput) => {
```
therealemjy:
```suggestion
}: IGetHypotheticalAccountLiquidityInput): Promise< GetHypotheticalAccountLiquidityOutput > => {
```

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
```diff
@@ -0,0 +1,37 @@
...
+      account?.toLowerCase(),
+      asset.vtokenAddress,
+      balanceOf,
+      borrowAmount,
+    )
+    .call();
+  if (response) {
+    a.hypotheticalLiquidity = response;
+  }
+  return a;
```
therealemjy: It seems you're returning the wrong thing here, according to the return type `GetHypotheticalAccountLiquidityOutput`.
I do think we should only return a number representing the hypothetical account liquidity for a given asset, rather than the whole asset (since we already have that info).

src/clients/api/queries/getMarkets.spec.ts
```diff
@@ -0,0 +1,89 @@
+import { restService } from 'utilities/restService';
+import getMarkets from './getMarkets';
+
+jest.mock('utilities/restService');
```
therealemjy, line 4: Nice. I think we'll want to always mock this service when testing, so we could put this line inside a `jest.config.ts` file at the root of the project.

src/clients/api/queries/getMarkets.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  underlyingDecimal: 6,
+  underlyingName: 'USDC',
+  underlyingPrice: '999900000000000000000000000000',
+  underlyingSymbol: 'USDC',
+  venusBorrowIndex: '2513615398720335958062590645849037196062852386675',
+  venusSpeeds: '17361111000000000',
+  venusSupplyIndex: '18116117080571365091301617497156799904494329',
+};
+
+describe('api/queries/getMarketss', () => {
```
therealemjy:
```suggestion
describe('api/queries/getMarkets', () => {
```

src/clients/api/queries/getMarkets.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    (restService as jest.Mock).mockImplementationOnce(async () => ({
+      status: 200,
+      data: { data: { markets: [supportedMarket] } },
+    }));
+
+    const response = await getMarkets();
+
+    expect(response).toHaveLength(1);
+  });
+  test('filters unsupported markets', async () => {
```
therealemjy:
```suggestion
  });
  
  test('filters unsupported markets', async () => {
```

src/clients/api/queries/getMarkets.ts
therealemjy: In that case, let's use it. If we want to go the extra mile, we can update the `restService` function so it can take a generic which types the returned data. This way you could pass this interface to it and have your response typed:
```
const response = restService<IGetMarketsResponse>({ api: '/governance/venus', method: 'GET' });
```

src/clients/api/queries/getVTokenBalancesAll.ts
```diff
@@ -0,0 +1,46 @@
...
+  tokenAllowance: string;
+  tokenBalance: string;
+  vToken: string;
+}
+
+const getVTokenBalancesAll = async ({
+  venusLensContract,
+  vtAddresses,
+  account,
+}: IGetVTokenBalancesAllInput) => {
```
therealemjy, line 31:
```suggestion
}: IGetVTokenBalancesAllInput): Promise<IGetVTokenBalancesAllOutput> => {
```

src/hooks/useUserMarketInfo.ts
```diff
@@ -0,0 +1,183 @@
...
+  const { data: vTokenBalancesAccount } = useQuery(
+    [FunctionKey.GET_VTOKEN_BALANCES_ACCOUNT, vtAddresses, account],
+    () => getVTokenBalancesAll({ venusLensContract, vtAddresses, account }),
+    { placeholderData: [], enabled: !!account },
+  );
+  const { data: vTokenBalancesTreasury } = useQuery(
+    [FunctionKey.GET_VTOKEN_BALANCES_ACCOUNT, TREASURY_ADDRESS, account],
+    () => getVTokenBalancesAll({ venusLensContract, vtAddresses, account: TREASURY_ADDRESS }),
+    { placeholderData: [], enabled: !!account },
+  );
```
therealemjy: Let's move each of these into separate files too, so that they are reusable.

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
```diff
@@ -0,0 +1,37 @@
+import BigNumber from 'bignumber.js';
+import { Contract } from 'web3-eth-contract';
+import { Asset } from 'types';
+
+export interface IGetHypotheticalAccountLiquidityInput {
+  comptrollerContract: Contract; // @TODO: use contract type (through Typechain?)
+  account: string | undefined | null;
```
therealemjy, line 6: `account` should be mandatory here.

src/hooks/useUserMarketInfo.ts
```diff
@@ -0,0 +1,183 @@
...
+      supplyBalance,
+      borrowBalance,
+      isEnabled,
+      collateral,
+      percentOfLimit,
+      hypotheticalLiquidity: ['0', '0', '0'] as [string, string, string],
+    };
+    return [...acc, asset];
+  }, []);
+  console.log(assetList);
```
therealemjy: Let's remove this `console.log`

src/pages/Dashboard/Markets/SupplyMarket.tsx
```diff
@@ -76,23 +98,85 @@ export const SupplyMarketUi: React.FC<ISupplyMarketUiProps> = ({
...
+      throw new ToastError(
+        'Collateral Required',
+        'Please repay all borrowed assets or set other assets as collateral.',
+      );
+    } else if (!asset.collateral) {
+      try {
+        setConfirmCollateral(asset);
+        enterMarkets({ comptrollerContract, vtokenAddresses: [asset.vtokenAddress], account });
+      } catch (error) {
+        console.log('enter markets error :>> ', error);
```
therealemjy, line 147: Let's remove this `console.log`

src/pages/Dashboard/Markets/SupplyMarket.tsx
```diff
@@ -76,23 +98,85 @@ export const SupplyMarketUi: React.FC<ISupplyMarketUiProps> = ({
...
+      }
+    } else if (+asset.hypotheticalLiquidity['1'] > 0 || +asset.hypotheticalLiquidity['2'] === 0) {
+      try {
+        setConfirmCollateral(asset);
+        exitMarkets({ comptrollerContract, vtokenAddress: asset.vtokenAddress, account });
+      } catch (error) {
+        throw new ToastError(
+          'Collateral Disable Error',
+          `There was a problem disabeling collateral ${asset.name}. Please try again.`,
+        );
```
therealemjy, line 161: This case can't happen, `exitMarkets` doesn't throw. You'll have to handle errors through the `onError` callback of `useExitMarket`

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
coreyar: It is mandatory. but it can be undefined or null

src/hooks/useUserMarketInfo.ts
coreyar: I've done this. I didn't get a chance to finish updating

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
coreyar: We could do that but there are so many loops in the hook that I decided to follow this original pattern to avoid another loop. I was able to refactor some out but not all

src/clients/api/mutations/exitMarket.ts
coreyar: No we should not. I am not clear on the benefit of putting the I in front of interfaces. It is new to me and I can look into it. This was something that Dmitriy and you seemed to be interested in. There is even an eslint plugin to enforce it. 

These types are defined correctly. We can change the naming convention to drop the I.

coreyar: This is what eslint says `It can be hard to differentiate between classes and interfaces. Prefixing interfaces with "I" can help telling them apart at a glance.``

This is motivated by the ability to use classes as types.

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
coreyar: This function is not returning the promise so it is typed correctly.

src/clients/api/queries/getMarkets.spec.ts
coreyar: You mean like a setup file. I don't think this would go in the config

src/clients/api/queries/getVTokenBalancesAll.ts
coreyar: This is not returning a promise. so this type is not correct.

coreyar: What happened here was odd. The contract returns an array with named properties. Rather than index the values we use the named properties somehow in the `indexBy` function these named properties on the array were missing. which would throw an error. This was a quick solution.

src/pages/Dashboard/Markets/SupplyMarket.tsx
coreyar: ExitMarkets does throw. If you send the wrong arguments it will throw an error

src/clients/api/queries/getVTokenBalancesAll.ts
coreyar: I was able to remove this

src/clients/api/queries/getMarkets.ts
coreyar: I think this is a nice idea. I don't think it is worth digging more into this rest service at the moment because we will be replacing it. I think the comment works for now. I have no problem removing it because it is easy to find and recreate this information but I don't see the downside to keeping it. We have informative comments in other places as well.

I'm almost done with this update

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
coreyar: I was able to refactor this.

src/clients/api/queries/getHypotheticalAccountLiquidity.ts
coreyar: Updated after refactor

therealemjy (approved):

PR #266, opened 2022-04-06
Value update component

coreyar:
This component takes two numbers and displays them horizontall with a colored arrow
designation the direction change in value ( up or down)

### Props
|name|type|required|description|
|------|--------|---|-------------|
|classname| string| false|optional className for applying css object styles|
|original| number or BigNumber|true| the compared left value|
|new|number or BigNumber|true|the compared right value|
|format|function|false| optional function to format numbers defaults to formatCentsToReadableValue|


src/components/v2/Icon/icons/arrowShaft.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none"
```
therealemjy:
```suggestion
<svg width="16" height="16" viewBox="0 0 16 16" fill="currentColor"
```

This way we can use the `color` attribute directly. `fill` works too, but if you check the other icons so far we've been using `color` to colorize them.

src/components/v2/ValueUpdate/styles.ts
```diff
@@ -0,0 +1,19 @@
...
+export const useStyles = ({ increase }: { increase: boolean }) => {
+  const theme = useTheme();
+
+  return {
+    container: css`
+      display: flex;
+      align-items: center;
+    `,
+    icon: css`
+      fill: ${increase ? theme.palette.interactive.success : theme.palette.interactive.error};
```
therealemjy: Following my comment on using `color` instead of `fill` to be consistent:
```suggestion
      color: ${increase ? theme.palette.interactive.success : theme.palette.interactive.error};
```

src/components/v2/Icon/icons/arrowShaft.svg
coreyar: This is somethin we should probably do a find and replace for because many icons still use `fill="none"`

src/components/v2/ValueUpdate/styles.ts
coreyar: And are a few are styled with fill if this is the standard we want to hold

baomastr (approved):

therealemjy (approved):

PR #267, opened 2022-04-06
Wire up MintVai container/component

therealemjy:
- set `REACT_APP_CHAIN_ID` environment variable to 97 (testnet ID) in Storybook, so that stories are always run in the test environment
- update how requests to RPC provider are mocked in Storybook so that all RPC urls are mocked for (only one was being mocked previously). **This breaks all the current pages in Storybook, since previously the calls weren't actually mocked. For that reason, I've removed all of them except the dashboard. We should add the other pages back as we refactor them (and mock the relevant requests).**
- create `mintVai` api function and `useMintVai` hook, to handle minting VAI through a contract call
- create `getVaiTreasuryPercentage ` api function and `useGetVaiTreasuryPercentage` hook, to handle fetching the VAI treasury percentage (our fee basically) through a contract call
- create `constants` folder to contain all the global constants (token info and function keys for now)
- remove `withThemeProvider` from the list of decorators added to each stories, since this is included by default through the `preview.js` file
- update validation schema of `AmountForm` to forbid `undefined` values and give a more meaningful error message (although it isn't currently displayed, at least it gives a better context to devs)
- wire up `MintVai` to the blockchain (and our backend)


coreyar:
I don't think we should remove existing storybooks. We should accept the new baselines and take them going forward.
In addition to visual regressions, the snapshots also test that our components render and there are no errors rendering components with basic props

therealemjy:
Just to clarify: only the stories for the pages are removed, all the ones for components are staying in (so that as you say we can make sure components are rendering properly).
I've removed the page stories because they'll be broken until we eventually replace them one by one as we refactor the pages. I can keep them in if you see value in that.

src/clients/api/mutations/mintVai.spec.ts
```diff
@@ -0,0 +1,57 @@
+import BigNumber from 'bignumber.js';
+
+import mintVai from './mintVai';
+
+describe('api/mutation/mintVai', () => {
+  test('throws an error when request fails', async () => {
```
coreyar, line 6: I see as similar test case in other mutations but I don't see anything in the mutations where we are writing code to trigger this. My understanding is that this is handled by ReactQuery so this is a test case covering more ReactQuery than our code. Am I missing something?

src/constants/contractTokenAddresses.ts
```diff
@@ -53,7 +53,7 @@ import vust from 'assets/img/coins/vust.png';
 import vluna from 'assets/img/coins/vluna.png';
 
 import { isOnTestnet } from 'config';
-import contracts from './contracts';
+import contracts from '../utilities/contracts';
```
coreyar: I updated this file in the typing PR to be `contracts`. Which name would you like to use? You requested we drop addresses from the names of all the constants. I think it would make sense to also drop it from the name of the file.

src/clients/api/queries/getVaiTreasuryPercentage.ts
```diff
@@ -0,0 +1,21 @@
+import BigNumber from 'bignumber.js';
+
+export interface IGetVaiTreasuryPercentageInput {
+  vaiControllerContract: $TSFixMe; // @TODO: use contract type (through Typechain?)
+}
+
+export type IGetVaiTreasuryPercentageOutput = number;
```
coreyar: This is not an interface so it should not be prefaced with an `I`

src/clients/api/mutations/mintVai.spec.ts
therealemjy: In this case React Query isn't involved, we're just checking that the error isn't silenced.
For example, we could have something like this:
```typescript
const mintVai = async () => {
  try {
    await contractCall();
  } catch(error) {
    console.log(error)
  }
};
```

In that case, the error gets silenced and so React Query wouldn't pick it up but instead would mark it as having been successful. That's why I added this test, to make sure we don't silence errors.

src/constants/contractTokenAddresses.ts
therealemjy: Ah yeah sorry for that, I think I made these changes before our discussion and forgot to make amendments. Let me update to what you've done in your PR instead

src/clients/api/queries/getVaiTreasuryPercentage.ts
therealemjy: You're right. That applies to the types of other queries and mutations as well. I'll make the change 👍

src/containers/Main/VrtConversion.tsx
```diff
@@ -5,7 +5,7 @@ import { useWeb3Account } from 'clients/web3';
 import { Row, Col } from 'antd';
 import LoadingSpinner from '../../components/Basic/LoadingSpinner';
 import useRefresh from '../../hooks/useRefresh';
-import * as constants from '../../utilities/constants';
+import * as constants from '../../constants/contractTokenAddresses';
```
coreyar: Per the deleted file below.  I would not delete or remove any stories because it reduces our test coverage. If you really want to stop snapshotting a screen you can skip it but I'd highly discourage skipping and removing test cases

therealemjy: Ok, I'll add them all back in

src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+  // Convert limit into VAI
+  const readableVaiLimit = React.useMemo(
+    () =>
+      !limitWei
+        ? '-'
+        : convertWeiToCoins({
+            value: limitWei,
+            tokenSymbol: VAI_SYMBOL,
+            returnInReadableFormat: true,
+          }).toString(),
```
coreyar: nit: As a rule of thumb I think ternaries that check if a value is truthy and then return a truthy value on the left and a falsey value on the right are the easiest to read

```suggestion
      limitWei
        ? convertWeiToCoins({
            value: limitWei,
            tokenSymbol: VAI_SYMBOL,
            returnInReadableFormat: true
            : '-',
          }).toString(),
```

src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  const hasMintableVai = limitWei?.isGreaterThan(0) || false;
+
+  const getReadableMintFee = React.useCallback(
+    (valueWei: BigNumber | '') => {
+      if (!mintFeePercentage) {
+        return '-';
+      }
+
+      const readableFeeVai = !valueWei ? '0' : getReadableFeeVai({ valueWei, mintFeePercentage });
```
coreyar:
```suggestion
      const readableFeeVai = valueWei ? getReadableFeeVai({ valueWei, mintFeePercentage }) : '0';
```

src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+            Mint VAI
+          </SecondaryButton>
+        </>
+      )}
+    </AmountForm>
+  );
+};
+
+const MintVai: React.FC = () => {
+  const { account } = React.useContext(AuthContext);
```
coreyar: I think we should consistently access the account across the application. This method doesn't currently exist anywhere. We use `const { account } = useWeb3Account();`

therealemjy: The reason I started using the context is because it also contains the methods for opening the auth modal. That's why `/src/components/v2/ConnectWallet` uses this method too.
We can use `useWeb3Account`, I don't care to be honest since at the end they all point to the same data.

src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
therealemjy: Works for me

coreyar:
> Just to clarify: only the stories for the pages are removed, all the ones for components are staying in (so that as you say we can make sure components are rendering properly). I've removed the page stories because they'll be broken until we eventually replace them one by one as we refactor the pages. I can keep them in if you see value in that.

I think there is value because we are shipping those pages and we don't have storybook coverage on v1 components

How did they break?

coreyar (approved):

PR #268, opened 2022-04-06
Fix: toggle styles

baomastr:


src/components/v2/Toggle/styles.ts
```diff
@@ -1,44 +1,47 @@
...
+
 export const useStyles = () => {
   const theme = useTheme();
   return css`
-    width: 44px;
-    height: 22px;
+    width: calc(${thumbSize} * 2);
+    height: ${thumbSize};
     padding: 0;
     & .MuiSwitch-switchBase {
```
therealemjy, line 12:
```suggestion
    &.MuiSwitch-switchBase {
```

therealemjy (approved):

src/components/v2/Toggle/styles.ts
baomastr: No, the space has been deliberately removed here in order to increase the specificity of the styles.

baomastr: Well, I also want to note that such class inheritance hurts more than the use of native optimization, which mui provides and is also completely incapacitated in the context of other libraries.

therealemjy: You mean the space has been added? Maybe I understand this wrong, but then in this case you can just get rid of the `&`.

therealemjy: By the way this is minor, feel free to merge

PR #270, opened 2022-04-07
Fix Wallet Connect integration

therealemjy:
- upgrade `@web3-react/walletconnect-connector` to `6.2.13`
- only enable Wallet Connect on mainnet (it doesn't support testnet)

### DIFF
--- a/src/components/Basic/ConnectButton.tsx
+++ b/src/components/Basic/ConnectButton.tsx
@@ -319,16 +319,18 @@ function ConnectButton() {
                   <img src={arrowRightImg} alt="arrow" />
                 </div>
 
-                <div
-                  className="flex align-center just-between metamask-connect-btn"
-                  onClick={() => onConnect(ConnectorNames.WalletConnect)}
-                >
-                  <div className="flex align-center">
-                    <img src={walletconnectImg} alt="coinbase wallet" />
-                    <span>WalletConnect</span>
+                {process.env.REACT_APP_CHAIN_ID === '56' && (
+                  <div
+                    className="flex align-center just-between metamask-connect-btn"
+                    onClick={() => onConnect(ConnectorNames.WalletConnect)}
+                  >
+                    <div className="flex align-center">
+                      <img src={walletconnectImg} alt="coinbase wallet" />
+                      <span>WalletConnect</span>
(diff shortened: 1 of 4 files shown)


src/components/Basic/ConnectButton.tsx
```diff
@@ -319,16 +319,18 @@ function ConnectButton() {
...
                 </div>
 
-                <div
-                  className="flex align-center just-between metamask-connect-btn"
-                  onClick={() => onConnect(ConnectorNames.WalletConnect)}
-                >
-                  <div className="flex align-center">
-                    <img src={walletconnectImg} alt="coinbase wallet" />
-                    <span>WalletConnect</span>
+                {process.env.REACT_APP_CHAIN_ID === '56' && (
```
baomastr, line 322: Would be nice to have a short util like `isDev`

src/utilities/connectors.ts
```diff
@@ -12,9 +12,10 @@ const chainId = parseInt(process.env.REACT_APP_CHAIN_ID || '97', 10);
 export const injected = new InjectedConnector({ supportedChainIds: [chainId] });
 
 const walletconnect = new WalletConnectConnector({
-  rpc: { [chainId]: rpcUrl },
+  rpc: { 56: rpcUrl },
```
baomastr, line 15: Shouldn't we move such things to constants/config builder functions?

baomastr (approved):

src/components/Basic/ConnectButton.tsx
therealemjy: We do in `develop`: `isOnTestnet`.
This is just a hotfix so I wanted to write as less code as possible. I've opened a draft PR to introduce this fix too, there it does use the config appropriately: https://github.com/VenusProtocol/venus-protocol-interface/pull/269

src/utilities/connectors.ts
therealemjy: Yeah definitely. Here too, sorry for the confusion, since this is a hotfix I went the easy way. The draft PR I opened to add this fix to `develop` uses the config :)
https://github.com/VenusProtocol/venus-protocol-interface/pull/269

PR #274, opened 2022-04-07
Add "Repay VAI" tab content inside MintRepayVai component

therealemjy:
- create `repayVai` method for API client and create `useRepayVai` mutation hook
- update `AmountForm` so its styling can be extended
- unify functions and hooks inside API client
- simplify `restService`: it was built to handle sending `FormData`, however we don't support that functionality in the dApp, so I simplified the function to just support `JSON` requests
- create `RepayVai` tab component for`MintRepayVai` component:
Screenshot 2022-04-08 at 08 31 35


src/clients/api/mutations/useEnterMarkets.ts
```diff
@@ -0,0 +1,29 @@
+import { useMutation, MutationObserverOptions } from 'react-query';
+
+import { enterMarkets, IEnterMarketsInput, EnterMarketsOutput } from 'clients/api';
+import FunctionKey from 'constants/functionKey';
+import { useComptroller } from 'hooks/useContract';
+
+const useEnterMarkets = (
+  // TODO: use custom error type
```
baomastr: do we have tickets for this TODOs? maybe we can have links in comments to proper tasks?

therealemjy: We don't, that's a good point.
I've created it and added the reference inside the relevant comments 👍

baomastr (approved):

PR #276, opened 2022-04-08
Add Delimeter component

baomastr:
image

### DIFF
--- a/src/components/v2/Delimiter/index.tsx
+++ b/src/components/v2/Delimiter/index.tsx
@@ -1,0 +1,13 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+
+import { useDelimiterStyles as useStyles } from './styles';
+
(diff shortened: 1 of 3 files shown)


src/components/v2/Delimiter/index.tsx
```diff
@@ -0,0 +1,14 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+
+import { useDelimiterStyles as useStyles } from './styles';
+
+interface IDelimiterProps {
+  className?: string;
+  spaceSize?: 3 | 4 | 5;
```
therealemjy: I'd remove this prop. We should extend the delimiter's styling if we want to apply margins to it from a parent component, this way we're consistent with the rest of the codebase.

baomastr: done

therealemjy (approved):

PR #277, opened 2022-04-08
Add tooltips for Slider component

baomastr:
image


src/components/v2/Slider/index.stories.tsx
```diff
@@ -9,19 +9,34 @@ export default {
...
+    onChange={console.log}
+    defaultValue={50}
+    step={5}
     mark={75}
     ariaLabel="Storybook slider"
     min={0}
     max={100}
+    trackTooltip="Storybook tooltip text for Track"
+    markTooltip="Storybook tooltip text for Mark"
+    isDisabled // @TODO: find a way to disable clicks
```
baomastr: got conflict with events.
have to disable clicks but leave hovers.

src/components/v2/Slider/index.tsx
```diff
@@ -33,19 +41,53 @@ export const Slider = ({
...
+
+    return <span {...props} css={styles.mark} />;
+  };
+
+  const renderTrack = (props?: NonNullable<SliderTypeMap['props']['componentsProps']>['track']) => {
+    if (trackTooltip) {
+      return (
+        <div style={props?.style} css={[styles.trackWrapper, styles.hasTooltip]}>
+          <Tooltip title={trackTooltip}>
+            <Box {...props} style={{ width: '100%' }} />
```
therealemjy: Let's not have inline styles

therealemjy (commented):
Just one small comment. Looks good!

src/components/v2/Slider/index.tsx
baomastr: it has to be inlined in order to reassign inline styles. left a comment here.
image

baomastr: passed undefined and left a comment

src/components/v2/Slider/index.stories.tsx
baomastr: solved with passing value prop

src/components/v2/Slider/index.tsx
```diff
@@ -1,20 +1,29 @@
...
   defaultValue: number;
   mark: number;
   step: number;
   ariaLabel: string;
   min: number;
   max: number;
   onChange: (value: number | number[]) => void;
-  isDisabled?: boolean;
+  trackTooltip?: string;
+  markTooltip?: string;
```
therealemjy: Let's get rid of `defaultValue` then, since on our case the slider is never updated manually.

baomastr: so, do you think we shouldn't have ability to change slider in future?

baomastr: it should be just progressBar component then

therealemjy: We don't have any slider in the designs, so yeah it is progress bar :)

baomastr: done

therealemjy (approved):
Nice 👍

PR #278, opened 2022-04-10
Wire up BorrowMarket component

therealemjy:
- Move title outside of `TableContainer` inside `Table` component, so that it isn't horizontally scrollable on mobile when content is too wide to fit in the screen
- Fix liquidity and APY values in `BorrowMarketUi` component
- Wire up `BorrowMarket`

### DIFF
--- a/src/components/v2/Table/index.tsx
+++ b/src/components/v2/Table/index.tsx
@@ -20,13 +20,14 @@ export interface ITableProps {
   title: string;
   data: ITableRowProps[][];
   columns: { key: string; label: string; orderable: boolean }[];
+  rowKeyIndex: number;
   minWidth?: string;
   initialOrder?: {
     orderBy: string;
     orderDirection: 'asc' | 'desc';
   };
   rowOnClick?: (row: ITableRowProps[]) => void;
-  rowKeyIndex: number;
+  className?: string;
 }
 
 export const Table = ({
@@ -37,6 +38,7 @@ export const Table = ({
   initialOrder,
   rowOnClick,
   rowKeyIndex,
+  className,
 }: ITableProps) => {
   const styles = useStyles();
   const [orderBy, setOrderBy] = useState<typeof columns[number]['key'] | undefined>(
@@ -74,36 +76,39 @@ export const Table = ({
   }, [data, orderBy, orderDirection]);
 
   return (
-    <TableContainer css={styles.tableContainer} component={Paper}>
+    <div className={className}>
       <h4 css={styles.title}>{title}</h4>
-      <TableMUI css={styles.table({ minWidth: minWidth ?? '0' })} aria-label={title}>
-        <Head
-          columns={columns}
-          orderBy={orderBy}
-          orderDirection={orderDirection}
-          onRequestOrder={onRequestOrder}
-        />
 
-        {/* TODO: add loading state */}
+      <TableContainer css={styles.tableContainer} component={Paper}>
+        <TableMUI css={styles.table({ minWidth: minWidth ?? '0' })} aria-label={title}>
+          <Head
+            columns={columns}
+            orderBy={orderBy}
+            orderDirection={orderDirection}
+            onRequestOrder={onRequestOrder}
+          />
 
-        {/* TODO: add error state */}
+          {/* TODO: add loading state */}
 
-        <TableBody>
-          {rows.map(row => (
-            <TableRow
-              hover
-              key={row[rowKeyIndex].value.toString()}
-              onClick={() => rowOnClick && rowOnClick(row)}
-            >
-              {row.map(({ key, render }: ITableRowProps) => (
-                <TableCell key={uid(key)}>
-                  <div>{render()}</div>
-                </TableCell>
-              ))}
-            </TableRow>
-          ))}
-        </TableBody>
-      </TableMUI>
-    </TableContainer>
+          {/* TODO: add error state */}
+
+          <TableBody>
+            {rows.map(row => (
+              <TableRow
+                hover
+                key={row[rowKeyIndex].value.toString()}
+                onClick={() => rowOnClick && rowOnClick(row)}
+              >
+                {row.map(({ key, render }: ITableRowProps) => (
+                  <TableCell key={uid(key)}>
+                    <div>{render()}</div>
+                  </TableCell>
+                ))}
+              </TableRow>
+            ))}
+          </TableBody>
+        </TableMUI>
+      </TableContainer>
+    </div>
   );
 };
--- a/src/pages/Dashboard/Markets/BorrowMarket/index.tsx
+++ b/src/pages/Dashboard/Markets/BorrowMarket/index.tsx
@@ -6,9 +6,10 @@ import {
   formatApy,
 } from 'utilities/common';
 import { Asset, TokenSymbol } from 'types';
-import { CONTRACT_TOKEN_ADDRESS } from 'constants/contracts';
 import { Token } from 'components/v2/Token';
 import { Table, ITableProps } from 'components/v2/Table';
+import useUserMarketInfo from 'hooks/useUserMarketInfo';
+import { AuthContext } from 'context/AuthContext';
 import { useStyles } from '../styles';
 
 export interface IBorrowMarketUiProps {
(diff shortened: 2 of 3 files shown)


coreyar:
Two storybooks fail
```
FAIL chrome.app/chrome.ipad/Pages/Transaction/Transaction: 1 request failed to load; https://testnetapi.venus.io/apiundefined

FAIL chrome.app/chrome.ipad/Pages/Vote/Vote: 1 request failed to load; https://testnetapi.venus.io/apiundefined
```

src/clients/api/mutations/useEnterMarkets.ts
```diff
@@ -0,0 +1,29 @@
...
+  options?: MutationObserverOptions<
+    EnterMarketsOutput,
+    // @TODO: use custom error type (see https://app.clickup.com/t/2rvwhnt)
+    Error,
+    Omit<IEnterMarketsInput, 'comptrollerContract'>
+  >,
+) => {
+  const comptrollerContract = useComptroller();
+
+  // @TODO: invalidate related queries on success
```
coreyar: We can remove this because there is a PR open - https://github.com/VenusProtocol/venus-protocol-interface/pull/280

src/clients/api/mutations/useEnterMarkets.ts
```diff
@@ -0,0 +1,29 @@
+import { useMutation, MutationObserverOptions } from 'react-query';
+
+import { enterMarkets, IEnterMarketsInput, EnterMarketsOutput } from 'clients/api';
+import FunctionKey from 'constants/functionKey';
+import { useComptroller } from 'hooks/useContract';
+
+const useEnterMarkets = (
+  options?: MutationObserverOptions<
+    EnterMarketsOutput,
+    // @TODO: use custom error type (see https://app.clickup.com/t/2rvwhnt)
```
coreyar: I think if we are going to leave TODOs having a ticket and pointing to it is the right way to do it

src/clients/api/mutations/useExitMarket.ts
```diff
@@ -0,0 +1,29 @@
...
+  // @TODO: use custom error type (see https://app.clickup.com/t/2rvwhnt)
+  options?: MutationObserverOptions<
+    ExitMarketOutput,
+    Error,
+    Omit<IExitMarketInput, 'comptrollerContract'>
+  >,
+) => {
+  const comptrollerContract = useComptroller();
+
+  // @TODO: invalidate related queries on success
```
coreyar: I don't think we need this comment because the PR is open and requires adjustments

src/clients/api/mutations/useRepayVai.ts
```diff
@@ -0,0 +1,29 @@
...
+type Options = MutationObserverOptions<
+  IRepayVaiOutput,
+  Error,
+  Omit<IRepayVaiInput, 'vaiControllerContract'>
+>;
+
+const useRepayVai = (options?: Options) => {
+  const vaiControllerContract = useVaiUnitroller();
+
+  // @TODO: invalidate queries related to fetching the user VAI balance and
```
coreyar: Let's do this or add a ticket

src/clients/api/mutations/useEnterMarkets.ts
therealemjy: Yeah I think that adds information to it. Dmitriy suggested that 👍

src/clients/api/queries/useGetHypotheticalLiquidityQueries.ts
```diff
@@ -1,21 +1,22 @@
...
+
 import getHypotheticalAccountLiquidity from 'clients/api/queries/getHypotheticalAccountLiquidity';
-import { IGetVTokenBalancesAllOutput } from 'clients/api/queries/getVTokenBalancesAll';
+import { IGetVTokenBalancesAllOutput } from 'clients/api';
 import { useComptroller } from 'hooks/useContract';
 import FunctionKey from 'constants/functionKey';
 import { Asset } from 'types';
 
-type Options = QueryObserverOptions;
+type Options = QueryObserverOptions; // @TODO: add query specific generics
```
coreyar: I looked into this and the types that I could find in the library wanted options to be an array, which we are using an object. At the time I didn't have internet access so I was only referencing the installed types. Hopefully there is a better solution

src/clients/api/mutations/useEnterMarkets.ts
therealemjy: I think we should keep it in until either PR is merged in, this way if yours gets merged in first I'll just pull the changes and if mine gets merged in first we keep the TODO and don't forget to update this part.

src/clients/api/mutations/useExitMarket.ts
therealemjy: Same thing as previous comment

src/clients/api/mutations/useRepayVai.ts
therealemjy: We can't invalidate the queries yet because they don't use React Query at the moment (we're still relying on `useVaiUser` for the time being).
I'll create a ticket and reference it there 👍

src/pages/Dashboard/Markets/SupplyMarket.tsx
```diff
@@ -127,7 +126,7 @@ const SupplyMarket: React.FC = () => {
       throw error;
     },
   });
-  const comptrollerContract = useComptroller();
```
coreyar: I literally just opened a PR updating this 😂

therealemjy:
Yeah, that's what I mentioned in one of my PRs when you wanted to have the old pages in Storybook despite the mock breaking them.
As I mentioned earlier, all the current pages will be broken until we refactor them, which is why I'd suggest removing them from Storybook altogether. They don't test the old components, because they don't render at all. Fixing the mocks so the pages render would take some efforts and in my opinion we shouldn't focus on this right now since we're refactoring these pages.

src/pages/Dashboard/Markets/SupplyMarket.tsx
therealemjy: I know, I've asked you multiple times to review the PRs lol

src/pages/Dashboard/MintRepayVai/RepayVai/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const { userVaiMinted, userVaiBalance } = useVaiUser();
+
+  const { mutate: repayVai, isLoading: isRepayVaiLoading } = useRepayVai({
+    onError: error => {
+      toast.error({ title: error.message });
+    },
+    onSuccess: (_data, variables) => {
+      // @TODO: display success modal instead of toast once it's been
+      // implemented
+      toast.success({
```
coreyar: A strategy I've been following is to keep the container pure of all UI interactions. This is why I introduced the toastError. So the container could throw the error and we catch it and handle the appropriate UI updates.

I think this should also make testing nicer because we can test for errors on the containers and not UI interactions.

src/pages/Dashboard/MintRepayVai/RepayVai/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const onSubmit: IRepayVaiUiProps['onSubmit'] = amountWei => {
+    if (account) {
+      repayVai({
+        fromAccountAddress: account.address,
+        amountWei,
+      });
+    }
+  };
+
+  // @TODO: wrap with EnableToken component once created
```
coreyar: I would wrap the UI element directly not the returned UI element

src/pages/Dashboard/MintRepayVai/RepayVai/index.tsx
therealemjy: Ok I see, that's a nice pattern.

Since our fetch functions already throw errors, I can use `mutateAsync` inside the UI component here to catch errors and display a toast (on success too).

src/pages/Dashboard/MintRepayVai/RepayVai/index.tsx
therealemjy: I wouldn't; ideally the UI component is pure, this would bring side effect to it. It also makes it harder to render in Storybook and harder to test.

src/pages/Dashboard/MintRepayVai/RepayVai/index.tsx
therealemjy: By the way, this PR originally branched out of another PR I was working and that I needed the work of. Now that's it's been merged in and this PR is rebased with `develop`, these changes no longer appear.
I've made the changes related to your comment in that PR: https://github.com/VenusProtocol/venus-protocol-interface/pull/282

src/clients/api/queries/useGetHypotheticalLiquidityQueries.ts
therealemjy: Yeah might be worth looking at the documentation

coreyar:
If we aren't going to fix them we should remove those storybooks or skip them with loki so that we don't clutter test results with expected failures and prevent snapshots from running on the rest of the components. Other than that this PR looks good

therealemjy:
Yeah, check out: https://github.com/VenusProtocol/venus-protocol-interface/pull/287

coreyar (approved):

PR #280, opened 2022-04-10
Refetch assets on collateral retoggle

coreyar:
When we toggle collateral refetch the users assets so we can check that the token is added to their wallet.

### DIFF
--- a/src/clients/api/mutations/useEnterMarkets.ts
+++ b/src/clients/api/mutations/useEnterMarkets.ts
@@ -1,6 +1,6 @@
 import { useMutation, MutationObserverOptions } from 'react-query';
 
-import { enterMarkets, IEnterMarketsInput, EnterMarketsOutput } from 'clients/api';
+import { queryClient, enterMarkets, IEnterMarketsInput, EnterMarketsOutput } from 'clients/api';
 import FunctionKey from 'constants/functionKey';
 import { useComptroller } from 'hooks/useContract';
 
@@ -22,7 +22,19 @@ const useEnterMarkets = (
         comptrollerContract,
         ...params,
       }),
-    options,
+    {
+      ...options,
+      onSuccess: (
+        data: void,
+        variables: Omit<IEnterMarketsInput, 'comptrollerContract'>,
+        context: unknown,
+      ) => {
+        queryClient.invalidateQueries(FunctionKey.GET_ASSETS_IN_ACCOUNT);
+        if (options?.onSuccess) {
+          options.onSuccess(data, variables, context);
+        }
+      },
+    },
   );
 };
 
--- a/src/clients/api/mutations/useExitMarket.ts
+++ b/src/clients/api/mutations/useExitMarket.ts
@@ -1,6 +1,6 @@
 import { useMutation, MutationObserverOptions } from 'react-query';
 
-import { exitMarket, IExitMarketInput, ExitMarketOutput } from 'clients/api';
+import { queryClient, exitMarket, IExitMarketInput, ExitMarketOutput } from 'clients/api';
 import FunctionKey from 'constants/functionKey';
 import { useComptroller } from 'hooks/useContract';
 
@@ -22,7 +22,19 @@ const useExitMarket = (
         comptrollerContract,
         ...params,
       }),
-    options,
+    {
+      ...options,
+      onSuccess: (
+        data: void,
+        variables: Omit<IExitMarketInput, 'comptrollerContract'>,
+        context: unknown,
+      ) => {
+        queryClient.invalidateQueries(FunctionKey.GET_ASSETS_IN_ACCOUNT);
+        if (options?.onSuccess) {
+          options.onSuccess(data, variables, context);
+        }
+      },
+    },
   );
 };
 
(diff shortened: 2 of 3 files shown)


src/pages/Dashboard/Markets/SupplyMarket.tsx
```diff
@@ -114,14 +116,20 @@ const SupplyMarket: React.FC = () => {
   const userMarketInfo = useUserMarketInfo({ account });
   const [confirmCollateral, setConfirmCollateral] = useState<Asset | undefined>(undefined);
   const { mutate: enterMarkets } = useEnterMarkets({
-    onSuccess: () => setConfirmCollateral(undefined),
+    onSuccess: () => {
+      queryClient.invalidateQueries(FunctionKey.GET_ASSETS_IN_ACCOUNT);
```
therealemjy: I'd suggest invalidating the query directly from the `useEnterMarkets` hook, this way the hook becomes also responsible for handling cache (so if we import this mutation from somewhere else, cache will be handled out of the box).
This pattern will become particularly interesting if we get into optimistic updates, or even just updating the cache on the fly after a successful mutation.

src/pages/Dashboard/Markets/SupplyMarket.tsx
```diff
@@ -114,14 +116,20 @@ const SupplyMarket: React.FC = () => {
...
+    },
     onError: error => {
       setConfirmCollateral(undefined);
       throw error;
     },
   });
   const { mutate: exitMarkets } = useExitMarket({
-    onSuccess: () => setConfirmCollateral(undefined),
+    onSuccess: () => {
+      queryClient.invalidateQueries(FunctionKey.GET_ASSETS_IN_ACCOUNT);
```
therealemjy: Same thing here about calling this from the `useExitMarket` hook directly.

src/pages/Dashboard/Markets/SupplyMarket.tsx
coreyar: That makes sense. It brings me back to a thought I had that all options should be included in the hook because it seems like all options should belong in the hook with the exception of acceptable overrides

therealemjy: So the way I see it is that we should assign all the options related to the hook inside it directly, but keep the option of overriding them as you mentioned.
I wouldn't filter what options can be overridden or not, since some very specific cases will sometimes need to override options that were defined in the hook directly.

coreyar: In that case I think we should have a merge strategy so that everything gets merged as desired. Otherwise we'll have lots of duplicated code managing merging options

therealemjy: I don't know if we can generalize the merging strategy though, as some components won't extend the options at all and just pass the ones they receive, while some will have custom behaviors like in your case:
```typescript
const useCustomQuery = (options?: Options) => (
  useQuery(
    FunctionKey.REPAY_VAI,
    fetchFn,
    {
      ...(options || {})
      onSuccess: (...props) => {
        // Invalid queries here

        if (options?.onSucess) {
          onSuccess(...props);
        }
      }
    },
  )
);
```

coreyar: I think it would work actually including the situation you outlined. The situations where it wouldn't work are if you wanted to override default values on functions, for example override the default onSuccess completely and not just call an additional function. That would get complicated and brings me back to not passing the whole options object to the hook but only exposing the options that can be customized because I think the smaller argument surface area the better.

That being said I am going to make updates to this PR similar to what I did [here](https://github.com/VenusProtocol/venus-protocol-interface/pull/285/files#diff-81dafb448987ab30820a67b8571c722c176c0f73c5836d6238b3b7641b221373)

therealemjy: I see what you mean, you're right.
I think for now let's go with the approach you've followed. If later on we encounter a lot of cases where we need to override the props completely, we can think of a different merge strategy as you proposed.

src/clients/api/mutations/useEnterMarkets.ts
```diff
@@ -22,7 +22,19 @@ const useEnterMarkets = (
...
+      ...options,
+      onSuccess: (
+        data: void,
+        variables: Omit<IEnterMarketsInput, 'comptrollerContract'>,
+        context: unknown,
+      ) => {
+        if (options?.onSuccess) {
+          options.onSuccess(data, variables, context);
+        }
+        queryClient.invalidateQueries(FunctionKey.GET_ASSETS_IN_ACCOUNT);
```
therealemjy: I'd call `options.onSuccess` after the queries invalidation, since the mutation has only finished executing after the queries have been invalidated.

src/clients/api/mutations/useExitMarket.ts
```diff
@@ -22,7 +22,19 @@ const useExitMarket = (
...
+      ...options,
+      onSuccess: (
+        data: void,
+        variables: Omit<IExitMarketInput, 'comptrollerContract'>,
+        context: unknown,
+      ) => {
+        if (options?.onSuccess) {
+          options.onSuccess(data, variables, context);
+        }
+        queryClient.invalidateQueries(FunctionKey.GET_ASSETS_IN_ACCOUNT);
```
therealemjy: Same thing here about the order of actions

therealemjy (approved):

PR #282, opened 2022-04-10
Add tests for MintRepayVai component

therealemjy:
- add mock for api client
- add tests for `MintRepayVai` component and its subcomponents
- decouple components inside `Market` folder of `Dashboard`


coreyar (approved):

coreyar:
Looks good. Linting and storybooks need to pass though

therealemjy:
Cool, thanks for the review.
I've just pushed changes to fix the linting issue.

About the issue with Storybook, the only way we can make them pass is to remove the old pages that are now broken, or fix them. But as I mentioned earlier I don't fix we should spend time on fixing them now since we're refactoring them.

therealemjy:
I opened a PR to fix the issue with Storybook: https://github.com/VenusProtocol/venus-protocol-interface/pull/287.
Since this is separate from this PR, I'll merge it in.

PR #283, opened 2022-04-11
Create translation client

therealemjy:
- remove `react-intl` library
- create translation client using  `react-i18next`
- internationalize texts of `AuthModal` and `BscLink` components
- upgrade node engine to `16.x.x`
- create `.nvmrc` file specifying usage of node `v16.14.2`
- update node version used Github actions to `16`


src/clients/translation/index.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
```
coreyar: nit: I think of clients as interacting with services outside of the browser, with that definition this wouldn't belong here. The translation code is spread between here and assets when they could alternatively be grouped in one folder.

Main takeaway though is what is the definition of the clients folder?

src/clients/translation/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+
+// Only use this function when you need to render a string from outside a
+// component. Otherwise, use the t function or Trans component returned by the
+// useTranslation hook.
+export const t = (params: TFunctionKeys, values?: Record<string, unknown>) =>
+  i18next.t(params, values);
+
+// Function to check whether a translation exists
+export const translationExists: ExistsFunction = params => i18next.exists(params);
```
coreyar: I'm curious what the use case for this would be? When would we display something different then the default translation if it is missing?

src/clients/translation/index.tsx
therealemjy: Got you. About the translations, I hesitated but then I figured since we place all the images used in different pages/components inside the `assets` folder, then translations would technically receive the same treatment since they are the texts contained in all pages/components.
That said, I don't mind having them inside the client too.

I see clients as local libraries personally, so they could contain anything: functions, classes, hooks, components etc... In this case, I'd rather have the `Trans` component inside the client itself, because it relies on the `Trans` component imported from i18-next. This way, the usage of i18next is restricted to our translation client only, making it easier to replace in the future if needed.

src/containers/App.tsx
```diff
@@ -33,81 +30,51 @@ import { VaiContextProvider } from '../context/VaiContext';
...
-                </RefreshContextProvider>
-              </Provider>
-            </QueryClientProvider>
-          </Web3Wrapper>
-        </IntlProvider>
-      </Theme>
-    );
-  }
-}
+const App: React.FC = () => (
```
coreyar: In this file we can just delete `react-intl` and then in `pages/App.tsx` we can setup the `react-i18next`. This file will be eventually deleted and we'll remove the flag for running in dev and the root of the application will be `src/pages/App.tsx`

therealemjy: Ok got you, that makes sense. I've made the update.

src/clients/translation/index.tsx
therealemjy: To be honest, your comment made me rethink about it and I think we should get rid of it.
I thought we could have cases when the translation key comes from a backend and so we'd need to check it exists before we call the `t` function, but that doesn't really apply actually since our contracts don't use error codes or anything we can use to identify errors and get their translations from our locales.

coreyar (approved):

PR #285, opened 2022-04-11
Add Enable token component

coreyar:
Add Enable token component that renders its children if enabled or an prompt to enable to the token if it is not enabled

### DIFF
--- a/src/clients/api/index.ts
+++ b/src/clients/api/index.ts
@@ -21,6 +21,10 @@ export { default as exitMarket } from './mutations/exitMarket';
 export * from './mutations/exitMarket';
 export { default as useExitMarket } from './mutations/useExitMarket';
 
+export { default as approveToken } from './mutations/approveToken';
+export * from './mutations/approveToken';
+export { default as useApproveToken } from './mutations/useApproveToken';
+
 // Queries
 export { default as getVaiTreasuryPercentage } from './queries/getVaiTreasuryPercentage';
 export * from './queries/getVaiTreasuryPercentage';
--- a/src/clients/api/mutations/approveToken.ts
+++ b/src/clients/api/mutations/approveToken.ts
@@ -1,0 +1,20 @@
+import BigNumber from 'bignumber.js';
+
+export interface IApproveTokenInput {
+  tokenContract: $TSFixMe; // @TODO: use contract type (through Typechain?)
+  account: string | undefined;
+  vtokenAddress: string;
+  allowance?: string;
+}
+
+export type ApproveTokenOutput = void;
+
+const approveToken = ({
+  tokenContract,
+  account,
+  vtokenAddress,
+  allowance = new BigNumber(2).pow(256).minus(1).toString(10),
+}: IApproveTokenInput): Promise<ApproveTokenOutput> =>
+  tokenContract.methods.approve(vtokenAddress, allowance).send({ from: account });
+
+export default approveToken;
--- a/src/clients/api/mutations/useApproveToken.ts
+++ b/src/clients/api/mutations/useApproveToken.ts
@@ -1,0 +1,35 @@
+import { useMutation, MutationObserverOptions } from 'react-query';
+import { queryClient, approveToken, IApproveTokenInput, ApproveTokenOutput } from 'clients/api';
+import FunctionKey from 'constants/functionKey';
+import { useToken } from 'hooks/useContract';
+
+const useApproveToken = (
+  { assetId }: { assetId: string },
+  // TODO: use custom error type https://app.clickup.com/t/2rvwhnt
+  options?: MutationObserverOptions<
+    ApproveTokenOutput,
+    Error,
+    Omit<IApproveTokenInput, 'tokenContract'>
+  >,
+) => {
+  const tokenContract = useToken(assetId);
+  return useMutation(
+    FunctionKey.APPROVE_TOKEN,
+    params =>
+      approveToken({
+        tokenContract,
+        ...params,
+      }),
+    {
+      ...options,
+      onSuccess: (data, variables, context) => {
+        queryClient.invalidateQueries(FunctionKey.GET_VTOKEN_BALANCES_ALL);
+        if (options?.onSuccess) {
+          options.onSuccess(data, variables, context);
+        }
+      },
+    },
+  );
+};
+
+export default useApproveToken;
--- a/src/components/v2/EnableToken/index.tsx
+++ b/src/components/v2/EnableToken/index.tsx
@@ -1,9 +1,11 @@
 /** @jsxImportSource @emotion/react */
-import React from 'react';
+import React, { useContext } from 'react';
 import BigNumber from 'bignumber.js';
 import Typography from '@mui/material/Typography';
+import { AuthContext } from 'context/AuthContext';
 import { TokenSymbol } from 'types';
 import { formatApy } from 'utilities/common';
+import useApproveToken from 'clients/api/mutations/useApproveToken';
 import { Icon, IconName } from '../Icon';
 import { PrimaryButton } from '../Button';
 import useStyles from './styles';
@@ -13,19 +15,23 @@ export interface IEnableTokenProps {
   isEnabled: boolean;
   title: string;
   tokenInfo: { symbol: TokenSymbol; text: string; apy: BigNumber | number }[];
-  children: React.ReactElement;
+  approveToken: () => void;
+  vtokenAddress: string;
+  disabled: boolean;
 }
 
-export const EnableToken: React.FC<IEnableTokenProps> = ({
+export const EnableTokenUi: React.FC<Omit<IEnableTokenProps, 'vtokenAddress'>> = ({
   symbol,
   title,
   tokenInfo,
   isEnabled,
   children,
-}: IEnableTokenProps) => {
+  approveToken,
+  disabled,
+}) => {
   const styles = useStyles();
   if (isEnabled) {
-    return children;
+    return <>{children}</>;
   }
   return (
     <div css={styles.container}>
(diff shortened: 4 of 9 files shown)


src/components/v2/EnableToken/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export interface IEnableTokenProps {
+  symbol: TokenSymbol;
+  isEnabled: boolean;
+  title: string;
+  tokenInfo: { symbol: TokenSymbol; text: string; apy: BigNumber | number }[];
+  children: React.ReactElement;
+  approveToken: UseMutateFunction<void, Error, Omit<IApproveTokenInput, 'tokenContract'>, unknown>;
+  vtokenAddress: string;
+  account: string | null;
```
therealemjy: This is a small thing, but I think we should remove the `account` prop and instead have `disabled` prop.
The UI component technically doesn't care whether an account is connected or not, it only needs to know whether it can submit its form or not.

src/components/v2/EnableToken/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { Icon, IconName } from '../Icon';
+import { PrimaryButton } from '../Button';
+import useStyles from './styles';
+
+export interface IEnableTokenProps {
+  symbol: TokenSymbol;
+  isEnabled: boolean;
+  title: string;
+  tokenInfo: { symbol: TokenSymbol; text: string; apy: BigNumber | number }[];
+  children: React.ReactElement;
```
therealemjy: No need to declare `children`, `React.FC` declares it already.

src/components/v2/EnableToken/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    </div>
+  );
+};
+
+const EnableToken = ({
+  symbol,
+  ...rest
+}: Omit<IEnableTokenProps, 'approveToken' | 'account' | 'approveToken'>) => {
+  const { mutate: approveToken } = useApproveToken(symbol);
+  const { account = '' } = useWeb3Account();
```
therealemjy: Let's import the account using the `AuthContext`.
Also, you don't need to assign a default value to account, you UI component accepts an `undefined` value.

src/clients/api/mutations/useApproveToken.ts
```diff
@@ -0,0 +1,44 @@
+import { useMutation, MutationObserverOptions } from 'react-query';
+import { queryClient, approveToken, IApproveTokenInput, ApproveTokenOutput } from 'clients/api';
+import FunctionKey from 'constants/functionKey';
+import { useToken } from 'hooks/useContract';
+
+const useApproveToken = (
+  assetId: string,
```
therealemjy: As a rule of thumb I'd always use an object to declare the params of a query/mutation hook. I feel it makes it more explicit what each argument corresponds to:
```typescript
useApproveToken({ assetId: 'vai' });
```

src/clients/api/mutations/useApproveToken.ts
```diff
@@ -0,0 +1,44 @@
...
+  // TODO: use custom error type
```
therealemjy: You can add the link to the ticket: https://app.clickup.com/t/2rvwhnt

src/clients/api/mutations/useApproveToken.ts
```diff
@@ -0,0 +1,44 @@
...
+      ...options,
+      onSuccess: (
+        data: void,
+        variables: Omit<IApproveTokenInput, 'tokenContract'>,
+        context: unknown,
+      ) => {
+        if (options?.onSuccess) {
+          options.onSuccess(data, variables, context);
+        }
+        queryClient.invalidateQueries(FunctionKey.GET_VTOKEN_BALANCES_ACCOUNT);
```
therealemjy: I'd call the success callback only after the queries have been invalidated (since it's only then that the mutation call is truly done).

src/clients/api/index.ts
```diff
@@ -21,6 +21,9 @@ export { default as exitMarket } from './mutations/exitMarket';
 export * from './mutations/exitMarket';
 export { default as useExitMarket } from './mutations/useExitMarket';
 
+export { default as approveToken } from './mutations/approveToken';
+export * from './mutations/approveToken';
+
```
therealemjy, line 27: You need to add the hook too:
```suggestion
export { default as useApproveToken } from './mutations/useApproveToken';
```

src/clients/api/mutations/approveToken.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const enterMarkets = ({
+  tokenContract,
+  account,
+  vtokenAddress,
+  allowance = new BigNumber(2).pow(256).minus(1).toString(10),
+}: IApproveTokenInput): Promise<ApproveTokenOutput> =>
+  tokenContract.methods.approve(vtokenAddress, allowance).send({ from: account });
+
+export default enterMarkets;
```
therealemjy:
```suggestion
const approveToken = ({
  tokenContract,
  account,
  vtokenAddress,
  allowance = new BigNumber(2).pow(256).minus(1).toString(10),
}: IApproveTokenInput): Promise<ApproveTokenOutput> =>
  tokenContract.methods.approve(vtokenAddress, allowance).send({ from: account });

export default approveToken;
```

src/clients/api/mutations/approveToken.ts
```diff
@@ -0,0 +1,20 @@
+import BigNumber from 'bignumber.js';
+
+export interface IApproveTokenInput {
+  tokenContract: $TSFixMe; // @TODO: use contract type (through Typechain?)
+  account: string | undefined | null;
```
therealemjy: Can we avoid using `null`? it has some terrible attributes, like being considered an object 😅

src/components/v2/EnableToken/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { PrimaryButton } from '../Button';
+import useStyles from './styles';
+
+export interface IEnableTokenProps {
+  symbol: TokenSymbol;
+  isEnabled: boolean;
+  title: string;
+  tokenInfo: { symbol: TokenSymbol; text: string; apy: BigNumber | number }[];
+  children: React.ReactElement;
+  approveToken: UseMutateFunction<void, Error, Omit<IApproveTokenInput, 'tokenContract'>, unknown>;
```
therealemjy: To make this component even more modular, I'd avoid linking its type to React Query or even the function itself.
Technically, the component has its own prop types it needs to make the UI work, and the container is in charge of making sure it passes props in the correct format. So I'd just declare `approveToken` like that:
```typescript
approveToken: (vtokenAddress: string) => Promise<void>;
```

This way you also don't need to pass the account to the function, the container can do that for you.

src/components/v2/EnableToken/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+        Enable
+      </PrimaryButton>
+    </div>
+  );
+};
+
+const EnableToken = ({
+  symbol,
+  ...rest
+}: Omit<IEnableTokenProps, 'approveToken' | 'account' | 'approveToken'>) => {
```
therealemjy:
```suggestion
}: Omit<IEnableTokenProps, 'approveToken' | 'account'>) => {
```

src/clients/api/mutations/useApproveToken.ts
```diff
@@ -0,0 +1,44 @@
...
+  options?: MutationObserverOptions<
+    ApproveTokenOutput,
+    Error,
+    Omit<IApproveTokenInput, 'tokenContract'>
+  >,
+) => {
+  const tokenContract = useToken(assetId);
+  return useMutation(
+    FunctionKey.APPROVE_TOKEN,
+    (params: Omit<IApproveTokenInput, 'tokenContract'>) =>
```
therealemjy: So I was looking into this, and it turns out React Query already infers the types by itself!
```suggestion
    (params) =>
```

src/clients/api/mutations/useApproveToken.ts
```diff
@@ -0,0 +1,44 @@
...
+        tokenContract,
+        ...params,
+      }),
+    {
+      ...options,
+      onSuccess: (
+        data: void,
+        variables: Omit<IApproveTokenInput, 'tokenContract'>,
+        context: unknown,
+      ) => {
```
therealemjy: Same thing here, React Query infers the types from `approveToken`:
```suggestion
      onSuccess: (data, variables, context) => {
```

src/clients/api/mutations/useApproveToken.ts
```diff
@@ -0,0 +1,44 @@
...
+        if (options?.onSuccess) {
+          options.onSuccess(data, variables, context);
+        }
+        queryClient.invalidateQueries(FunctionKey.GET_VTOKEN_BALANCES_ACCOUNT);
+      },
+      onError: (
+        error: Error,
+        variables: Omit<IApproveTokenInput, 'tokenContract'>,
+        context: unknown,
+      ) => (options?.onError ? options.onError(error, variables, context) : undefined),
```
therealemjy: Since you're already spreading the options, you don't need to add this logic.
```suggestion
```

therealemjy (commented):
Thanks for the effort to make it as close to the designs as possible 👍

src/clients/api/mutations/approveToken.ts
coreyar: Yes, this is the type coming from `useWeb3React` and will be replaced by the type returned by the AuthContext

src/components/v2/EnableToken/index.tsx
coreyar: This is required because I am directly returning children. The return type of React.FC doesn't like returning children directly.

therealemjy: Ah yeah, that's because you need to wrap it in a fragment when you return it (so it's detected as valid JSX):
```
<>{children}</>
```

src/components/v2/EnableToken/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+  if (isEnabled) {
+    return <>{children}</>;
+  }
+  return (
+    <div css={styles.container}>
+      <Icon name={symbol as IconName} css={styles.mainLogo} />
+      <Typography component="h3" variant="h3" css={styles.mainText}>
+        {title}
+      </Typography>
+      <hr />
```
therealemjy, line 42: We have a separate component for that: `Delimiter`. It might not be necessary though

therealemjy (approved):

PR #286, opened 2022-04-11
Move contract argument to market hooks

coreyar:
Move the contract argument to inside the market hook so that it doesn't have to be passed as an argument when using the hook


therealemjy:
Hey man,
Check the PRs, this is already done on: https://github.com/VenusProtocol/venus-protocol-interface/pull/278/files#diff-9233b7c99e6968c4123d38da8f37b3a8c4492c53f21e3efb6bb361065e78bb33

PR #292, opened 2022-04-12
Move texts to content json

baomastr:
Added texts for v2 components into locale json file

### DIFF
--- a/src/components/v2/Layout/Header/index.tsx
+++ b/src/components/v2/Layout/Header/index.tsx
@@ -1,27 +1,34 @@
 /** @jsxImportSource @emotion/react */
 import React, { useMemo } from 'react';
-import { RouteComponentProps, withRouter } from 'react-router-dom';
+import { useLocation } from 'react-router-dom';
 import AppBar from '@mui/material/AppBar';
 import Box from '@mui/material/Box';
+
+import { useTranslation } from 'translation';
 import { Toolbar } from '../Toolbar';
 import ConnectButton from './ConnectButton';
 import { XvsCoinInfo, VaiCoinInfo } from '../CoinInfo';
-import { useStyles } from './styles';
 import { menuItems } from '../constants';
+import { useStyles } from './styles';
 
-type HeaderProps = RouteComponentProps;
-
-const Header = ({ location }: HeaderProps) => {
-  const title = useMemo(
-    () => menuItems.find(item => item.href === location.pathname)?.text,
-    [location.pathname],
-  );
+const Header = () => {
+  const { pathname } = useLocation();
   const styles = useStyles();
+  const { t } = useTranslation();
+
+  const title = useMemo(() => {
+    const currentItemKey = menuItems.find(item => item.href === pathname)?.i18nKey;
+    if (!currentItemKey) {
+      console.error(`missed key for translation: ${pathname}`);
+      return null;
+    }
+    return t(currentItemKey);
+  }, [pathname]);
+
   return (
     <AppBar position="relative" css={styles.appBar}>
       <Toolbar css={styles.toolbar}>
-        <h3>{title}</h3>
-
+        {title && <h3>{title}</h3>}
         <Box
           flexDirection="row"
           display="flex"
@@ -31,11 +38,11 @@ const Header = ({ location }: HeaderProps) => {
         >
           <XvsCoinInfo css={styles.rightItemPaper} className="coinInfo" />
           <VaiCoinInfo css={styles.rightItemPaper} className="coinInfo" />
-          <ConnectButton css={styles.rightItemPaper} title="Connect wallet" />
+          <ConnectButton css={styles.rightItemPaper} title={t('header.connectButton.title')} />
         </Box>
       </Toolbar>
     </AppBar>
   );
 };
 
-export default withRouter(Header);
+export default Header;
--- a/src/components/v2/Layout/Sidebar/index.tsx
+++ b/src/components/v2/Layout/Sidebar/index.tsx
@@ -7,16 +7,18 @@ import ListItemIcon from '@mui/material/ListItemIcon';
 import Drawer from '@mui/material/Drawer';
 import Typography from '@mui/material/Typography';
 import Menu from '@mui/material/Menu';
+
 import { ReactComponent as LogoDesktop } from 'assets/img/v2/venusLogoWithText.svg';
 import { ReactComponent as LogoNoText } from 'assets/img/v2/venusLogoPure.svg';
 import { ReactComponent as LogoMobile } from 'assets/img/v2/venusLogoMobile.svg';
 import { useWeb3Account } from 'clients/web3';
+import { useTranslation } from 'translation';
 import { XvsCoinInfo, VaiCoinInfo } from '../CoinInfo';
 import { Toolbar } from '../Toolbar';
 import { Icon } from '../../Icon';
-import { useStyles } from './styles';
 import { menuItems } from '../constants';
 import ConnectButton from '../Header/ConnectButton';
+import { useStyles } from './styles';
 
 interface ISidebarProps {
   account: undefined | null | string;
@@ -25,6 +27,8 @@ interface ISidebarProps {
 export const SidebarUi: React.FC<ISidebarProps> = ({ account }) => {
   const [anchorEl, setAnchorEl] = useState<Element | null>(null);
   const open = Boolean(anchorEl);
+  const { t } = useTranslation();
+  const styles = useStyles();
 
   const openMenu = (event: React.MouseEvent) => {
     setAnchorEl(event.currentTarget);
@@ -32,7 +36,7 @@ export const SidebarUi: React.FC<ISidebarProps> = ({ account }) => {
   const closeMenu = () => {
     setAnchorEl(null);
   };
-  const styles = useStyles();
+
   return (
     <>
       <Drawer variant="permanent" css={styles.drawer}>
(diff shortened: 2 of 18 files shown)


src/components/v2/Layout/Header/index.tsx
```diff
@@ -18,10 +19,12 @@ export const Header = ({ location }: HeaderProps) => {
     [location.pathname],
   );
   const styles = useStyles();
+  const { t } = useTranslation();
+
   return (
     <AppBar position="relative" css={styles.appBar}>
       <Toolbar css={styles.toolbar}>
-        <h3>{title}</h3>
+        {title && <h3>{t(title)}</h3>}
```
therealemjy: In this case, the title would be passed as a string directly so no need to wrap it with `t()`

src/components/v2/Layout/constants.tsx
```diff
@@ -10,37 +10,37 @@ interface IMenuItem {
 export const menuItems: IMenuItem[] = [
   {
     href: '/dashboard',
-    text: 'Dashboard',
+    text: 'layout.menuItems.dashboard',
```
therealemjy: I think text should remain being the actual content rather than a translation key:
```typescript
text: t('layout.menuItems.dashboard')
```

src/pages/Dashboard/Markets/SupplyMarket/CollateralConfirmModal/index.tsx
```diff
@@ -16,18 +17,23 @@ export const CollateralConfirmModal: React.FC<IConfirmCollateralModalProps> = ({
...
     <Modal
       className="venus-modal"
       isOpened={!!asset}
       handleClose={handleClose}
-      title={`${asset?.collateral ? 'Disable' : 'Enable'} as collateral`}
+      title={`${
+        asset?.collateral
+          ? t('markets.collateralConfirmModal.disable')
+          : t('markets.collateralConfirmModal.enable')
+      } ${t('markets.collateralConfirmModal.asCollateral')}`}
```
therealemjy: Once these are translated, the order of the words might not be the same depending on the language it is translated to. For that reason, we should include the full sentence for each string:
```typescript
title={asset?.collateral ? t('markets.collateralConfirmModal.disableAsCollateral') : t('markets.collateralConfirmModal.enableAsCollateral') }
```

src/pages/Dashboard/Markets/SupplyMarket/index.tsx
```diff
@@ -135,22 +142,24 @@ const SupplyMarket: React.FC = () => {
...
         setConfirmCollateral(asset);
         enterMarkets({ vtokenAddresses: [asset.vtokenAddress], account });
       } catch (error) {
         throw new ToastError(
-          'Collateral Enable Error',
-          `There was a problem enabeling collateral ${asset.name}. Please try again.`,
+          t('markets.errors.collateralEnableError.title'),
+          `${t('markets.errors.collateralEnableError.description')} ${asset.name}. ${t(
+            'markets.errors.tryAgain',
+          )}`,
```
therealemjy: Same thing here, the line needs to be one full sentence. You can use [interpolation](https://react.i18next.com/legacy-v9/interpolate) to insert `asset.name` within the sentence:
```typescript
t('markets.errors.collateralEnablingError', { assetName: asset.name })
```

Then in your JSON file the translation will be like that:
```json
"collateralEnablingError": "There was a problem enabling {{assetName}} as a collateral. Please try again."
```

src/pages/Dashboard/Markets/SupplyMarket/index.tsx
```diff
@@ -159,14 +168,16 @@ const SupplyMarket: React.FC = () => {
         exitMarkets({ vtokenAddress: asset.vtokenAddress, account });
       } catch (error) {
         throw new ToastError(
-          'Collateral Disable Error',
-          `There was a problem disabeling collateral ${asset.name}. Please try again.`,
+          t('markets.errors.collateralDisableError.title'),
+          `${t('markets.errors.collateralDisableError.description')} ${asset.name}. ${t(
+            'markets.errors.tryAgain',
+          )}`,
```
therealemjy: Same thing as previous comment about using interpolation.

src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
```diff
@@ -63,7 +65,7 @@ export const MintVaiUi: React.FC<IMintVaiUiProps> = ({
       // @TODO: display success modal instead of toast once it's been
       // implemented
       toast.success({
-        title: `You successfully minted ${convertWeiToCoins({
+        title: `${t('mintRepayVai.mintVai.successMessage')}${convertWeiToCoins({
```
therealemjy: Same thing here, let's use interpolation.

src/pages/Dashboard/MintRepayVai/RepayVai/index.tsx
```diff
@@ -54,7 +56,7 @@ export const RepayVaiUi: React.FC<IRepayVaiUiProps> = ({
       // @TODO: display success modal instead of toast once it's been
       // implemented
       toast.success({
-        title: `You successfully repaid ${convertWeiToCoins({
+        title: `${t('mintRepayVai.repayVai.successMessage')}${convertWeiToCoins({
```
therealemjy: Same thing here, let's use interpolation

src/components/v2/Layout/Header/index.tsx
```diff
@@ -1,27 +1,30 @@
...
-  );
+const Header = () => {
+  const location = useLocation();
   const styles = useStyles();
+  const { t } = useTranslation();
+
+  const title = useMemo(() => {
+    const currentItemId = menuItems.find(item => item.href === location.pathname)?.id;
+    return t(`layout.menuItems.${currentItemId}`);
+  }, [location.pathname]);
```
therealemjy: There's two issues with that unfortunately:
- it's risky, because there's no check done by the `t` function to verify the key passed exists in the translations
- automatically extracting the translation keys by running `yarn extract-translation-keys` will not catch these. Even worse, it will delete them from the translation file if you added them manually

The solution for this is to include the text directly inside each menu item.

src/components/v2/Layout/Sidebar/index.tsx
```diff
@@ -41,16 +45,15 @@ export const SidebarUi: React.FC<ISidebarProps> = ({ account }) => {
...
-              <ListItemButton key={text} component="li" css={styles.listItem} disableRipple>
+            {menuItems.map(({ href, icon, id }) => (
+              <ListItemButton key={id} component="li" css={styles.listItem} disableRipple>
                 <NavLink to={href} activeClassName="active-menu-item">
                   <ListItemIcon css={styles.listItemIcon}>
                     <Icon name={icon} />
                   </ListItemIcon>
                   <Typography variant="body2" css={styles.listItemText}>
-                    {text}
+                    {t(`layout.menuItems.${id}`)}
```
therealemjy: Same thing as previous comment about needing to include the text itself inside each menu item.

src/components/v2/Layout/Sidebar/index.tsx
```diff
@@ -112,7 +115,7 @@ export const SidebarUi: React.FC<ISidebarProps> = ({ account }) => {
                       component="span"
                       css={[styles.listItemText, styles.mobileListItemText]}
                     >
-                      {text}
+                      {t(`layout.menuItems.${id}`)}
```
therealemjy: Same thing as previous comment about needing to include the text itself inside each menu item.

src/components/v2/Layout/constants.tsx
```diff
@@ -2,45 +2,48 @@ import { isOnTestnet } from 'config';
...
   href: string;
-  text: string;
   icon: IconName;
 }
 
 export const menuItems: IMenuItem[] = [
   {
+    id: 'dashboard',
     href: '/dashboard',
-    text: 'Dashboard',
```
therealemjy, line 13: To elaborate on my previous comment, that's where you'd insert your text:
```typescript
text: t('layout.menuItems.dashboard'),
```

src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
```diff
@@ -141,7 +145,7 @@ const MintVai: React.FC = () => {
     if (!account) {
       // This error should never happen, since the form inside the UI component
       // is disabled if there's no logged in account
-      throw new Error('An internal error occurred: account undefined. Please try again later.');
+      throw new Error(t('mintRepayVai.mintVai.errorMessage'));
```
therealemjy: What do you think about making this error reusable? We use this exact phrasing inside the v2 components for all internal components, so we could have something like that:
```typescript
throw new Error(t('errors.internal', { errorMessage: t('mintRepayVai.mintVai.undefinedAccountErrorMessage') }));
```

The error would look like that in the translation file:
```json
"errors": {
  "internal": "An internal error occurred: {{errorMessage}}. Please try again later."
}
```

src/pages/Dashboard/MintRepayVai/RepayVai/index.tsx
```diff
@@ -127,7 +131,7 @@ const RepayVai: React.FC = () => {
     if (!account) {
       // This error should never happen, since the form inside the UI component
       // is disabled if there's no logged in account
-      throw new Error('An internal error occurred: account undefined. Please try again later.');
+      throw new Error(t('mintRepayVai.repayVai.errorMessage'));
```
therealemjy: Same as previous comment about perhaps using a generic error message

therealemjy (commented):
Looking at the translation file, I'm guessing you added each key manually.
You can use `yarn extract-translation-keys` to do that for you :)

src/components/v2/Layout/constants.tsx
baomastr: it doesn't work with such construction. we have no texts when trying to extract translation outside of component

src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
baomastr: yeah. nice suggestion

therealemjy (approved):

PR #294, opened 2022-04-12
Update connect wallet

coreyar:
Update ConnectWallet component so that the UI component handles the logic for rendering children.
Removes unnecessary wrapper  from story that was preventing it from rendering


therealemjy (approved):
Nicer 👍

PR #296, opened 2022-04-13
Refactor contract logic

therealemjy:
Don't be scared by the size of the PR, most of it is file path updates and moving files around.

The context of this PR is I started looking at integrating TypeChain to add types to our contracts and realized our logic for handling contracts is a bit all over the place. I found contract addresses in `.ts` files, others in `.json` files, some declared as constants etc.
For that reason, I first wanted to clean up our logic for handling contracts before I integrate TypeChain to the project.
Here's the full list of changes:
- create a `contract` client that works similarly to our `api` client (contracts can be retrieved using getter functions and a hooks) and is used to get an instance of a contract. Once I integrate TypeChain, this client will return strongly typed contract instances, making them safer to use
- create a `contracts` folder inside the `constants` folder to store all the contract ABIs and addresses inside `.json` files that follow the same format. This folder is also used to store the token objects
- create three utilities: `getContractAddress`, `getToken` and `getVBepToken` (self explanatory)
- create `generateBscScanUrl` util to generate all sorts of URL going to BSS scan, including URLs of tokens (for which it uses `getToken`


src/components/Dashboard/VaiInfo.tsx
```diff
@@ -70,7 +70,8 @@ function VaiInfo({ settings }: VaiInfoProps) {
   const { account } = useWeb3Account();
   const { userVaiBalance } = useVaiUser();
   const handleLink = () => {
-    window.open(`${BASE_BSC_SCAN_URL}/token/${getVaiTokenAddress()}?a=${account}`, '_blank');
+    const vaiTokenAddress = getToken('vai').address;
+    window.open(`${BASE_BSC_SCAN_URL}/token/${vaiTokenAddress}?a=${account}`, '_blank');
```
baomastr: same construction used second time. what about reusable urlBuilder function for these cases?

src/components/Dashboard/CoinInfo.tsx
```diff
@@ -59,10 +60,7 @@ function CoinInfo() {
   const { userXvsBalance } = useMarketsUser();
 
   const handleLink = () => {
-    window.open(
-      `${BASE_BSC_SCAN_URL}/token/${constants.CONTRACT_TOKEN_ADDRESS.xvs.address}?a=${account}`,
-      '_blank',
-    );
+    window.open(`${BASE_BSC_SCAN_URL}/token/${getToken('xvs').address}?a=${account}`, '_blank');
```
baomastr: same construction used second time. what about reusable urlBuilder function for these cases?

src/containers/Main/XVS/index.tsx
```diff
@@ -270,13 +271,13 @@ function XVS({ settings }: XVSProps) {
             <img src={coinImg} alt="xvs" />
             <a
               className="highlight"
-              href={`${BASE_BSC_SCAN_URL}/token/${constants.XVS_TOKEN}`}
+              href={`${BASE_BSC_SCAN_URL}/token/${getToken('xvs').address}`}
```
baomastr: same here. I'd make something like `buildBscUrl('xsv')`

src/utilities/common.ts
```diff
@@ -121,15 +119,15 @@ export const formatCommaThousandsPeriodDecimal = commaNumber.bindWith(',', '.');
 export const format = (bigNumber: BigNumber, dp = 2) =>
   formatCommaThousandsPeriodDecimal(bigNumber.dp(dp, 1).toString(10));
 
-export const getTokenDecimals = (tokenSymbol: keyof typeof CONTRACT_TOKEN_ADDRESS) =>
-  CONTRACT_TOKEN_ADDRESS[tokenSymbol]?.decimals || 18;
+export const getTokenDecimals = (tokenSymbol: keyof typeof TOKENS) =>
```
baomastr: you already have `TokenSymbol` type for `keyof typeof TOKENS`

src/utilities/common.ts
```diff
@@ -121,15 +119,15 @@ export const formatCommaThousandsPeriodDecimal = commaNumber.bindWith(',', '.');
...
+  TOKENS[tokenSymbol]?.decimals || 18;
 
 export const formatCoinsToReadableValue = ({
   value,
   tokenSymbol,
 }: {
   value: BigNumber;
-  tokenSymbol: keyof typeof CONTRACT_TOKEN_ADDRESS;
+  tokenSymbol: keyof typeof TOKENS;
```
baomastr: so you can just make autoreplace overall the project

baomastr (approved):

src/utilities/common.ts
therealemjy: That's a good point.
And it actually made me realize `TokenSymbol` is actually an ID in our case, and that the vBEP tokens should get their own `VBepTokenId` type too. I've made those changes.

src/components/Dashboard/VaiInfo.tsx
therealemjy: Good idea. I've made a `generateBscScanUrl` util that can be used to generate all kinds of BNC scan urls

src/clients/contracts/getters.ts
```diff
@@ -0,0 +1,88 @@
+import Web3 from 'web3';
```
coreyar, line 1: It would be nice to track this file as being moved in git

therealemjy: I'm not too sure why GitHub doesn't track is as having been moved on the final result to be honest, since it did track it when I moved it: https://github.com/VenusProtocol/venus-protocol-interface/pull/296/commits/8c59e25b45909b02bb20cfb346b665dbeee9ca23#diff-43ab2a040e52b1a9646b6078715f89921951c1fbebd15d7ffa7b136d34f7b20a

src/components/Vault/BasicVault/CardContent.tsx
```diff
@@ -5,7 +5,7 @@ import BigNumber from 'bignumber.js';
 import { useWeb3Account } from 'clients/web3';
 import NumberFormat from 'react-number-format';
 import { commaFormat } from 'utilities/common';
-import { TokenSymbol } from 'types';
+import { TokenId } from 'types';
```
coreyar, line 8: nice. I like the change because it clarifies what property we are using. Per the comment in Slack I went with symbol for what we display. We could add name back if we every display the long name of tokens but we currently don't

coreyar (approved):

PR #297, opened 2022-04-14
Fix small design issues on MyAccount

therealemjy:
- fix spacings of `MyAccount` component 
- display `ProgressBarHorizontal` as a block

### DIFF
--- a/src/components/v2/ProgressBarHorizontal/styles.ts
+++ b/src/components/v2/ProgressBarHorizontal/styles.ts
@@ -9,7 +9,7 @@ export const useStyles = ({ over }: { over: boolean }) => {
       color: ${over ? theme.palette.interactive.error50 : theme.palette.interactive.success};
       background-color: ${theme.palette.background.default};
       height: ${theme.spacing(2)};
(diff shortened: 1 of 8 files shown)


coreyar (approved):

src/components/v2/ProgressBarHorizontal/styles.ts
```diff
@@ -5,10 +5,11 @@ export const useStyles = ({ over }: { over: boolean }) => {
   const theme = useTheme();
   return {
     slider: css`
+      display: block;
       color: ${over ? theme.palette.interactive.error50 : theme.palette.interactive.success};
       background-color: ${theme.palette.background.default};
       height: ${theme.spacing(2)};
-      padding: 0;
+      padding: 0 !important;
```
baomastr, line 12: Is there any way to avoid using important?

therealemjy: Nope, MUI's CSS is horrible and uses a weird rule based on the cursor --'

baomastr (approved):

PR #299, opened 2022-04-14
Wire up MyAccount component

therealemjy:
- update `trackTooltip` and `markTooltip` props of the `ProgressBarHorizontal` component so that they reflect what MUI's `Tooltip` component accepts, which is the component used under the hood
- add a `SAFE_BORROW_LIMIT_PERCENTAGE` constant to the config
- add option to remove decimals from value returned by `formatCentsToReadableValue`
- decouple `MyAccount` component into a UI and a container components
- wire up `MyAccount`


src/config/index.ts
```diff
@@ -65,3 +65,5 @@ export const VENUS_GITHUB_URL = 'https://github.com/VenusProtocol/';
 
 // TODO: update
 export const VENUS_TERMS_OF_SERVICE_URL = 'https://www.swipe.io/terms';
+
+export const SAFE_BORROW_LIMIT_PERCENTAGE = 80;
```
coreyar, line 69: nice

src/components/v2/Tooltip/Tooltip.tsx
```diff
@@ -3,7 +3,7 @@ import { Global } from '@emotion/react';
 import MuiTooltip, { TooltipProps } from '@mui/material/Tooltip';
 import { useStyles } from './styles';
 
-type ITooltipProps = TooltipProps;
+export type ITooltipProps = TooltipProps;
```
coreyar: shouldn't be an prefaced with an `I` as it is not an interface. There is an eslint plugin for this. 
I also updated this to take an element so `<Trans>` can be used here

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -0,0 +1,217 @@
...
+  className,
+}: IMyAccountUiProps) => {
+  const styles = useStyles();
+  const { t, Trans } = useTranslation();
+
+  const handleXvsToggleChange: IToggleProps['onChange'] = (_event, checked) => onXvsToggle(checked);
+
+  const readableBorrowBalance =
+    typeof borrowBalanceCents === 'number'
+      ? formatCentsToReadableValue(borrowBalanceCents)
```
coreyar: Since these functions produce a readable value, I'd include the undefined check inside the function and return the `'-'` as the readable value if `undefined` or unparseable

src/components/v2/Tooltip/Tooltip.tsx
therealemjy: Good point about the interface.

About the tooltip, we should pass the full text directly otherwise the parser can't extract the translation keys.

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
therealemjy: I think my naming is wrong here, I see why it's confusing. We need it to be undefined if `borrowBalanceCents` is not a number so that we can reuse it for the `trackTooltip` prop of `ProgressBarHorizontal`, where we only pass some content if it isn't `undefined`.

I could use the prefix `formatted` rather than `readable`?

coreyar: You could also use the a constant for the empty value and refuse the tooltip if it is the empty content. Alternatively, we could skip the empty formatted value here. I'll comment other places where the default formatted logic is also duplicated.

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -0,0 +1,217 @@
...
+            {t('myAccount.netApy')}
+          </Typography>
+
+          <Tooltip css={styles.tooltip} title={t('myAccount.netApyTooltip')}>
+            <Icon css={styles.infoIcon} name="info" />
+          </Tooltip>
+        </div>
+
+        <Typography variant="h1" color="interactive.success">
+          {typeof netApyPercentage === 'number' ? `${netApyPercentage}%` : '-'}
```
coreyar: Should this be formatted from a function?

coreyar: This is the same duplicated logic where we check if a value is a number, if it is we do something, if it isn't we return a `-`

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -0,0 +1,217 @@
...
+        </Typography>
+      </div>
+
+      <ul css={styles.list}>
+        <Typography component="li" variant="h4" css={styles.item}>
+          <Typography component="div" variant="small2" css={styles.labelListItem}>
+            {t('myAccount.dailyEarnings')}
+          </Typography>
+
+          {typeof dailyEarningsCents === 'number'
```
coreyar: This is the same duplicated logic where we check if a value is a number, if it is we do something, if it isn't we return a `-`

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -0,0 +1,217 @@
...
+            ? formatCentsToReadableValue(dailyEarningsCents)
+            : '-'}
+        </Typography>
+
+        <Typography component="li" variant="h4" css={styles.item}>
+          <Typography component="div" variant="small2" css={styles.labelListItem}>
+            {t('myAccount.supplyBalance')}
+          </Typography>
+
+          {typeof supplyBalanceCents === 'number'
```
coreyar: This is the same duplicated logic where we check if a value is a number, if it is we do something, if it isn't we return a `-`

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -0,0 +1,217 @@
...
+            ? formatCentsToReadableValue(supplyBalanceCents)
+            : '-'}
+        </Typography>
+
+        <Typography component="li" variant="h4" css={styles.item}>
+          <Typography component="div" variant="small2" css={styles.labelListItem}>
+            {t('myAccount.borrowBalance')}
+          </Typography>
+
+          {readableBorrowBalance || '-'}
```
coreyar: This is the same duplicated logic where we check if a value is a number (in this case truthy), if it is we do something, if it isn't we return a `-`

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -0,0 +1,217 @@
...
+      </ul>
+
+      <div css={[styles.row, styles.topProgressBarLegend]}>
+        <div css={styles.inlineContainer}>
+          <Typography component="span" variant="small2" css={styles.inlineLabel}>
+            {t('myAccount.borrowLimitUsed')}
+          </Typography>
+
+          <Typography component="span" variant="small1" color="text.primary">
+            {readableBorrowLimitUsedPercentage || '-'}
```
coreyar: This is the same duplicated logic where we check if a value is a number (in this case truthy), if it is we do something, if it isn't we return a `-`

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -0,0 +1,217 @@
...
+        </div>
+
+        <div css={styles.inlineContainer}>
+          <Typography component="span" variant="small2" css={styles.inlineLabel}>
+            {t('myAccount.limit')}
+          </Typography>
+
+          <Typography component="span" variant="small1" color="text.primary">
+            {typeof borrowLimitCents === 'number'
+              ? formatCentsToReadableValue(borrowLimitCents, true)
```
coreyar: This is the same duplicated logic where we check if a value is a number, if it is we do something, if it isn't we return a `-`

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+      <Typography component="div" variant="small2" css={styles.bottom}>
+        <Icon name="shield" css={styles.shieldIcon} />
+
+        <Typography component="span" variant="small2" css={styles.inlineLabel}>
+          {t('myAccount.safeLimit')}
+        </Typography>
+
+        <Typography component="span" variant="small1" color="text.primary" css={styles.safeLimit}>
+          {readableSafeBorrowLimit || '-'}
```
coreyar: This is the same duplicated logic where we check if a value is a number (in this case truthy), if it is we do something, if it isn't we return a `-`

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
coreyar: A better option would be to wait to format until you've decided to render the tooltip or use the original value to decided to render the tooltip `borrowBalanceCents`

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -0,0 +1,217 @@
...
+        </div>
+
+        <Typography variant="h1" color="interactive.success">
+          {typeof netApyPercentage === 'number' ? `${netApyPercentage}%` : '-'}
+        </Typography>
+      </div>
+
+      <ul css={styles.list}>
+        <Typography component="li" variant="h4" css={styles.item}>
+          <Typography component="div" variant="small2" css={styles.labelListItem}>
```
coreyar: Could we use a more descriptive tag here to render the text? I typically think of using span for all nested text?

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -0,0 +1,217 @@
...
+            {t('myAccount.dailyEarnings')}
+          </Typography>
+
+          {typeof dailyEarningsCents === 'number'
+            ? formatCentsToReadableValue(dailyEarningsCents)
+            : '-'}
+        </Typography>
+
+        <Typography component="li" variant="h4" css={styles.item}>
+          <Typography component="div" variant="small2" css={styles.labelListItem}>
```
coreyar: Same here

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -0,0 +1,217 @@
...
+            {t('myAccount.supplyBalance')}
+          </Typography>
+
+          {typeof supplyBalanceCents === 'number'
+            ? formatCentsToReadableValue(supplyBalanceCents)
+            : '-'}
+        </Typography>
+
+        <Typography component="li" variant="h4" css={styles.item}>
+          <Typography component="div" variant="small2" css={styles.labelListItem}>
```
coreyar: and here

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -0,0 +1,217 @@
...
+          </Typography>
+        </div>
+      </div>
+
+      <ProgressBarHorizontal
+        css={styles.progressBar}
+        value={borrowLimitUsedPercentage || 0}
+        mark={safeBorrowLimitPercentage}
+        step={1}
+        ariaLabel="Borrow limit"
```
coreyar: This needs to be translated

src/components/v2/Tooltip/Tooltip.tsx
```diff
@@ -3,9 +3,9 @@ import { Global } from '@emotion/react';
 import MuiTooltip, { TooltipProps } from '@mui/material/Tooltip';
 import { useStyles } from './styles';
 
-type ITooltipProps = TooltipProps;
+export { type TooltipProps } from '@mui/material/Tooltip';
 
-export const Tooltip = ({ children, ...rest }: ITooltipProps) => {
+export const Tooltip = ({ children, ...rest }: TooltipProps) => {
```
coreyar: I think we should consistently type functional components with `React.FC`

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
therealemjy: I see what you mean, I like that idea.
I'll add the constant, but also create a util that can be reused throughout the app to render values that can be `undefined`.

About the tooltip, we use this formatted value both in the component itself and the tooltip. We only render the tooltip if the `borrowBalanceCents` isn't `undefined`, hence why `readableBorrowBalance` needs to be casted as `undefined` if that is the case. I know it's confusing, I think the util will help streamline all that.

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
therealemjy: Works for me 👍
I picked this up so didn't really check the DOM to be honest.

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
therealemjy: Yep, I deleted it when I rebased :(

src/components/v2/Tooltip/Tooltip.tsx
therealemjy: Yep definitely, didn't realize that.

src/pages/Dashboard/MyAccount/MyAccountUi.tsx
therealemjy: I made some updates that should clear things up.
I've updated the `formatCentsToReadableValue` so that it accepts an `undefined` value and returns a placeholder (defined as a constant; `PLACEHOLDER_KEY`) in that case. I figured this will can be used a lot throughout the codebase. I also updated `formatToReadablePercentage` to follow the same behavior.

This meant I could strip out a lot of duplicated logic from my PR and use these utils instead.
Let me know if that's clearer now.

coreyar (approved):

PR #300, opened 2022-04-14
Update GitHub actions config to run checks on PRs only

therealemjy:
As Corey proposed a few days ago, this should help decluttering our pipelines.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,6 @@
 name: Tests and Checks
 
-on: [push]
+on: [pull_request]
 
 env:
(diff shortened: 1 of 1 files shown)


coreyar (approved):

coreyar:
I wonder how we got into this state 
```
[reg-keygen-git-hash-plugin] error Can't detect branch name because HEAD is on detached commit node.
[reg-suit] warn Failed to detect the previous snapshot key
```

therealemjy:
That's a weird one I'll look into it.
I wonder if changing the trigger for GitHub actions reset the snapshots somehow and so now it doesn't find any previous snapshot. I'll merge and see if it continues to be a problem.

PR #301, opened 2022-04-14
Remove name in favor of symbol

coreyar:
To clarify the properties on the asset, this PR removes name in favor of using Id to identify the asset and symbol to display its ticket letters.

If we ever need to display the full name of an asset, we can add back name and use it to display the full name


therealemjy (approved):
Nice 👍

PR #303, opened 2022-04-15
Add luna and tusd coin icons

baomastr:


therealemjy (approved):
An SVG that's 797 line! 😳

PR #306, opened 2022-04-15
Type contracts

therealemjy:
- add support for TypeChain
- apply generated types on contract getters and hooks
- fix type errors in the codebase following this change


src/clients/contracts/getters.ts
```diff
@@ -10,79 +12,161 @@ import vaiUnitrollerAbi from 'constants/contracts/abis/vaiUnitroller.json';
...
+    return getContract(xvsTokenAbi as AbiItem[], tokenAddress, web3) as unknown as TokenContract<T>;
   }
 
-  return getContract(abi, getToken(name).address, web3);
+  if (tokenId === 'vai') {
+    return getContract(vaiTokenAbi as AbiItem[], tokenAddress, web3) as unknown as TokenContract<T>;
+  }
+
+  if (tokenId === 'vrt') {
+    return getContract(vrtTokenAbi as AbiItem[], tokenAddress, web3) as unknown as TokenContract<T>;
```
coreyar: I think we could improve this by making `getContract` generic

therealemjy: It is generic: you can get any contract with it (which is what we do in all getters in this file)

coreyar: It does not have a generic type. This is what I was thinking 
```
const getContract = <T>(abi: AbiItem | AbiItem[], address: string, web3Instance: Web3) => {
  const web3 = web3Instance ?? getWeb3NoAccount();
  return new web3.eth.Contract(abi, address) as unknown as T;
};

export const getTokenContract = <T extends TokenId>(tokenId: T, web3: Web3): TokenContract<T> => {
  const tokenAddress = getToken(tokenId).address;

  if (tokenId === 'xvs') {
    return getContract<TokenContract<T>>(xvsTokenAbi as AbiItem[], tokenAddress, web3);
  }

  if (tokenId === 'vai') {
    return getContract<TokenContract<T>>(vaiTokenAbi as AbiItem[], tokenAddress, web3);
  }

  if (tokenId === 'vrt') {
    return getContract<TokenContract<T>>(vrtTokenAbi as AbiItem[], tokenAddress, web3);
  }
  ```

therealemjy: Oh I see, I didn't understand that. Works for me!

coreyar (approved):

PR #307, opened 2022-04-15
Add Supply Modal presentational component

coreyar:
Adds the UI for the supply modal with supply and withdraw tabs.

- Creates InfoRow component to display horizontal layout of information
- Adds pointer styles to TableRow
- Update AmountForm to have custom initialValues
- Update Value component can only show a single value


src/components/v2/EnableToken/index.tsx
```diff
@@ -1,20 +1,20 @@
...
 export interface IEnableTokenProps {
   symbol: TokenId;
   isEnabled: boolean;
-  title: string;
-  tokenInfo: { symbol: TokenId; text: string; apy: BigNumber | number }[];
+  title: string | React.ReactElement;
+  tokenInfo: IInfoRowProps['info'];
   approveToken: () => void;
   vtokenAddress: string;
   disabled: boolean;
```
therealemjy: I think this prop should be optional, to have the same as with buttons for example.

src/components/v2/InfoRow/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+          <Typography variant="body1" component="span" css={styles.right}>
+            {right}
+          </Typography>
+        </div>
+      ))}
+    </div>
+  );
+};
+
+export default InfoRow;
```
therealemjy: We have a component for that already: `LabeledInlineContent`. Feel free to rename it if you'd prefer another name

src/components/v2/ProgressBarHorizontal/index.tsx
```diff
@@ -14,10 +14,11 @@ export interface ISliderProps {
   ariaLabel: string;
   min: number;
   max: number;
-  trackTooltip?: string;
-  markTooltip?: string;
+  trackTooltip?: string | React.ReactElement;
+  markTooltip?: string | React.ReactElement;
```
therealemjy: We could extend the props from the tooltip component:
```suggestion
  trackTooltip?: ITooltipProps['title'];
  markTooltip?: ITooltipProps['title'];
```

src/components/v2/ValueUpdate/index.tsx
```diff
@@ -20,9 +20,17 @@ export const ValueUpdate: React.FC<IValueUpdateProps> = ({
...
-
+  if (typeof update !== 'number' && !(update instanceof BigNumber)) {
+    return (
+      <div className={className} css={styles.container}>
+        <Typography component="span" variant="body1">
+          {format(original)}
+        </Typography>
+      </div>
+    );
+  }
```
therealemjy: You can avoid having to repeat the logic and have just one return statement:

```jsx
return (
  <div className={className} css={styles.container}>
    <Typography component="span" variant="body1">
      {format(original)}
    </Typography>

    {(typeof update === 'number' || update instanceof BigNumber) && (
      <Icon name="arrowShaft" css={styles.icon} />

      <Typography component="span" variant="body1">
        {format(update)}
      </Typography>
    )}
  </div>
)
```

src/context/AuthContext.tsx
```diff
@@ -6,8 +6,7 @@ import { Connector, useAuth } from 'clients/web3';
 import toast from 'components/Basic/Toast';
 import { AuthModal } from 'components/v2/AuthModal';
 
-// eslint-disable-next-line no-spaced-func
-export const AuthContext = React.createContext<{
+export interface IAuthContextValue {
```
therealemjy, line 9: Nice

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const { id: assetId } = asset;
+  const { amount } = values;
+  const validAmount = amount && !amount.isZero() && !amount.isNaN();
+  const styles = useStyles();
+  const { t, Trans } = useTranslation();
+  const [newBorrowLimit, setNewBorrowLimit] = useState<BigNumber | undefined>(undefined);
+  const [newBorrowPercent, setNewBorrowPercent] = useState(
+    userTotalBorrowBalance.div(userTotalBorrowLimit).times(100),
+  );
+  useEffect(() => resetForm, []);
```
therealemjy: Why do you need that?

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+        .times(tokenPrice)
+        .times(collateralFactor);
+      const temp = newBalance(amountInUsd);
+      setNewBorrowLimit(BigNumber.maximum(temp, 0));
+      setNewBorrowPercent(userTotalBorrowBalance.div(temp).times(100));
+    } else if (BigNumber.isBigNumber(userTotalBorrowLimit)) {
+      setNewBorrowLimit(undefined);
+      setNewBorrowPercent(userTotalBorrowBalance.div(userTotalBorrowLimit).times(100));
+    }
+  }, [amount, asset?.id, userTotalBorrowBalance, userTotalBorrowLimit]);
```
therealemjy: Instead of using a state and an effect, you can use `useMemo` to store `newBorrowLimit` and `newBorrowPercent`.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+  // Convert wallet/ supply balance into wei
+  const maxInputWei = React.useMemo(
+    () => convertCoinsToWei({ value: maxInput, tokenSymbol: assetId }),
+    [maxInput.toString()],
```
therealemjy:
```suggestion
    [maxInput.toString(), assetId],
```

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+        rightMaxButtonLabel={t('supplyWithdraw.max').toUpperCase()}
+        css={styles.input}
+      />
+      <Typography component="span" variant="small1" css={styles.greyLabel}>
+        <Trans i18nKey="supplyWithdraw.walletBalance">
+          <Typography component="span" variant="small1" css={styles.whiteLabel}>
+            {{ amount: asset.walletBalance, symbol: assetId?.toUpperCase() }}
+          </Typography>
+        </Trans>
+      </Typography>
```
therealemjy: The second `Typography` tag shouldn't exist here (you're already defining the typography style with the tag that encapsulate the whole thing).
Also, you need to include the full sentence here, so that when you run `yarn extract-translation-keys` it picks it up and out it in the doc automatically. Otherwise it looks like we're trying to render an object inside the DOM, which is not allowed in React.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+        <div css={styles.totalAndLimit}>
+          <Tooltip
+            placement="bottom"
+            title={
+              <Trans i18nKey="supplyWithdraw.currentBorrowBalance">
+                {{
+                  amount: currencyFormatter(userTotalBorrowBalance),
+                  percent: newBorrowPercent.toFixed(0),
+                }}
+              </Trans>
```
therealemjy: No need to use the trans component here, you're only rendering text

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+            }
+          >
+            <Typography component="span" variant="body1" css={styles.greyLabel}>
+              <Trans i18nKey="supplyWithdraw.currentAmount">
+                <Typography component="span" variant="body1" css={styles.whiteLabel}>
+                  {{ amount: currencyFormatter(userTotalBorrowBalance) }}
+                </Typography>
```
therealemjy: Same thing here: `t`.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+              </Trans>
+            </Typography>
+          </Tooltip>
+          <Tooltip
+            placement="bottom"
+            title={
+              <Trans i18nKey="supplyWithdraw.maxBorrowTooltip">
+                {{ amount: currencyFormatter(userTotalBorrowLimit) }}
+              </Trans>
```
therealemjy: Same thing here: `t`.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+          </Tooltip>
+        </div>
+        <ProgressBarHorizontal
+          value={newBorrowPercent.toNumber()}
+          mark={80}
+          step={1}
+          min={0}
+          max={100}
+          ariaLabel="borrow percentage meter"
+          markTooltip={<Trans i18nKey="supplyWithdraw.safeBorrowLimit">{{ amount }}</Trans>}
```
therealemjy: Same thing here: t.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+            {
+              left: t('supplyWithdraw.dailyEarnings'),
+              right: formatCentsToReadableValue(dailyEarnings),
+            },
+            {
+              left: t('supplyWithdraw.supplyBalance'),
+              right: (
+                <Trans i18nKey="supplyWithdraw.supplyBalanceValue">
+                  {{ amount: format(asset.supplyBalance), symbol: asset.symbol }}
+                </Trans>
```
therealemjy: Same thing here: `t`.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+interface ISupplyWithdrawFormProps extends FormikProps<FormValues> {
+  asset: Asset;
+  tokenInfo: IInfoRowProps['info'];
+  userTotalBorrowBalance: BigNumber;
+  userTotalBorrowLimit: BigNumber;
+  dailyEarnings: BigNumber;
+  onSubmit: IAmountFormProps['onSubmit'];
+  enabledButtonKey: string;
+  disabledButtonKey: string;
```
therealemjy, line 38: We should pass the full text instead of the translation keys.
I'll explain in another comment why.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+                </Trans>
+              ),
+            },
+          ] as IInfoRowProps['info']
+        }
+        css={[styles.infoRow, styles.bottomInfo]}
+        className="info-row"
+      />
+      <PrimaryButton fullWidth disabled={!validAmount}>
+        {validAmount ? t(enabledButtonKey) : t(disabledButtonKey)}
```
therealemjy: Doing that means the parser won't be able to pick up the translation keys from the code automatically. We should instead pass the texts to the component directly.

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+const TAB_TEXT_KEYS = [
+  {
+    connect: 'supplyWithdraw.connectWalletToSupply',
+    enable: 'supplyWithdraw.enableToSupply',
+  },
+  {
+    connect: 'supplyWithdraw.connectWalletToWithdraw',
+    enable: 'supplyWithdraw.enableToWithdraw',
+  },
+];
```
therealemjy: These will have to be defined inside the component, otherwise as it is the parser can't pick up the translation keys automatically (it only detects keys that are wrapped with `t`: `t('supplyWithdraw.connectWalletToSupply')`).

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+                  {{ amount: format(asset.supplyBalance), symbol: asset.symbol }}
+                </Trans>
+              ),
+            },
+          ] as IInfoRowProps['info']
+        }
+        css={[styles.infoRow, styles.bottomInfo]}
+        className="info-row"
+      />
+      <PrimaryButton fullWidth disabled={!validAmount}>
```
therealemjy:
```suggestion
      <PrimaryButton fullWidth disabled={!validAmount} type="submit">
```

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+              </Trans>
+            }
+          >
+            <Typography component="span" variant="body1" css={styles.greyLabel}>
+              <Trans i18nKey="supplyWithdraw.maxAmount">
+                <Typography component="span" variant="body1" css={styles.whiteLabel}>
+                  {{ amount: currencyFormatter(userTotalBorrowLimit) }}
+                </Typography>
+              </Trans>
+            </Typography>
```
therealemjy: Same thing here: we should remove the nested `Typography` tag.

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+                        enabledButtonKey="supplyWithdraw.withdraw"
+                        disabledButtonKey="supplyWithdraw.enterValidAmountWithdraw"
+                        maxInput={asset.supplyBalance}
+                        newBalance={newBorrowAmount}
+                        {...formikBag}
+                      />,
+                    ][activeTabIndex]
+                  }
+                </AmountForm>
+              </EnableToken>
```
therealemjy: I think these should be two separate forms, likely two separate components. They don't share the same state.

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+          right: formatApy(asset.xvsSupplyApy),
+        },
+      ]
+    : [];
+
+  const onSubmit: IAmountFormProps['onSubmit'] = () => {
+    // TODO: https://app.clickup.com/t/24quhp4
+  };
+  const newSupplyAmount = (amount: BigNumber) => userTotalBorrowLimit.plus(amount);
+  const newBorrowAmount = (amount: BigNumber) => userTotalBorrowLimit.minus(amount);
```
therealemjy: The names are a bit confusing, I would expect those to be the actual values rather than functions.
Also, since you're passing `userTotalBorrowLimit` to `SupplyWithdrawForm` anyway, we don't need these functions as `userTotalBorrowLimit` already contains those methods.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+} from 'utilities/common';
+import { useStyles } from '../styles';
+
+interface ISupplyWithdrawFormProps extends FormikProps<FormValues> {
+  asset: Asset;
+  tokenInfo: IInfoRowProps['info'];
+  userTotalBorrowBalance: BigNumber;
+  userTotalBorrowLimit: BigNumber;
+  dailyEarnings: BigNumber;
+  onSubmit: IAmountFormProps['onSubmit'];
```
therealemjy: You don't need to pass the `onSubmit` callback, just to assign `type="submit"` to the submit button.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+                <Typography component="span" variant="body1" css={styles.whiteLabel}>
+                  {{ amount: currencyFormatter(userTotalBorrowLimit) }}
+                </Typography>
+              </Trans>
+            </Typography>
+          </Tooltip>
+        </div>
+        <ProgressBarHorizontal
+          value={newBorrowPercent.toNumber()}
+          mark={80}
```
therealemjy, line 160: Can you add a `TODO` to indicate we should use the constant `SAFE_BORROW_LIMIT_PERCENTAGE`? I know the PR that contains the change isn't merged in yet.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+          step={1}
+          min={0}
+          max={100}
+          ariaLabel="borrow percentage meter"
```
therealemjy: We need to render this using the `t` function.

therealemjy:
Some design issues (spacings, info rows should take all the horizontal space, submit button uses the wrong variant):
Screen Shot 2022-04-15 at 18 17 29

Figma:
Screen Shot 2022-04-15 at 18 16 44

Also, I'm guessing you added the translations to the `en.json` file manually. Check out the `README`, I've added some details on how the internationalization works. Basically, we need to run `yarn extract-translation-keys` on the project to add them to the doc.

src/components/v2/EnableToken/index.tsx
coreyar: I don't mind changing this but I think it is 100% acceptable as written. It would be easy to make this optional when we have an implementation that doesn't use it which we currently don't.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
coreyar: Because we want to clear the form when we switch tabs. The form isn't re-rendered only its contents. This is because the form value is needed in the render function which is why I broke out the contents into its own component. We could remove this by creating another component that combines the form contents and the AmountForm wrapper

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
coreyar: `useMemo` is not a replacement for `useState`. `useMemo` is a performance optimization while `useState` is to guarantee updates. I believe `useState` is the correct choice here and `useMemo` could be used to memoize values being set in state so they aren't calculated on every render.

```
You may rely on useMemo as a performance optimization, not as a semantic guarantee.
In the future, React may choose to “forget” some previously memoized values and recalculate
them on next render, e.g. to free memory for offscreen components. 
Write your code so that it still works without useMemo — and then add it to optimize performance.
```
[source](https://reactjs.org/docs/hooks-reference.html#usememo)

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
coreyar: I chose to use the Typography tag because I think that is preferential to using styling text outside of our theme. Two tags are used to style a single translation with different styles.

These are picked up when I run the command `yarn extract-translation-keys`.

I found this syntax for notating tags in translations which is why I went down this route.
[Source](https://react.i18next.com/latest/trans-component#usage-with-simple-html-elements-like-less-than-br-greater-than-and-others-v10.4.0)

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
coreyar: I believe this is correct because it is a single translation with two different styles. I don't see how that is possible with `t()`

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
coreyar: I am nesting the typography tag to add different styles. I chose to use typography so that we can leverage the theme as opposed to using dom tags directly. 

Per a previous discussion I am using Typography exclusively for rendering text.

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
coreyar: that works for me

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
coreyar: I was not but You are right about needing the type. I have added the type

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
coreyar: I can rename them so they are clearer. This are required because depending on which form is rendered we will add or subtract the user input amount from `userTotalBorrowLimit`

coreyar:
Pavel clarified the button variant. He changed it while you were reviewing. I'm not sure what happened with the spacings. They were correct when the component was merged. Some look like they could have been thrown off when the bace spacing value changed. They have been corrected.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
therealemjy: Ah right my bad, I thought this was the prop types of the UI component.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,202 @@
...
+  onSubmit: IAmountFormProps['onSubmit'];
+  enabledButtonKey: string;
+  disabledButtonKey: string;
+  maxInput: BigNumber;
+  calculateNewBalance: (amount: BigNumber) => BigNumber;
+}
+
+export const SupplyWithdrawContent: React.FC<
+  Omit<ISupplyWithdrawFormProps, 'onSubmit'> & FormikProps<FormValues>
+> = ({
```
therealemjy: So realizing that the prop types at the top of the file are the container's types, I think we should take this case the other way around. The container wraps the UI, so technically it should be the container's type that extends the UI. What do you think?

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,202 @@
...
+        .times(tokenPrice)
+        .times(collateralFactor);
+      const temp = calculateNewBalance(amountInUsd);
+      setNewBorrowLimit(BigNumber.maximum(temp, 0));
+      setNewBorrowPercent(userTotalBorrowBalance.div(temp).times(100));
+    } else if (BigNumber.isBigNumber(userTotalBorrowLimit)) {
+      setNewBorrowLimit(undefined);
+      setNewBorrowPercent(userTotalBorrowBalance.div(userTotalBorrowLimit).times(100));
+    }
+  }, [amount, asset?.id, userTotalBorrowBalance, userTotalBorrowLimit]);
```
therealemjy: I can't respond to your previous comment about using `useMemo` for optimizations only. I agree with that, however my point was that this code doesn't need a state to be written (so it could also be written without `useMemo`):
```typescript
let uncappedBorrowLimit = userTotalBorrowLimit;
let borrowLimit: BigNumber | undefined;

if (tokenPrice && validAmount) {
  const tokenPrice = getBigNumber(asset?.tokenPrice);
  const collateralFactor = getBigNumber(asset?.collateralFactor);

  const amountInUsd = convertWeiToCoins({
    value: amount as BigNumber,
    tokenSymbol: asset.id,
  })
    .times(tokenPrice)
    .times(collateralFactor);

  uncappedBorrowLimit = calculateNewBalance(amountInUsd);
  borrowLimit = BigNumber.maximum(uncappedBorrowLimit, 0);
}

const borrowPercent = userTotalBorrowBalance.div(uncappedBorrowLimit).times(100).toNumber();
```

Because you're not updating `newBorrowLimit` or `newBorrowLimit` from within the component expecting it to react to the change, you don't need to store these values in states.
The `else if` statement isn't needed in your logic, you're already initializing these variables with these values.

Now using `useMemo` means the calculation of these two variables wouldn't happen on every render, thus optimizing the process a bit:

```typescript
const [newBorrowLimit, newBorrowPercent] = React.useMemo(() => {
  let uncappedBorrowLimit = userTotalBorrowLimit;
  let borrowLimit: BigNumber | undefined;

  if (tokenPrice && validAmount) {
    const tokenPrice = getBigNumber(asset?.tokenPrice);
    const collateralFactor = getBigNumber(asset?.collateralFactor);

    const amountInUsd = convertWeiToCoins({
      value: amount as BigNumber,
      tokenSymbol: asset.id,
    })
      .times(tokenPrice)
      .times(collateralFactor);

    uncappedBorrowLimit = calculateNewBalance(amountInUsd);
    borrowLimit = BigNumber.maximum(uncappedBorrowLimit, 0);
  }

  const borrowPercent = userTotalBorrowBalance.div(uncappedBorrowLimit).times(100).toNumber();

  return [borrowLimit, borrowPercent];
}, [asset?.tokenPrice, asset?.collateralFactor, userTotalBorrowBalance, userTotalBorrowLimit]);
```

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -0,0 +1,202 @@
...
+        css={styles.input}
+      />
+      <Typography component="span" variant="small1" css={styles.greyLabel}>
+        <Trans i18nKey="supplyWithdraw.walletBalance">
+          <Typography component="span" variant="small1" css={styles.whiteLabel}>
+            {{ amount: asset.walletBalance, symbol: assetId?.toUpperCase() }}
+            {asset.walletBalance} {assetId?.toUpperCase()}
+          </Typography>
+        </Trans>
+      </Typography>
```
therealemjy: Same thing here the comment disappeared.

You need to add the actual text inside the `Trans` tag:
```jsx
<Typography component="span" variant="small1" css={styles.greyLabel}>
  <Trans i18nKey="supplyWithdraw.walletBalance">
    Wallet balance:{' '}
    <Typography component="span" variant="small1" css={styles.whiteLabel}>
      {amount} {symbol}
    </Typography>
  </Trans>
</Typography>;
```

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
therealemjy: Yeah my bad, it contains HTML so it needs to use `t`. I got confused from seeing the object inside the tag.
We can't render objects like that though, we have to render the whole sentence inside it.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
therealemjy: Got you.

Same thing here though: the whole sentence needs to be rendered (not just an object).

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+    >
+      <>
+        <Tabs
+          tabTitles={[t('supplyWithdraw.supply'), t('supplyWithdraw.withdraw')]}
+          onTabChange={setActiveTabIndex}
+          fullWidth
+        />
+        <div className={className} css={styles.container}>
+          <ConnectWallet message={activeTabTextKeys.connect}>
+            {asset ? (
```
therealemjy:
```suggestion
            {asset && (
```

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+                      userTotalBorrowBalance={userTotalBorrowBalance}
+                      userTotalBorrowLimit={userTotalBorrowLimit}
+                      dailyEarnings={dailyEarnings}
+                      onSubmit={onSubmit}
+                      enabledButtonKey={t('supplyWithdraw.withdraw')}
+                      disabledButtonKey={t('supplyWithdraw.enterValidAmountWithdraw')}
+                      maxInput={asset.supplyBalance}
+                      calculateNewBalance={calculateNewBorrowAmount}
+                    />,
+                  ][activeTabIndex]
```
therealemjy: It took me a minute to understand what the nested array was for ^^
How about just using a ternary operator?
```jsx
{activeTabIndex == 0 ? (
  <SupplyWithdrawForm
    key="supply"
    asset={asset}
    tokenInfo={tokenInfo}
    userTotalBorrowBalance={userTotalBorrowBalance}
    userTotalBorrowLimit={userTotalBorrowLimit}
    dailyEarnings={dailyEarnings}
    onSubmit={onSubmit}
    enabledButtonKey={t('supplyWithdraw.supply')}
    disabledButtonKey={t('supplyWithdraw.enterValidAmountSupply')}
    maxInput={asset.walletBalance}
    calculateNewBalance={calculateNewSupplyAmount}
  />
) : (
  <SupplyWithdrawForm
    key="withdraw"
    asset={asset}
    tokenInfo={tokenInfo}
    userTotalBorrowBalance={userTotalBorrowBalance}
    userTotalBorrowLimit={userTotalBorrowLimit}
    dailyEarnings={dailyEarnings}
    onSubmit={onSubmit}
    enabledButtonKey={t('supplyWithdraw.withdraw')}
    disabledButtonKey={t('supplyWithdraw.enterValidAmountWithdraw')}
    maxInput={asset.supplyBalance}
    calculateNewBalance={calculateNewBorrowAmount}
  />
)}
```

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
coreyar: The duplicate value references do need to be removed.

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
coreyar: I think the array is preferable because it supports more than two tabs. I imagine that when we update the tab to mange its own state and render children that it will use an array

therealemjy: Ok got you. Sounds good to me in that case

therealemjy:
I had another look and tested the component and spotted a few issues:
- The text underneath the input on the Withdraw form says "Wallet ballance: X" when it should say "Withdrawable amount: X"
- The "Daily Earnings" value doesn't display what the updated value would be, on either form when changing the input value
- The component displaying the borrow limit update behaves incorrectly. Passed a certain value, the arrow becomes red when it should stay green all the way:
Screen Shot 2022-04-18 at 08 57 41
Screen Shot 2022-04-18 at 08 57 37
- Some spacing issues:
Screen Shot 2022-04-18 at 08 59 37

Designs:
Screen Shot 2022-04-18 at 09 00 15

coreyar:
> * The "Daily Earnings" value doesn't display what the updated value would be, on either form when changing the input value

This is ticked as a todo. I stopped working on daily earnings because we are going to extract it to a reusable function after this [PR ](https://github.com/VenusProtocol/venus-protocol-interface/pull/299)is merged per a discussion in slack. 

TODO is updated to reference both [Wire up ticket](https://app.clickup.com/t/24quhp4) and [daily earnings ticket](https://app.clickup.com/t/26pg8j3)

therealemjy:
Got you, perfect. Thank you.

therealemjy (approved):
Looks good 👍

therealemjy (approved):

PR #309, opened 2022-04-15
Return borrow balance and limit from user market info

coreyar:
We are calculating the users borrow limit and total borrow balance when setting up the asset list with the user's asset information. This PR updates useUserMarketInfo to return an object with the comprehensive asset list and users borrow balance and limit


therealemjy (approved):
Nice 👍

PR #310, opened 2022-04-16
Update style of using Trans component

coreyar:
Update Trans component to use props to configure components rather than default text style

### DIFF
--- a/src/components/v2/AuthModal/WalletList/index.tsx
+++ b/src/components/v2/AuthModal/WalletList/index.tsx
@@ -56,18 +56,19 @@ export const WalletList: React.FC<IWalletListProps> = ({ onLogin }) => {
 
       <div css={styles.footer}>
         <Typography variant="small2">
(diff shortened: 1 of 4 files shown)


src/components/v2/AuthModal/WalletList/index.tsx
```diff
@@ -56,18 +56,19 @@ export const WalletList: React.FC<IWalletListProps> = ({ onLogin }) => {
...
-              css={styles.footerLink}
-            >
-              Terms of Service
-            </a>
-            .
-          </Trans>
+          <Trans
+            i18nKey="authModal.walletList.termsOfServiceLink"
+            components={{
+              anchor: (
```
therealemjy: Since this is a component, how about capitalizing its name (same goes for other `Trans` instances)?

therealemjy (approved):

src/components/v2/AuthModal/WalletList/index.tsx
coreyar: works for me

PR #311, opened 2022-04-18
Fix size of icon in TextField

therealemjy:
Following the update of the `spacing` constant, the size of the icon in the `TextField` component needed to be doubled.


coreyar (approved):

coreyar:
Weird that this didn't break the viz reg threshold

PR #313, opened 2022-04-18
Move flag to switch individual routes

coreyar:
- Move App root from `containers` to `pages`
- Move v2 flag to switch on individual routes


therealemjy (approved):
Nice

PR #317, opened 2022-04-19
Fix: Add layout width for content block

baomastr:


coreyar:
These viz reg tests should not be marked as new

coreyar (approved):

therealemjy:
Unfortunately this doesn't solve the issue:
Screen Shot 2022-04-20 at 08 20 27

Maybe park this for now. I think the issue has to do with the content of the dashboard rather than its container.

src/assets/styles/Global.scss
```diff
@@ -31,6 +31,9 @@
   --color-border-main: #eaeaea;
   --color-primary: var(--color-yellow);
   --color-secondary: #d3d3d3;
+
+  /* Scrollbar width */
+  --scrollbar-width: 2px;
```
therealemjy, line 36: Let's not rely on our css files. In the future we'll ideally get rid of them completely.

therealemjy (changes requested):

src/assets/styles/Global.scss
baomastr: it was made just for understanding where we get this value (2px) for our new layout.

baomastr:
> Maybe park this for now. I think the issue has to do with the content of the dashboard rather than its container.

I think, we should both avoid scrolling of layout and fix the content to get parent width. 
I'll add the overflow: hidden to avoid horizontal scrolling of layout. And then we should fix the tables component

therealemjy:
I'd avoid both setting a fixed width and applying `overflow: hidden`.
The content should be stylized so it doesn't overflow, without the need for modifications on the parent container. If it was to overflow because of an issue, we'd want the scrollbar to be there so users can still see all the content and keep using the website (even if it wouldn't be the best UX, at least we wouldn't block them from being able to do actions).

baomastr:
Okay then. Will close the PR and fix the tables component in new ticket.

PR #319, opened 2022-04-19
Display dailyEarningsCents correctly when 0

coreyar:
Updates caclulation to return dailyEarningsCents when yearlyEarningsCents is not NaN


therealemjy (commented):
Shouldn't we check why `yearlyEarningsCents` is a `NaN` instead?
If it only happens on testnet because of some values that are incorrect, I don't think we should provide a fix for it.
If it happens on mainnet too, then the safeguard inside the `MyAccount` container isn't working properly.

coreyar:
It is related to the testnet values

PR #320, opened 2022-04-19
Update header UI

therealemjy:
- fix issue where the loading icon displayed inside the `Button` component would push the container and extend its height
- create `ClaimXvsRewardButton` (UI only)
- add `ClaimXvsRewardButton` to layout (header and mobile menu)
- update layout to fit designs
- remove `CoinInfo` component from `Layout`


src/components/v2/Layout/ClaimXvsRewardButton/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+          amount: readableAmount,
+        }}
+      />
+    </SecondaryButton>
+  );
+};
+
+export const ClaimXvsRewardButton: React.FC<IButtonProps> = props => {
+  const claimXvs = () => {
+    // @TODO: send transaction to claim XVS
```
coreyar, line 51: Can we add a ticket?

src/components/v2/Layout/Header/index.tsx
```diff
@@ -29,17 +32,13 @@ const Header = () => {
...
-          flex={1}
-          justifyContent="right"
-          css={styles.rightItemContainer}
-        >
-          <XvsCoinInfo css={styles.rightItemPaper} className="coinInfo" />
-          <VaiCoinInfo css={styles.rightItemPaper} className="coinInfo" />
-          <ConnectButton css={styles.rightItemPaper} title={t('header.connectButton.title')} />
-        </Box>
+
+        {!isOnMobile && (
```
coreyar, line 36: Any reason not to use 'display: none;' here? It is nicer to let css handle this then listening for window changes in JS.

src/components/v2/Layout/Header/styles.ts
```diff
@@ -23,11 +23,6 @@ export const useStyles = () => {
         padding: ${theme.spacing(6, 4, 0)} !important;
       }
     `,
-    rightItemContainer: css`
```
coreyar, line 26: I'm curious why this query didn't work.

coreyar (approved):

src/components/v2/Layout/Header/styles.ts
therealemjy: It hides the SVG icon of the logo, which also impacts the icon of the XVS token. We had this issue before.
I agree using pure css would be a better option here, but I couldn't find an alternative that wouldn't impact both SVGs :/

src/components/v2/Layout/ClaimXvsRewardButton/index.tsx
therealemjy: There's none directly related to that. This is part of the work on the header which I'm working on right now, there will be a PR open for that today.

PR #323, opened 2022-04-20
Add Success modal

coreyar:
Adds a success modal to display after a transaction to withdraw or supply is successful

### DIFF
--- a/src/pages/Dashboard/Modals/Success/index.tsx
+++ b/src/pages/Dashboard/Modals/Success/index.tsx
@@ -1,0 +1,44 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+import { Typography } from '@mui/material';
+import { BscLink, Modal, Icon, IModalProps } from 'components';
+import { useStyles } from '../styles';
+
+export interface ISuccessModalProps {
+  className?: string;
+  onClose: IModalProps['handleClose'];
+  isOpen: boolean;
+  title: string;
+  description: string | React.ReactElement;
+}
+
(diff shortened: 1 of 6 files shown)


src/pages/Dashboard/Modals/Success/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+            {title}
+          </Typography>
+          <Typography css={styles.successMessage} variant="small1">
+            {description}
+          </Typography>
+          <BscLink
+            css={styles.bscScan}
+            hash="0x6b8a5663cd46f7b719391c518c60e2f45427b95a082e3e47739b011faccbfc96"
+          />
+        </div>
```
therealemjy, line 38: We have a component for this interface already: `src/components/v2/SuccessfulTransactionMessage`

src/pages/Dashboard/Modals/Success/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+      </>
+    </Modal>
+  );
+};
+
+export default SuccessModal;
```
therealemjy, line 44: I think that's a good idea and we should make it global to the app. It will be reused quite a lot, so other parts of the app can benefit from it.
I actually created a ticket exactly for that. It's assigned to me, but feel free to pick it up if you want: https://app.clickup.com/t/2849k4u
Just one thing if you pick it up: we'll need this done today (Wednesday), because the release `0.3.1` depends on it (it's used in the "Claim XVS" flow). Let me know how you feel about this.

PR #325, opened 2022-04-20
Add queries and mutations for withdrawing and supplying

coreyar:
### Queries and mutations for withdrawing and supplying

Adds queries
- getBalanceOf

Add mutations
- supply
- supplyBnb
- redeem
- redeemUnderlying


src/clients/api/queries/useGetBalanceOf.ts
```diff
@@ -0,0 +1,31 @@
...
+
+type Options = QueryObserverOptions<
+  GetBalanceOfOutput,
+  Error,
+  GetBalanceOfOutput,
+  GetBalanceOfOutput,
+  [FunctionKey.BALANCE_OF, VTokenId]
+>;
+
+const useGetBalanceOf = (
```
therealemjy: I'd suggest renaming this hook so it's obvious only Venus tokens are accepted (we could have another function that takes a `TokenId` as param and returns the balance for that token). Maybe `useGetVTokenBalance` or something like that

src/constants/functionKey.ts
```diff
@@ -5,6 +5,7 @@ enum FunctionKey {
   GET_ASSETS_IN_ACCOUNT = 'GET_ASSETS_IN_ACCOUNT',
   GET_VTOKEN_BALANCES_ALL = 'GET_VTOKEN_BALANCES_ALL',
   GET_HYPOTHETICAL_LIQUIDITY = 'GET_HYPOTHETICAL_LIQUIDITY',
+  BALANCE_OF = 'BALANCE_OF',
```
therealemjy: This function key needs to be prefixed with `GET_` (it makes it obvious it's a query).

src/clients/api/mutations/useSupply.ts
```diff
@@ -0,0 +1,65 @@
...
+      });
+    },
+    {
+      ...options,
+      onSuccess: (data, variables, context) => {
+        queryClient.invalidateQueries(FunctionKey.GET_VTOKEN_BALANCES_ALL);
+        if (options?.onSuccess) {
+          options.onSuccess(data, variables, context);
+        }
+      },
```
therealemjy: A small thing: the params of `onSuccess` could change in a future update of React Query. For that reason, it's safer to just pass all the params:
```suggestion
      onSuccess: (...onSuccessParams) => {
        if (options?.onSuccess) {
          options.onSuccess(...onSuccessParams);
        }
        
        queryClient.invalidateQueries(FunctionKey.GET_VTOKEN_BALANCES_ALL);
      },
```

src/clients/api/mutations/useSupply.ts
```diff
@@ -0,0 +1,65 @@
...
+          account,
+          amount,
+        });
+      }
+      return supply({
+        tokenContract: vContract,
+        account,
+        amount,
+      });
+    },
```
therealemjy: I feel like this should be two separate hooks, for these reasons:
- `vContract` is always declared, but unused if `assetId` is `bnb` or if no `web3` instance was provided
- `supplyBnb` and `supply` have expect different inputs: `supplyBnb` expects `web3` to be provided while `supply` doesn't

Separating these into two different hooks would solve these issues, and also means you could write `useSupplyBnb` so it fetched the `web3` instance automatically using the `useWeb3` hook. This way its `mutate` function would only need to be passed the account address and an amount.

src/clients/api/mutations/supplyBnb.ts
```diff
@@ -0,0 +1,32 @@
...
+  web3,
+  tokenContract,
+  account,
+  amount,
+}: ISupplyBnbInput): Promise<SupplyBnbOutput> => {
+  const contractData = tokenContract.methods.mint().encodeABI();
+  const tx = {
+    account,
+    to: vBnbAddress,
+    value: amount.toString(),
```
therealemjy:
```suggestion
    value: amount.toFixed(),
```

So I found out yesterday that the `toString` method will write large numbers just like they're written in JavaScript: 1e20 for example. I've had a few issues with it on the front end already, so it might be safer to use `toFixed()` instead.

src/clients/api/mutations/supplyBnb.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export type SupplyBnbOutput = TransactionReceipt;
+
+const supplyBnb = async ({
+  web3,
+  tokenContract,
+  account,
+  amount,
+}: ISupplyBnbInput): Promise<SupplyBnbOutput> => {
+  const contractData = tokenContract.methods.mint().encodeABI();
```
therealemjy: You can use `getTokenContract` to get an instance of the contract. It will be strongly typed and you can call its methods directly:
```
await tokenContract.methods.mint(account, vBnbAddress, amount.toFixed()).send({ from: account });
```

src/clients/api/mutations/redeemUnderlying.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+  test('throws an error when request fails', async () => {
+    const fakeContract = {
+      methods: {
+        redeemUnderlying: () => ({
+          send: async () => {
+            throw new Error('Fake error message');
+          },
+        }),
+      },
+    } as any;
```
therealemjy, line 13: Small thing: we now have access to the contract types (`types/contracts`).

package.json
```diff
@@ -77,6 +77,7 @@
     "sass": "^1.49.7",
     "styled-components": "^4.4.1",
     "web3": "^1.7.1",
+    "web3-core": "^1.7.3",
```
therealemjy: Since this is only used for types, it should be added to the dev dependencies.
TypeScript will complain that it needs to be added to the main dependencies when you try to import it, so you'll need to change your import to:
```
import type { TransactionReceipt } from 'web3-core';
```

src/clients/api/mutations/redeemUnderlying.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export type RedeemUnderlyingOutput = void;
+
+const redeemUnderlying = async ({
+  tokenContract,
+  account,
+  amount,
+}: IRedeemUnderlyingInput): Promise<RedeemUnderlyingOutput> =>
+  tokenContract.methods.redeemUnderlying(amount).send({ from: account! });
```
therealemjy, line 14: I'd avoid using the `!` here. If you want account to also be defined, I'd change its type to disallow being `undefined`.

src/clients/api/mutations/supply.ts
```diff
@@ -0,0 +1,15 @@
...
+export interface ISupplyInput {
+  tokenContract: Bep20;
+  account: string | undefined;
+  amount: string;
+}
+
+export type SupplyOutput = void | TransactionReceipt;
+
+const supply = async ({ tokenContract, account, amount }: ISupplyInput): Promise<SupplyOutput> =>
+  tokenContract.methods.mint(amount).send({ from: account || undefined });
```
therealemjy, line 13: `account` can only be a string or undefined, so no need to default it `undefined`.

therealemjy (approved):
Just two minor comments, I'm happy to merge that anyway!

PR #328, opened 2022-04-20
Add content render for Tabs

baomastr:


src/components/v2/Tabs/styles.tsx
```diff
@@ -5,14 +5,34 @@ export const styles = () => {
...
 
   return {
-    getContainer: ({ fullWidth }: { fullWidth: boolean }) => css`
+    getContainer: ({ fullWidth, hasTitle }: { fullWidth: boolean; hasTitle: boolean }) => css`
       display: flex;
 
+      ${hasTitle &&
+      css`
+        align-items: center;
+        margin-bottom: ${theme.spacing(4)};
```
therealemjy: I think we should apply this margin all the time.
I'd also make it `24px`, since it seems to be the smallest value used in the designs:
Screen Shot 2022-04-20 at 16 03 19

baomastr: done

therealemjy (approved):
Looks good :)

PR #332, opened 2022-04-20
Wire up "Claim XVS" button

therealemjy:
- add necessary queries and mutations to API client
- connect "Claim XVS" container
- add decimals to vTokens in constants


baomastr (approved):

src/clients/api/queries/getXvsReward/getVTokenData.ts
```diff
@@ -0,0 +1,65 @@
...
+  const [
+    supplyState,
+    borrowState,
+    userBorrowIndex,
+    userSupplyIndex,
+    tokenBorrowIndex,
+    userBorrowBalanceStoredWei,
+    userBalanceWei,
+  ] = await Promise.all([
+    comptrollerContract.methods.venusSupplyState(tokenAddress).call(),
```
coreyar, line 44: We've been setting up every contract call as a separate queries and then grouping them together in a hook

src/clients/api/queries/getXvsReward/index.ts
```diff
@@ -0,0 +1,107 @@
...
+        web3,
+        comptrollerContract,
+        tokenId: vToken.id,
+        tokenAddress: vToken.address,
+        accountAddress,
+      }),
+    ),
+  );
+
+  const xvsEarned = vTokensData.reduce(
```
coreyar, line 47: Similarly I wonder if we shouldn't keep queries as pure fetching and setup and transformations of the returned data in a hook

PR #335, opened 2022-04-20
Call withdraw and supply contract methods when submitting Form

coreyar:
Adds calls to contracts with submitting supply and withdraw form


src/clients/api/__mocks__/index.ts
```diff
@@ -36,5 +40,14 @@ export const useRepayVai = () => useMutation(FunctionKey.REPAY_VAI, repayVai);
 export const enterMarkets = jest.fn();
 export const useEnterMarkets = () => useMutation(FunctionKey.ENTER_MARKETS, enterMarkets);
 
-export const exitMarket = jest.fn();
-export const useExitMarket = () => useMutation(FunctionKey.EXIT_MARKET, exitMarket);
```
therealemjy: Should this be removed or is it unintentional?

src/clients/api/mutations/supplyBnb.ts
```diff
@@ -0,0 +1,34 @@
...
+export type SupplyBnbOutput = TransactionReceipt;
+
+const supplyBnb = async ({
+  web3,
+  tokenContract,
+  account,
+  amount,
+}: ISupplyBnbInput): Promise<SupplyBnbOutput> => {
+  const contractData = tokenContract.methods.mint().encodeABI();
+  const tx = {
```
therealemjy, line 24: That's a small thing, but given you're passing a `web3` instance you could grab the contract here directly (so you wouldn't have to pass the contract too).

src/clients/api/mutations/useRedeemUnderlying.ts
```diff
@@ -8,23 +8,24 @@ import {
   RedeemUnderlyingOutput,
 } from 'clients/api';
 import FunctionKey from 'constants/functionKey';
-import { useTokenContract } from 'clients/contracts/hooks';
+import { useVTokenContract } from 'clients/contracts/hooks';
 
 const useRedeemUnderlying = (
-  { assetId }: { assetId: VTokenId },
+  { assetId, account }: { assetId: VTokenId; account: string | undefined },
```
therealemjy: I think we should always have `account` as required and defined for mutations that require it. I checked the `redeem` function and although it allows `account` to be a `string` or `undefined`, it then asserts it is defined using an exclamation mark. That's risky, because it silents TypeScript and so can lead to runtime errors.

For cases where `account` can be undefined but we still need to render the hook, we should use the `enabled` prop:
```tsx
useRedeemUnderlying({
  assetId,
  account: accountAddress || ''
}, {
  enabled: !!account
});
```

src/clients/api/mutations/useSupply.ts
```diff
@@ -1,21 +1,26 @@
...
+
 import FunctionKey from 'constants/functionKey';
-import { useTokenContract } from 'clients/contracts/hooks';
+import { useVTokenContract } from 'clients/contracts/hooks';
+
+export type SupplyParams = Omit<ISupplyInput, 'tokenContract' | 'account'>;
 
 const useSupply = (
-  { assetId }: { assetId: TokenId },
+  { assetId, account }: { assetId: VTokenId; account: string | undefined },
```
therealemjy: Same thing here about forcing `account` to be defined

src/hooks/useSupply.ts
```diff
@@ -0,0 +1,30 @@
...
+const useSupply = ({ asset, account }: IUseSupplyArgs) => {
+  const { mutate: supplyNonBnb, isLoading: isSupplyNonBnBLoading } = useSupplyNonBnb({
+    assetId: asset?.id as VTokenId,
+    account,
+  });
+  const { mutate: supplyBnb, isLoading: isSupplyBnbLoading } = useSupplyBnb({ account });
+  if (asset.id === 'bnb') {
+    return { supply: supplyBnb, isLoading: isSupplyBnbLoading };
+  }
+  return { supply: supplyNonBnb, isLoading: isSupplyNonBnBLoading };
```
therealemjy: The other props returned by the mutations are useful too, we should return them all:
```suggestion
const useSupplyNonBnbResult = useSupplyNonBnb({
  assetId: asset.id as VTokenId,
  account,
});
const useSupplyBnbResult = useSupplyBnb({ account });

return asset.id === 'bnb' ? useSupplyBnbResult : constuseSupplyNonBnbResult;
```

src/hooks/useWithdraw.ts
```diff
@@ -0,0 +1,29 @@
...
+  const { mutate: redeemUnderlying, isLoading: isRedeemUnderlyingLoading } = useRedeemUnderlying({
+    assetId: asset?.id as VTokenId,
+    account,
+  });
+  return {
+    redeem,
+    redeemUnderlying,
+    isLoading: isRedeemLoading || isRedeemUnderlyingLoading,
+    vTokenBalance,
+  };
```
therealemjy: Given these mutations don't exchange results between one another, I don't think they should be merged into one hook, because doing that makes us lose functionalities (`onSuccess`, `onError` callbacks for example) as well as some of the results they return (errors for examples).

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -14,37 +14,52 @@ import {
...
   className,
   onClose,
   asset,
   userTotalBorrowBalance,
   userTotalBorrowLimit,
   dailyEarnings,
+  supply,
+  redeem,
+  redeemUnderlying,
+  isTransactionLoading,
```
therealemjy: This should be separated into two loading states. We don't need to block users from making another transaction (say supply tokens) while one is loading (say borrow tokens)

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -92,69 +123,81 @@ export const SupplyWithdrawUi: React.FC<ISupplyWithdrawUiProps> = ({
...
   const { userTotalBorrowBalance, userTotalBorrowLimit } = useUserMarketInfo({
     account: account?.address,
   });
+  const { supply, isLoading: isSupplyLoading } = useSupply({ asset, account: account?.address });
+  const {
+    redeem,
+    redeemUnderlying,
+    vTokenBalance,
+    isLoading: isWithdrawLoading,
+  } = useWithdraw({ asset, account: account?.address });
```
therealemjy: We need to handle errors as well. It doesn't have to be complicated, we can just display them in a toast, but we need to give an indication to users when a transaction fails.

src/clients/api/mutations/useSupply.ts
```diff
@@ -1,21 +1,26 @@
...
+import queryClient from 'clients/api/queryClient';
+import supply, { ISupplyInput, SupplyOutput } from 'clients/api/mutations/supply';
+
 import FunctionKey from 'constants/functionKey';
-import { useTokenContract } from 'clients/contracts/hooks';
+import { useVTokenContract } from 'clients/contracts/hooks';
+
+export type SupplyParams = Omit<ISupplyInput, 'tokenContract' | 'account'>;
 
 const useSupply = (
```
therealemjy, line 26: I think we should use a more explicit name, like `useSupplyNonBnb`. Otherwise we have two reusable `useSupply` hooks in the codebase, it might get confusing to know which one to use

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -92,69 +123,81 @@ export const SupplyWithdrawUi: React.FC<ISupplyWithdrawUiProps> = ({
...
+                    disabledButtonKey={t('supplyWithdraw.enterValidAmountSupply')}
+                    maxInput={asset.walletBalance}
+                    calculateNewBalance={calculateNewSupplyAmount}
+                    isTransactionLoading={isTransactionLoading}
+                  />,
+                  <SupplyWithdrawForm
+                    key="withdraw"
+                    asset={asset}
+                    tokenInfo={tokenInfo}
+                    userTotalBorrowBalance={userTotalBorrowBalance}
```
therealemjy, line 154: I caught this just now, but here we'll need to pass another prop. Currently, the withdraw form displays the total borrow balance of the user beneath the input, but it should display the amount of tokens the user have supplied instead.

src/hooks/useSupply.ts
```diff
@@ -0,0 +1,30 @@
...
+interface IUseSupplyArgs {
+  asset: Asset;
+  account: string | undefined;
+}
+
+export type UseSupplyParams =
+  | Omit<ISupplyInput, 'tokenContract' | 'assetId' | 'account'>
+  | Omit<ISupplyBnbInput, 'tokenContract' | 'assetId' | 'account'>;
+
+const useSupply = ({ asset, account }: IUseSupplyArgs) => {
```
therealemjy: We should be able to provide query options to this hook. We'll for example need to provide an `onError` callback to handle errors.

src/hooks/useSupply.ts
```diff
@@ -0,0 +1,30 @@
...
+    account,
+  });
+  const { mutate: supplyBnb, isLoading: isSupplyBnbLoading } = useSupplyBnb({ account });
+  if (asset.id === 'bnb') {
+    return { supply: supplyBnb, isLoading: isSupplyBnbLoading };
+  }
+  return { supply: supplyNonBnb, isLoading: isSupplyNonBnBLoading };
+};
+
+export default useSupply;
```
therealemjy: I think this hook should be in the "mutations" folder of the API client, because it's a mutation hook that handles sending requests. This way it could also be imported from `clients/api` directly

src/hooks/useWithdraw.ts
```diff
@@ -0,0 +1,29 @@
...
+};
+
+export default useWithdraw;
```
therealemjy: Same thing for this file, I think it should be in the "queries" folder of the API client

therealemjy (approved):

therealemjy (approved):

PR #343, opened 2022-04-22
Add calculations for daily and yearly returns and apy

coreyar:
Add utils for calculating yearly and daily returns for a single or array of assets as well as the apy


src/utilities/calculateDailyEarnings.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+import { calculateYearlyEarningsForAssets } from './calculateYearlyEarnings';
+
+describe('utilities/calculateDailyEarningsCentss', () => {
+  test('calculates yearly Earnings for array of assets', () => {
+    const earnings = calculateYearlyEarningsForAssets({
+      assets: assets as Asset[],
+      borrowBalanceCents: new BigNumber('90.0103018').multipliedBy(100),
+      isXvsEnabled: true,
+    });
+    expect(calculateDailyEarningsCents(earnings.yearlyEarningsCents || new BigNumber(0))).toBe(5);
```
therealemjy: Because we only want to test the util here, I'd just pass a value directly:
```suggestion
    expect(new BigNumber(1825).toFixed()).toBe('5');
```

src/utilities/calculateDailyEarnings.ts
```diff
@@ -0,0 +1,6 @@
+import BigNumber from 'bignumber.js';
+
+const calculateDailyEarningsCents = (yearlyEarningsCents: BigNumber) =>
+  +yearlyEarningsCents.dividedBy(365).toFixed(0);
```
therealemjy: I would return the `BigNumber` directly, so it can be directly reused for calculations.

src/utilities/calculateYearlyEarnings.ts
```diff
@@ -0,0 +1,102 @@
...
+      isXvsEnabled,
+      borrowBalanceCents,
+      supplyBalanceCents,
+      yearlyEarningsCents,
+    });
+    yearlyEarningsCents = assetYearlyEarningsCents;
+    supplyBalanceCents = assetSupplyBalanceCents;
+  });
+  return { yearlyEarningsCents, supplyBalanceCents };
+};
```
therealemjy, line 102: I don't think this function works, the return statement should be after the end of the loop

therealemjy (commented):
Looks good. I had just a few minor comments.

src/utilities/calculateYearlyEarnings.ts
coreyar: In each iteration of the loop, we take the cumulative `supplyBalanceCents` and `yearlyBalanceCents` to update them using the current asset. The returned values are then assigned to the cumulative `supplyBalanceCents` and `yearlyBalanceCents` in the outer scope to be used in the subsequent call to `calculateYearlyEarningsCents`.

Finally after iterated through all of the assets, we return the cumulative  and `supplyBalanceCents` and `yearlyBalanceCents` for all the assets

therealemjy: My bad, I thought the return statement was inside the loop, which confused me. Thanks for the explanation, that makes sense!

therealemjy (approved):

PR #346, opened 2022-04-22
Claim XVS updates

therealemjy:
- create `useSuccessfulTransactionModal` hook
- use `SuccessfulTransactionModal` to display success state after claiming XVS reward
- update invalidated queries list after claiming XVS reward
- fix default `urlType` param of `generateBscScanUrl`

### DIFF
--- a/src/utilities/generateBscScanUrl.ts
+++ b/src/utilities/generateBscScanUrl.ts
@@ -14,7 +14,7 @@ export const generateBscScanUrl = <T extends UrlType = 'address'>(
     suffix = getToken(identifier as TokenId).address;
   }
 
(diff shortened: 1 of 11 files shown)


src/utilities/generateBscScanUrl.ts
```diff
@@ -14,7 +14,7 @@ export const generateBscScanUrl = <T extends UrlType = 'address'>(
     suffix = getToken(identifier as TokenId).address;
   }
 
-  return `${BASE_BSC_SCAN_URL}/${urlType}/${suffix}`;
+  return `${BASE_BSC_SCAN_URL}/${urlType || 'address'}/${suffix}`;
```
baomastr: what do you think of defining 'address' as default value for urlType

baomastr (approved):

PR #347, opened 2022-04-22
Use daily earnings util in supply modal

coreyar:
Use the earnings utils to display daily earnings for asset


src/utilities/calculateDailyEarningsCents.spec.ts
```diff
@@ -3,6 +3,8 @@ import calculateDailyEarningsCents from './calculateDailyEarningsCents';
 
 describe('utilities/calculateDailyEarningsCentss', () => {
   test('calculates yearly Earnings for array of assets', () => {
```
therealemjy: You might have forgotten to change the name of this test

src/utilities/calculateDailyEarningsCents.spec.ts
```diff
@@ -3,6 +3,8 @@ import calculateDailyEarningsCents from './calculateDailyEarningsCents';
...
-    expect(+calculateDailyEarningsCents(new BigNumber('1924.21991227022443813375'))).toBe(5);
+    expect(calculateDailyEarningsCents(new BigNumber('1924.21991227022443813375')).toFixed(0)).toBe(
+      '5',
+    );
   });
```
therealemjy: I'm curious about this result. Since we essentially return the input divided by 365, shouldn't the input be 1865 for the result to be 5 here?

coreyar: It is 5 because I am dropping the decimals. I can update to not drop the decimals in the test

therealemjy: Yeah that part is fine, but even without the decimals then here it would be dividing 1924 by 365, which doesn't equal to 5 🤔

therealemjy: Ah right of course!

therealemjy (approved):

PR #348, opened 2022-04-22
Switch from tokenSymbol name to tokenId 

coreyar:
This Pr cleans up argument label `tokenSymbol` to be more accurately named `tokenId` since we use `asset.id` to identify the asset and `token.symbol` to display it


therealemjy (commented):
Nice 👍

therealemjy (approved):

PR #349, opened 2022-04-22
Add collateral util

coreyar:
Add utility function for calculating collateral value. It accepts an amount in wei and an asset to calculate the collateral value of the amount for the asset


src/utilities/getCollateralValue.ts
```diff
@@ -0,0 +1,12 @@
+import BigNumber from 'bignumber.js';
+import { Asset } from 'types';
+import { convertWeiToCoins } from './common';
+
+const getCollateralValue = ({ asset, amountWei }: { asset: Asset; amountWei: BigNumber }) => {
```
therealemjy: This is a minor thing, but to follow the same nomenclature as the other utils perhaps this one should be named `calculateCollateralValue`?

therealemjy (approved):

PR #351, opened 2022-04-22
Build dashboard route

baomastr:


baomastr:
we need one more ticket for fixing responsiveness of tables that displaying supply and borrow markets data to solve the issue with horizontal scrolling

src/pages/Dashboard/index.tsx
```diff
@@ -1,19 +1,72 @@
...
+      },
+      {
+        title: t('dashboard.markets.tabBorrow'),
+        content: <BorrowMarket css={styles.item} />,
+      },
+    ],
+    [],
+  );
+
+  const renderMarkets = () => {
```
coreyar: Functions labeled as render can usually be better written as a component with props

src/pages/Dashboard/index.tsx
```diff
@@ -1,19 +1,72 @@
...
+    if (isSmDown) {
+      return (
+        <div css={styles.tabsWrapper}>
+          <h4 css={[styles.tabsHeader, styles.tabsTitle]}>{t('dashboard.markets.title')}</h4>
+          <Tabs css={styles.tabsHeader} tabsContent={tabsContent} />
+        </div>
+      );
+    }
+
+    if (isLgDown) {
```
coreyar: Was there a reason that prevented us from doing this in css?

baomastr: more clear responsive cases

baomastr: there are 3 different cases for mobile/tablets/desktop with different layouts

src/pages/Dashboard/index.tsx
baomastr: refactored

src/pages/Dashboard/index.tsx
baomastr: also refactored a bit

baomastr:
horizontal scrolling fix resolved in https://github.com/VenusProtocol/venus-protocol-interface/pull/356

baomastr:
https://github.com/VenusProtocol/venus-protocol-interface/pull/361

PR #352, opened 2022-04-22
Custom contract error

coreyar:
This PR moves us in the direction of creating a standard error to be used by the application for all services. Since we plan on using axios for http requests in the future, this PR adds an custom error for contracts modeled after Axios

The call to contracts is done through the `_exectuteMethod` functions. To create a custom error decorate (override) this function by wrapping it in a try catch and creating our custom error from a thrown error

### DIFF
--- a/src/clients/contracts/getters.ts
+++ b/src/clients/contracts/getters.ts
@@ -36,12 +36,15 @@ import {
   VrtVault,
   VrtConverter,
 } from 'types/contracts';
-import { getContractAddress, getToken, getVBepToken } from 'utilities';
+import { getContractAddress, getToken, getVBepToken, setupContractErrorHandling } from 'utilities';
 import { TokenContract, VTokenContract } from './types';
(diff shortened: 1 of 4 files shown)


therealemjy (commented):
I see your approach, it makes sense but I think it's risky. It involves suppressing a lot of TS errors, which I'm really not comfortable with.

You mentioned during sprint planning that since there wasn't just one way contracts dealt with errors, we should handle this on a case-by-case basis. I still think this is a good idea.
That's why I wouldn't edit the contracts themselves, but instead have `try/catch` blocks on each mutation and return a custom error from the `catch` block when one is returned/thrown.

coreyar:
If we like this approach beyond the TS ignores we can sort those out. If we prefer to do this on a case by case basis then we can close this.

therealemjy:
Sounds good to me. I think the case by case basis makes sense

PR #356, opened 2022-04-25
Fix: table scroll

baomastr:


baomastr:
https://github.com/VenusProtocol/venus-protocol-interface/pull/360

PR #359, opened 2022-04-25
Chore: unify account fetching

baomastr:


src/clients/api/mutations/enterMarkets.ts
```diff
@@ -1,6 +1,8 @@
+import { IAccount } from 'context/AuthContext';
+
 export interface IEnterMarketsInput {
   comptrollerContract: $TSFixMe; // @TODO: use contract type (through Typechain?)
-  account: string | undefined | null;
+  account: IAccount;
```
therealemjy: I don't think we should change the implementation here, since we only need the address.
We can change the name and avoid using `null` eventually:
```suggestion
  accountAddress: string | undefined;
```

src/clients/api/mutations/exitMarket.ts
```diff
@@ -1,6 +1,8 @@
+import { IAccount } from 'context/AuthContext';
+
 export interface IExitMarketInput {
   comptrollerContract: $TSFixMe; // @TODO: use contract type (through Typechain?)
-  account: string | undefined | null;
+  account: IAccount;
```
therealemjy: I don't think we should change the implementation here, since we only need the address.
We can change the name and avoid using `null` eventually:
```suggestion
  accountAddress: string | undefined;
```

therealemjy (commented):
Nice one 👍
Just had two comments

src/clients/api/mutations/enterMarkets.ts
baomastr: done

src/clients/api/mutations/exitMarket.ts
baomastr: done

coreyar:
I started to follow a similar strategy.
After accessing the account from the context, assign the address to a variable with a default value of an empty string (we could even create an account class and do this in a getter). Pass this variable to accounts and type all contract calls that require an account with `account: string`.

In this way we have clean code for using the account address and all our contracts are typed correctly based on how they work.

If we are going to make changes to previous screens, I think we should consider turning their viz reg tests back on so that we have some minimal test coverage

therealemjy (commented):
Very nice 👍
I'll approve as soon as the checks pass.

therealemjy (approved):

PR #360, opened 2022-04-25
Fix: table scroll

baomastr:


src/components/v2/Table/styles.ts
```diff
@@ -18,9 +18,8 @@ export const useStyles = () => {
       .MuiTableCell-root {
         border-width: 0;
         font-weight: ${theme.typography.body1.fontWeight};
-        > * {
-          float: right;
-        }
+        text-align: right;
+        font-size: 14px;
```
therealemjy: Let's use the theme

src/components/v2/Table/styles.ts
```diff
@@ -29,7 +28,12 @@ export const useStyles = () => {
 
       .MuiTableCell-root:first-child {
         padding-left: ${theme.spacing(6)};
-        float: left;
+        text-align: left;
+        max-width: 110px;
```
therealemjy: Two things:
- we've agreed to use `theme.spacing` to express values like this
- we shouldn't have hard-coded values like this in my opinion. At the moment I'm guessing you're updating this component so it displays well on the dashboard, however that component will be reused on other screens that won't use the same columns. We should think of a way to make this component reusable

src/components/v2/Table/styles.ts
```diff
@@ -29,7 +28,12 @@ export const useStyles = () => {
...
+      }
+
+      .MuiTableCell-root:nth-child(2) {
+        max-width: 100px;
```
therealemjy: Same thing about as previous comment

src/components/v2/Table/styles.ts
```diff
@@ -82,5 +86,13 @@ export const useStyles = () => {
         color: ${active ? theme.palette.interactive.success : theme.palette.text.primary};
       }
     `,
+    cellWrapper: css`
+      max-width: 115px;
```
therealemjy: we've agreed to use theme.spacing to express values like this

src/components/v2/Table/styles.ts
baomastr: done

src/components/v2/Table/styles.ts
baomastr: added gridTemplateColumns prop that handles columns width.

src/components/v2/Table/styles.ts
baomastr: removed

src/components/v2/Table/styles.ts
baomastr: removed

src/components/v2/Table/index.tsx
```diff
@@ -28,6 +28,7 @@ export interface ITableProps {
   };
   rowOnClick?: (row: ITableRowProps[]) => void;
   className?: string;
+  gridTemplateColumns?: string;
```
baomastr, line 31: this prop now provides handling of columns width by parent components

therealemjy:
Looks good to me. Are you planning on fixing the tables on the dashboard in a next PR then?

baomastr:
> Are you planning on fixing the tables on the dashboard in a next PR then?

sure. it will be the next step in https://github.com/VenusProtocol/venus-protocol-interface/pull/361

therealemjy (approved):

PR #361, opened 2022-04-25
Build Dashboard route

baomastr:
https://github.com/VenusProtocol/venus-protocol-interface/pull/351

### DIFF
--- a/src/pages/Dashboard/Markets/BorrowMarket/index.tsx
+++ b/src/pages/Dashboard/Markets/BorrowMarket/index.tsx
@@ -69,13 +69,15 @@ export const BorrowMarketUi: React.FC<IBorrowMarketUiProps> = ({
   );
 };
 
-const BorrowMarket: React.FC<Omit<IBorrowMarketUiProps, 'className'>> = ({
+const BorrowMarket: React.FC<IBorrowMarketUiProps> = ({
+  className,
   isXvsEnabled,
   borrowMarketAssets,
   borrowingAssets,
   userTotalBorrowLimit,
 }) => (
   <BorrowMarketUi
+    className={className}
     borrowingAssets={borrowingAssets}
     borrowMarketAssets={borrowMarketAssets}
     isXvsEnabled={isXvsEnabled}
--- a/src/pages/Dashboard/Markets/SupplyMarket/index.tsx
+++ b/src/pages/Dashboard/Markets/SupplyMarket/index.tsx
@@ -24,13 +24,13 @@ interface ISupplyMarketProps {
 }
 
 export const SupplyMarketUi: React.FC<ISupplyMarketProps> = ({
+  className,
   isXvsEnabled,
   supplyMarketAssets,
   suppliedAssets,
   toggleAssetCollateral,
   confirmCollateral,
   setConfirmCollateral,
-  className,
 }) => {
   const [selectedAsset, setSelectedAsset] = useState<Asset | undefined>(undefined);
   const styles = useStyles();
--- a/src/pages/Dashboard/Markets/SupplyMarket/index.tsx
+++ b/src/pages/Dashboard/Markets/SupplyMarket/index.tsx
@@ -79,9 +79,10 @@ export const SupplyMarketUi: React.FC<ISupplyMarketProps> = ({
 
 const SupplyMarket: React.FC<
   Pick<ISupplyMarketProps, 'isXvsEnabled' | 'supplyMarketAssets' | 'suppliedAssets'> & {
+    className?: string;
     accountAddress: string;
   }
-> = ({ isXvsEnabled, supplyMarketAssets, suppliedAssets, accountAddress }) => {
+> = ({ className, isXvsEnabled, supplyMarketAssets, suppliedAssets, accountAddress }) => {
   const [confirmCollateral, setConfirmCollateral] = useState<Asset | undefined>(undefined);
   const { t } = useTranslation();
 
@@ -141,6 +142,7 @@ const SupplyMarket: React.FC<
 
   return (
     <SupplyMarketUi
+      className={className}
       suppliedAssets={suppliedAssets}
       supplyMarketAssets={supplyMarketAssets}
       isXvsEnabled={isXvsEnabled}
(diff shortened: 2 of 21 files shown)


therealemjy:
It looks good, however the tablet and mobile displays don't quite correspond to the designs.

In the designs, on tablet, the title next to the tabs depends on the active tab:
Screen Shot 2022-04-25 at 11 31 58
Screen Shot 2022-04-25 at 11 31 56

On mobile, the title and the tabs are outside of the card that contains the table:
Screen Shot 2022-04-25 at 11 27 06

Can you change that please?
Don't worry about the table itself, since it uses cards on mobile. We'll cover that in another ticket.

baomastr:
> Can you change that please?  

@therealemjy done

src/pages/Dashboard/Markets/BorrowMarket/index.tsx
```diff
@@ -77,7 +79,7 @@ export const BorrowMarketUi: React.FC<IBorrowMarketUiProps> = ({
   return (
     <div className={className} css={styles.tableContainer}>
       <Table
-        title={t('markets.borrowMarketTableTitle')}
+        title={isXlDown ? undefined : t('markets.borrowMarketTableTitle')}
```
coreyar: I'm hesitant to get in the habit of styling responsive styles in JS

coreyar: I also think it is easier to read if the ternaries return truthy values on the right and falsy values on the left

src/pages/Dashboard/Markets/SupplyMarket/index.tsx
```diff
@@ -13,6 +13,7 @@ import { useTranslation } from 'translation';
 import { SupplyWithdrawModal } from '../../Modals';
 import { CollateralConfirmModal } from './CollateralConfirmModal';
 import { useStyles } from '../styles';
+import { useIsXlDown } from '../../../../hooks/responsive';
```
coreyar: We can use an absolute import here. It is nice if local imports are listed last. We could even use a plugin to enforce this

src/pages/Dashboard/Markets/SupplyMarket/index.tsx
```diff
@@ -99,7 +101,7 @@ export const SupplyMarketUi: React.FC<ISupplyMarketUiProps> = ({
   return (
     <div className={className} css={styles.tableContainer}>
       <Table
-        title={t('markets.supplyMarketTableTitle')}
+        title={isXlDown ? undefined : t('markets.supplyMarketTableTitle')}
```
coreyar: same as other comment. It wasn't possible to do this using styles?

src/pages/Dashboard/Markets/BorrowMarket/index.tsx
baomastr: it is made by js because we want to control this in parent scope for current case. it's not usual case for tables.

src/pages/Dashboard/Markets/SupplyMarket/index.tsx
baomastr: done

src/pages/Dashboard/Markets/BorrowMarket/index.tsx
baomastr: do you mean using it like `!isXlDown ? t('markets.supplyMarketTableTitle' : undefined`

src/pages/Dashboard/Markets/SupplyMarket/index.tsx
baomastr: the reason of doing this with js described in first comment

src/components/v2/Tabs/index.tsx
```diff
@@ -54,7 +54,7 @@ export const Tabs = ({
             onClick={() => handleChange(index)}
             css={styles.getButton({
               active: index === activeTabIndex,
-              last: index === title.length - 1,
+              last: index === tabsContent.length - 1,
```
baomastr: @therealemjy here is the fix from your PR

src/pages/Dashboard/Markets/BorrowMarket/index.tsx
coreyar: Yeah this is an odd one because I would do like `isXLUp ? t('markets.supplyMarketTableTitle' : undefined` but we decided to have all down breakpoints.  I think the best solution here is to do it with css and not conditionally render it by subscribing to changes in the window size

src/components/v2/Tabs/index.tsx
therealemjy: Thanks

src/pages/Dashboard/Markets/BorrowMarket/index.tsx
baomastr: we don't want to remove titles for all tables. it's a special case, when we want to control titles in parent scope. that's why it was made here by js

src/pages/Dashboard/Markets/SupplyMarket/index.tsx
therealemjy: Because our `Table` component has the title embedded in it, Dmitriy is right that we unfortunately have to use JS for responsiveness.
An alternative approach would be to render the title outside of the table.

therealemjy:
Getting close to completion. I spotted a few minor design issues. By the way sorry if these aren't things you've changed here, I'm just marking them here since this PR is about building the Dashboard route. I'm happy to help working on those changes if needed.

The "My account" and "Mint Repai/Vai" sections are the same height, but we also need to align the "My account" title with "Mint Repay/Vai":
App:
Screen Shot 2022-04-27 at 11 22 30
Designs:
Screen Shot 2022-04-27 at 11 22 26


The "Supply market" and "Borrow market" sections need to have bigger top and bottom paddings:
App:
Screen Shot 2022-04-27 at 11 22 33
Screen Shot 2022-04-27 at 11 22 43

Designs:
Screen Shot 2022-04-27 at 11 22 36
Screen Shot 2022-04-27 at 11 52 49

therealemjy:
Looks good to me :) 
I agree with Corey that ideally we should use CSS for responsiveness, but as I mentioned earlier too I know the `Table` component takes a title and this title needs to be set to undefined based on the resolution. As I said, the alternative would be to render the title outside the table.
For now, I think it's great and we can always make this improvement later.

therealemjy (approved):

PR #363, opened 2022-04-25
Update convert icon + fix external link in sidebar

therealemjy:
- update "convert" icon
- update "Convert XVS" sidebar item label to "Convert VRT"
- use anchor tag when rendering external link in sidebar


baomastr (approved):

src/components/v2/Layout/Sidebar/index.tsx
```diff
@@ -37,6 +37,36 @@ export const SidebarUi: React.FC<ISidebarProps> = () => {
     setAnchorEl(null);
   };
 
+  const renderLink = ({ href, icon, i18nKey }: IMenuItem) => {
```
coreyar, line 40: Functions with render in the name can be better written as components

PR #365, opened 2022-04-25
Create AccountHealth component

therealemjy:
- rename `ProgressBarHorizontal` to `ProgressBar`
- create `AccountHealth` component that renders a `ProgressBar` instance with a legend on top of it
Screen Shot 2022-04-25 at 18 11 31

In context:
Screen Shot 2022-04-25 at 18 09 59
Screen Shot 2022-04-25 at 18 11 28


src/components/v2/ProgressBar/index.tsx
```diff
@@ -82,7 +80,7 @@ export const ProgressBarHorizontal = ({
       min={min}
       max={max}
       size="medium"
-      disabled={isDisabled}
+      disabled
```
coreyar, line 83: Should this be statically disabled?

src/components/v2/ProgressBar/AccountHealth/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const readableBorrowBalance = formatCentsToReadableValue({
+    value: borrowBalanceCents,
+  });
+
+  return (
+    <div className={className}>
+      <div css={styles.topProgressBarLegend}>
+        <div css={styles.inlineContainer}>
+          <Typography component="span" variant="small2" css={styles.inlineLabel}>
+            {variant === 'primary'
```
coreyar: I think this component would be more flexible if we dropped the variant and made these props.
If we keep the variant, I would label it more descriptive such as 'borrow' and 'supply'

src/components/v2/ProgressBar/index.tsx
baomastr: yes. we dont't have any cases with changing values of slider manually by user

baomastr (approved):

src/components/v2/ProgressBar/AccountHealth/index.tsx
therealemjy: I'd rather keep it as strict as possible, since these are the only two existing variants.
"borrow" and "supply" wouldn't describe what this actually is. I tried to come up with names that would make sense, but it's actually quite tricky ahah
One variant displays the borrow limit used, the other one display the borrow balance in dollars. Maybe the variants could be named "borrowLimitUsed" and "borrowBalance"?

coreyar: "borrowLimitUsed" and "borrowBalance"  work for me

coreyar (approved):

PR #373, opened 2022-04-26
Fix how last tab is defined on Tabs component

therealemjy:
The `Tabs` component was using the wrong prop to define which tab is the last.


baomastr:
It is already fixed in neighbor PR

baomastr:
https://github.com/VenusProtocol/venus-protocol-interface/pull/361/files#diff-c0e46dd2c5468e987e524ea4f4b0f206be2199aee093c913a895c6f18a7e33d5R57

therealemjy:
Ok sounds good, I'll close it then

PR #374, opened 2022-04-26
Add a secondary value to the progress bar

coreyar:
This PR adds a secondaryValue prop to the progress bar that is displayed below the primary value. If the primary value surpasses the secondary value it would be hidden


therealemjy (commented):
Looks good. Can we make it so the secondary rail displays in red if it goes beyond the `mark` value?
Like that:
Screen Shot 2022-04-26 at 22 22 32

therealemjy:
Nice one 👍

therealemjy (approved):

PR #375, opened 2022-04-26
Add Borrow tab of BorrowRepay modal

therealemjy:
- create `calculatePercentage` util
- update `ValueUpdate` component to make its type stricter + add `positiveDirection` prop
- add basic `BorrowRepay` modal
- add `Borrow` tab and wire it up
- update styles used by `SupplyWithdraw` and `BorrowRepay` modals so they can share them 
- update `initialValues` prop of `AmountForm` to `initialAmount`
- add optional `maxAmount` prop to `AmountForm` and update validation schema to account for it
- create global `walletNotConnected` error message in translations


src/pages/Dashboard/Markets/BorrowMarket/index.tsx
```diff
@@ -74,19 +76,33 @@ export const BorrowMarketUi: React.FC<IBorrowMarketUiProps> = ({
...
-        data={rows}
-        initialOrder={{
-          orderBy: 'apy',
-          orderDirection: 'asc',
-        }}
-        rowKeyIndex={0}
-      />
-    </div>
+    <>
+      <div className={className} css={styles.tableContainer}>
```
coreyar: We should use Paper here

src/pages/Dashboard/Markets/BorrowMarket/index.tsx
```diff
@@ -74,19 +76,33 @@ export const BorrowMarketUi: React.FC<IBorrowMarketUiProps> = ({
...
+          initialOrder={{
+            orderBy: 'apy',
+            orderDirection: 'asc',
+          }}
+          rowOnClick={rowOnClick}
+          rowKeyIndex={0}
+        />
+      </div>
+
+      {selectedAsset && (
```
coreyar, line 65: In PR 372 I put each table as its own component and laid them out with modals in the index file.

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -255,6 +255,7 @@ const SupplyWithdrawModal: React.FC<ISupplyWithdrawUiProps> = props => {
   // @TODO: elevate state so it can be shared with borrow and supply markets
   const isXvsEnabled = true;
   const borrowBalanceCents = userTotalBorrowBalance.multipliedBy(100);
+  // @TODO: include all assets in calculation of yearly earnings
```
coreyar, line 258: Why would we want to see the yearly earnings for all assets when we are supplying a single asset? It seems like you would want to see yearly earnings for the specific asset you are supplying to inform your supply decision

therealemjy: That's how I understood the designs, I'll check with everyone.

therealemjy: Pavel just confirmed we want to display the total daily earnings:
Screen Shot 2022-04-27 at 17 42 38

I think it makes sense, because here we just say "Daily earnings", we don't specify that they only concern the token that's relevant to the modal.

src/components/v2/ValueUpdate/index.tsx
```diff
@@ -2,39 +2,57 @@
 import React from 'react';
 import Typography from '@mui/material/Typography';
 import BigNumber from 'bignumber.js';
-import { getBigNumber, formatCentsToReadableValue } from 'utilities/common';
+import { formatCentsToReadableValue } from 'utilities/common';
 import { Icon } from '../Icon';
 import { useStyles } from './styles';
 
-export interface IValueUpdateProps {
+type NumberValueProps = {
```
coreyar: nit: This can be better written as an interface

src/components/v2/ValueUpdate/index.tsx
```diff
@@ -2,39 +2,57 @@
...
+  format?: (value: number | undefined) => string;
+};
+
+type BigNumberValueProps = {
+  original: BigNumber | undefined;
+  update: BigNumber | undefined;
+  format?: (value: BigNumber | undefined) => string;
+};
+
+export type ValueUpdateProps = {
```
coreyar: nit: This could be an interface and then the type would be the union and intersection of values

src/pages/Dashboard/Modals/BorrowRepay/AccountData/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+export interface IAccountDataProps {
+  asset: Asset;
+  amount: FormikProps<FormValues>['values']['amount'];
+}
+
+const AccountData: React.FC<IAccountDataProps> = ({ asset, amount }) => {
+  const { t } = useTranslation();
+  const styles = useStyles();
+  const { account } = React.useContext(AuthContext);
```
coreyar, line 32: If we wanted to save this call the `accountAddress` and `userTotals` are passed down from the Dashboard container.

coreyar: If we want to do that we can also refactor the Supply Modal

src/pages/Dashboard/Modals/BorrowRepay/Borrow/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+              </div>
+            )}
+          </div>
+
+          <AccountData amount={values.amount} asset={asset} />
+
+          <PrimaryButton
+            type="submit"
+            loading={isBorrowLoading}
+            disabled={!isValid || !dirty || isBorrowLoading}
```
coreyar, line 114: Id think the case where the firm is valid it would always be dirty and when it is not dirty it is always invalid. Is there a specific case where we need to check for the form being valid  but not dirty?

This logic is correct but I am curious if the dirty checks are necessary.

src/pages/Dashboard/Modals/BorrowRepay/Borrow/styles.ts
```diff
@@ -0,0 +1,21 @@
+import { css } from '@emotion/react';
+import { useTheme } from '@mui/material';
+
+export const useStyles = () => {
+  const theme = useTheme();
+
+  return {
+    liquidationWarning: css`
+      border: 2px solid ${theme.palette.interactive.error};
```
coreyar:
```suggestion
      border: ${theme.spacing(0.5)} solid ${theme.palette.interactive.error};
```

src/pages/Dashboard/Markets/BorrowMarket/index.tsx
```diff
@@ -23,21 +25,47 @@ export const BorrowMarketUi: React.FC<IBorrowMarketUiProps> = ({
   isXvsEnabled,
   userTotalBorrowLimit,
 }) => {
+  const [selectedAsset, setSelectedAsset] = React.useState<Asset | undefined>(undefined);
   const styles = useStyles();
+
+  const rowOnClick: IBorrowMarketTableProps['rowOnClick'] | IBorrowingUiProps['rowOnClick'] = (
```
coreyar, line 31: Should all our market components mirror each other? The supply modals have their `rowOnClick` defined within the component. In that implementation they don't combine the assets to a single array

therealemjy: Yeah I saw that, but since the `rowOnClick` props and the `selectedAsset` state were being duplicated, I figured I could have all that here instead. We can do the same thing on the supply market table if you want

src/pages/Dashboard/Modals/BorrowRepay/AccountData/index.tsx
therealemjy: I'd rather import the account from here. We're aren't doing any additional HTTPS request (thanks to the cache) or anything that should have a significant performance impact, but this way we don't have drill down props.

Also, I wouldn't refactor the Supply modal since it's not creating any issue right now. In the case of the Borrow modal, both tabs are too different to be merged into one component, so I hoped for merging the elements that could be merged together instead.

src/components/v2/ValueUpdate/index.tsx
therealemjy: Sure, works for me

src/pages/Dashboard/Modals/BorrowRepay/Borrow/index.tsx
therealemjy: Logically speaking, you're right. But the issue is that the validation is asynchronous, so when it happens on mount the form is first marked as `valid` until the validation has happened.
Adding `dirty` means the button is disabled even before the first validation has occurred.

src/pages/Dashboard/Markets/BorrowMarket/index.tsx
coreyar: Added to https://app.clickup.com/t/28vdcgd

src/pages/Dashboard/Modals/BorrowRepay/Borrow/index.tsx
coreyar: Interesting our validation shouldn't be async and we have `validateOnMount` set so this behaviour isn't what I would expect. I checked it out and I see what you mean though

coreyar (approved):

src/pages/Dashboard/Modals/BorrowRepay/Borrow/index.tsx
therealemjy: I agree with you. I tried a few things, every time it seemed the validation on mount would take some time during which the button wouldn't be disabled :/

src/pages/Dashboard/Markets/BorrowMarket/index.tsx
therealemjy: Ok sounds good 👍

PR #377, opened 2022-04-27
Update max button prop of TokenTextField so it takes a value

therealemjy:
This is so a custom value other than the max can be assigned on click.


src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
```diff
@@ -19,7 +19,7 @@ import { useStyles } from '../styles';
 export interface IMintVaiUiProps {
   disabled: boolean;
   isMintVaiLoading: boolean;
-  mintVai: (value: BigNumber) => Promise<void>;
+  mintVai: (value: BigNumber) => Promise<unknown>;
```
therealemjy, line 22: This bug was introduced by a previous PR and prevented the tests from running, hence the fix here.

baomastr (approved):

baomastr (approved):

PR #379, opened 2022-04-27
Accept number inputs with first user input as decimal

coreyar:
## Problem
If a user enters a decimal first the input wont' parse correctly

## Solution
Updates the TextField input to check if the number input is safe to parse by adding a zero if the first char is a period


therealemjy (approved):
Elegant 👍

PR #381, opened 2022-04-27
Add shorthand readable value for tables

coreyar:
Updated formatCoinsToReadableValue to take a shorthand boolean param and formats per the following rules.

If the value is inferior to 1, we display it with up to 8 decimals.
If the value is superior or equal to 1, we display with 2 decimals only. e.g.: 26,667.87 XVS

### DIFF
--- a/src/utilities/common.ts
+++ b/src/utilities/common.ts
@@ -123,13 +123,21 @@ export const format = (bigNumber: BigNumber, dp = 2) =>
 export const formatCoinsToReadableValue = ({
   value,
   tokenId,
+  shorthand = false,
 }: {
   value: BigNumber | undefined;
   tokenId: TokenId;
-}) =>
-  value === undefined
-    ? PLACEHOLDER_KEY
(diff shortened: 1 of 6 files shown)


src/utilities/common.ts
```diff
@@ -123,13 +123,21 @@ export const format = (bigNumber: BigNumber, dp = 2) =>
...
-    : `${formatCommaThousandsPeriodDecimal(value.toFixed())} ${tokenId.toUpperCase()}`;
+  shorthand?: boolean;
+}) => {
+  if (value === undefined) {
+    return PLACEHOLDER_KEY;
+  }
+  let valueString = value.toFixed();
+  if (shorthand && value.gt(1)) {
+    valueString = value.toFixed(2);
+  }
```
therealemjy: We're not quite following the rules here: we want to show up to 8 decimals only, if the value is greater than 1. Here we're showing all the decimals if the value is greater than 1. We should also filter out trailing 0s.

coreyar: Updated

therealemjy (approved):

PR #383, opened 2022-04-27
Place AuthContext as the highest so child contexts can use it

coreyar:
Place AuthContext as the highest context so child contexts can use it


therealemjy (commented):
We should do the same in `renderComponent` (`src/testUtils/renderComponent.tsx`)

therealemjy (approved):

PR #385, opened 2022-04-28
Add Repay tab to BorrowRepay modal

therealemjy:
- add default `staleTime` of 10 seconds on query client, so query hooks that are unmounted then quickly after mounted again don't trigger a refetch
- always display red border on `TextField` when `hasError` is `true`
- rename `amount` to `hypotheticalBorrowAmountTokens` on `AccountData`
- add `Repay` tab to `BorrowRepay modal`
- add text for global error `errors.walletNotConnected` in English translation file
- move `InternalError` to `errors` util
- rename `ToastError` to `UiError`

### DIFF
--- a/src/components/v2/TextField/styles.ts
+++ b/src/components/v2/TextField/styles.ts
@@ -27,7 +27,6 @@ export const useStyles = () => {
           : theme.palette.interactive.primary};
       }
     `,
-
     leftIcon: css`
       margin-right: ${theme.spacing(2)};
     `,
--- a/src/components/v2/TextField/styles.ts
+++ b/src/components/v2/TextField/styles.ts
@@ -17,7 +17,8 @@ export const useStyles = () => {
       align-items: center;
       padding: ${theme.spacing(2, 2, 2, 4)};
       border-radius: ${theme.spacing(3)};
-      border: 2px solid ${hasError ? theme.palette.interactive.error : 'transparent'};
+      border: ${theme.spacing(0.5)} solid
+        ${hasError ? theme.palette.interactive.error : 'transparent'};
       background-color: ${theme.palette.background.default};
 
       &:focus-within {
--- a/src/components/v2/TextField/styles.ts
+++ b/src/components/v2/TextField/styles.ts
@@ -17,7 +17,7 @@ export const useStyles = () => {
       align-items: center;
       padding: ${theme.spacing(2, 2, 2, 4)};
       border-radius: ${theme.spacing(3)};
-      border: 2px solid transparent;
+      border: 2px solid ${hasError ? theme.palette.interactive.error : 'transparent'};
       background-color: ${theme.palette.background.default};
 
       &:focus-within {
--- a/src/pages/Dashboard/Modals/BorrowRepay/AccountData/index.tsx
+++ b/src/pages/Dashboard/Modals/BorrowRepay/AccountData/index.tsx
@@ -22,9 +22,14 @@ import { useStyles } from '../../styles';
 export interface IAccountDataProps {
   asset: Asset;
   hypotheticalBorrowAmountTokens: number;
+  isXvsEnabled: boolean;
 }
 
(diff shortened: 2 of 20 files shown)


src/pages/Dashboard/Modals/BorrowRepay/Repay/styles.ts
```diff
@@ -0,0 +1,26 @@
...
+    walletBalance: css`
+      margin-bottom: ${theme.spacing(3)};
+    `,
+    selectButtonsContainer: css`
+      display: flex;
+      flex-direction: row;
+    `,
+    selectButton: css`
+      flex: 1;
+      border-radius: ${theme.spacing(13)};
```
baomastr, line 17: shouldn't we have shape for this case?

baomastr (approved):

src/pages/Dashboard/Modals/BorrowRepay/Repay/styles.ts
therealemjy: These buttons are the only ones I can see in the app using this border radius, so I figured perhaps it would make more sense to use `theme.spacing` directly. Do you think we should add this to the shape?

src/components/v2/TextField/styles.ts
```diff
@@ -17,7 +17,7 @@ export const useStyles = () => {
       align-items: center;
       padding: ${theme.spacing(2, 2, 2, 4)};
       border-radius: ${theme.spacing(3)};
-      border: 2px solid transparent;
+      border: 2px solid ${hasError ? theme.palette.interactive.error : 'transparent'};
```
coreyar:
```suggestion
      border: $theme.spacing(0.5)} solid ${hasError ? theme.palette.interactive.error : 'transparent'};
```

src/pages/Dashboard/Modals/BorrowRepay/AccountData/index.tsx
```diff
@@ -74,6 +72,8 @@ const AccountData: React.FC<IAccountDataProps> = ({ asset, amount }) => {
             : assetData.borrowBalance,
       }));
 
+      // @TODO: update once calculateYearlyEarningsForAssets has been refactored
```
coreyar: https://github.com/VenusProtocol/venus-protocol-interface/pull/387

src/pages/Dashboard/Modals/BorrowRepay/Repay/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+  const { t } = useTranslation();
+  const { account } = React.useContext(AuthContext);
+
+  const { mutateAsync: repay, isLoading: isRepayLoading } = useRepayVToken({
+    vTokenId: asset.id as VTokenId,
+  });
+
+  const handleRepay: IRepayFormProps['repay'] = async amountWei => {
+    if (!account?.address) {
+      throw new Error(t('errors.walletNotConnected'));
```
coreyar: We have a specific error for Toasts which includes title and description. In this case we aren't using the description.

I wonder too if it isn't better to make it more generic like a UI error that we catch and respond in the interface that could have other properties in the future

coreyar (approved):

src/pages/Dashboard/Modals/BorrowRepay/Repay/index.tsx
therealemjy: Ah good point! Missed that.

Yeah I agree. I renamed to `ToastError` to `UiError` to express that intention, let me know your thoughts.

src/pages/Dashboard/Modals/BorrowRepay/AccountData/index.tsx
therealemjy: I'll make the update once we merge your PR

src/pages/Dashboard/Modals/BorrowRepay/Repay/index.tsx
coreyar: Sounds good to me

coreyar (approved):

PR #387, opened 2022-04-28
Update calculateYearlyEarningsCents

coreyar:
- Use asset borrowBalance
- Do all calculations in cents
- Sum after calculating earnings
- Use APY as decimal
- Return APY as percent

### DIFF
--- a/src/pages/Dashboard/MyAccount/index.tsx
+++ b/src/pages/Dashboard/MyAccount/index.tsx
@@ -15,8 +15,9 @@ interface IMyAccountProps {
   isXvsEnabled: boolean;
   setIsXvsEnabled: (value: boolean) => void;
   assets: Asset[];
-  userTotalBorrowBalance: BigNumber;
   userTotalBorrowLimit: BigNumber;
+  userTotalBorrowBalance: BigNumber;
+  userTotalSupplyBalance: BigNumber;
 }
 
 const MyAccount: React.FC<IMyAccountProps> = ({
@@ -24,35 +25,29 @@ const MyAccount: React.FC<IMyAccountProps> = ({
   assets,
   isXvsEnabled,
   setIsXvsEnabled,
-  userTotalBorrowBalance,
   userTotalBorrowLimit,
+  userTotalBorrowBalance,
+  userTotalSupplyBalance,
 }) => {
   const calculations: Pick<
     IMyAccountUiProps,
-    | 'netApyPercentage'
-    | 'dailyEarningsCents'
-    | 'supplyBalanceCents'
-    | 'borrowBalanceCents'
-    | 'borrowLimitCents'
+    'netApyPercentage' | 'dailyEarningsCents' | 'supplyBalanceCents' | 'borrowLimitCents'
   > = React.useMemo(() => {
-    const borrowBalanceCents = userTotalBorrowBalance.multipliedBy(100);
-    const { supplyBalanceCents, yearlyEarningsCents } = calculateYearlyEarningsForAssets({
+    const yearlyEarningsCents = calculateYearlyEarningsForAssets({
       assets,
-      borrowBalanceCents,
       isXvsEnabled,
     });
+    const supplyBalanceCents = userTotalSupplyBalance.multipliedBy(100);
     const netApyPercentage =
       supplyBalanceCents &&
       yearlyEarningsCents &&
       calculateApy({ supplyBalanceCents, yearlyEarningsCents });
     const dailyEarningsCents =
       yearlyEarningsCents && +calculateDailyEarningsCents(yearlyEarningsCents).toFixed(0);
-
     return {
       netApyPercentage,
       dailyEarningsCents,
       supplyBalanceCents: supplyBalanceCents?.toNumber(),
-      borrowBalanceCents: borrowBalanceCents?.toNumber(),
       borrowLimitCents: userTotalBorrowLimit.multipliedBy(100).toNumber(),
     };
   }, [JSON.stringify(assets), isXvsEnabled]);
@@ -63,6 +58,7 @@ const MyAccount: React.FC<IMyAccountProps> = ({
       safeBorrowLimitPercentage={SAFE_BORROW_LIMIT_PERCENTAGE}
       isXvsEnabled={isXvsEnabled}
       onXvsToggle={setIsXvsEnabled}
+      borrowBalanceCents={+userTotalBorrowBalance.multipliedBy(100).toFixed()}
       {...calculations}
     />
   );
--- a/src/utilities/calculateYearlyEarnings.spec.ts
+++ b/src/utilities/calculateYearlyEarnings.spec.ts
@@ -1,4 +1,3 @@
-import BigNumber from 'bignumber.js';
 import { Asset } from 'types';
 import { assetData as assets } from '__mocks__/models/asset';
 import {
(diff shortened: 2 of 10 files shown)


src/pages/Dashboard/MyAccount/index.tsx
```diff
@@ -35,19 +33,17 @@ const MyAccount: React.FC<IMyAccountProps> = ({
...
+      calculateYearlyEarningsForAssets({
+        assets,
+        isXvsEnabled,
+      });
     const netApyPercentage =
       supplyBalanceCents &&
       yearlyEarningsCents &&
       calculateApy({ supplyBalanceCents, yearlyEarningsCents });
     const dailyEarningsCents =
       yearlyEarningsCents && +calculateDailyEarningsCents(yearlyEarningsCents).toFixed(0);
```
therealemjy, line 46: I know this is out of scope of this PR, but given we're currently always going to be using `calculateYearlyEarningsForAssets` to calculate the value that's then passed to `calculateDailyEarningsCents` wouldn't it make more sense to pass `assets` to `calculateDailyEarningsCents` instead and call `calculateYearlyEarningsForAssets` from within it?

src/utilities/calculateYearlyEarnings.spec.ts
```diff
@@ -1,29 +1,26 @@
-import BigNumber from 'bignumber.js';
 import { Asset } from 'types';
 import { assetData as assets } from '__mocks__/models/asset';
 import {
   calculateYearlyEarningsForAssets,
   calculateYearlyEarningsCents,
 } from './calculateYearlyEarnings';
-
+// APY refers to the amount of interest earned on your savings and APR is how much interest you owe.
```
therealemjy: Just to clarify, an APY and an APR both represent interests, the only difference is an APY includes compounding while an APR doesn't.
Right now we're using APYs for all our calculations because that's what the backend exposes, but in the future we'll want to calculate the daily earnings using the supply/borrow APYs and the distribution APRs (rather than APYs like we do now), because after speaking with Sean and Kirill we think it's fairer to show users daily earnings calculated with the consideration that they would do no action (since they'd have to reinvest the distribution interests themselves in order to benefit from compounding).

src/utilities/calculateYearlyEarnings.ts
```diff
@@ -12,47 +12,48 @@ import BigNumber from 'bignumber.js';
...
-  assetSupplyBalanceCents = assetSupplyBalanceCents.plus(
-    asset.supplyBalance.multipliedBy(asset.tokenPrice).multipliedBy(100),
-  );
-
-  const supplyYearlyEarnings = assetSupplyBalanceCents.multipliedBy(asset.supplyApy).dividedBy(100);
+  const supplyYearlyEarnings = assetSupplyBalanceCents.multipliedBy(asset.supplyApy);
   // Note that borrowYearlyInterests will always be negative (or 0), since
   // the borrow APY is expressed with a negative percentage)
-  const borrowYearlyInterests = borrowBalanceCents.multipliedBy(asset.borrowApy).dividedBy(100);
+  const borrowYearlyInterests = assetBorrowBalanceCents.multipliedBy(asset.borrowApy);
```
therealemjy: Just to keep the same naming convention: `supplyYearlyEarningsCents` and `borrowYearlyInterests `

src/utilities/calculateYearlyEarnings.ts
```diff
@@ -12,47 +12,48 @@ import BigNumber from 'bignumber.js';
...
+  borrowBalanceCents = new BigNumber(0),
   supplyBalanceCents = new BigNumber(0),
   yearlyEarningsCents = new BigNumber(0),
 }: {
   asset: Asset;
-  borrowBalanceCents: BigNumber;
   isXvsEnabled: boolean;
+  borrowBalanceCents?: BigNumber;
   supplyBalanceCents?: BigNumber;
   yearlyEarningsCents?: BigNumber;
```
therealemjy: I think it's confusing to pass these parameters given they aren't used in the calculation of the yearly interests.

Looking at this function, I would expect to pass `asset` only and be returned the yearly earnings only. Looking at the codebase, we don't use the extra parameters returned anyway (`supplyBalanceCents` and `borrowBalanceCents`) outside of the `calculateYearlyEarningsForAssets` function, which also passes and returns them but doesn't actually do anything with them either (apart from compounding them to return the total sums). I'd get rid of these completely and simplify this function so it only returns the yearly earnings generated by the asset passed through params.

src/pages/Dashboard/MyAccount/index.tsx
coreyar: We could. I like how this clearly differentiates calculation per asset and the summation of all the assets.

therealemjy: You can still have that by having two functions, but when it comes to how they'll be used in the codebase, given the current designs, we'll always use `calculateDailyEarningsCents` with `calculateYearlyEarningsForAssets` so this way we'd only have to import one function every time instead of two.

src/utilities/calculateYearlyEarnings.ts
coreyar: We do use those values in the `MyAccount` component. `borrowBalanceCents` should always equal `userBorrowBalance` so that is redundant. `supplyBalanceCents` is used in calculating the `netApy`

src/pages/Dashboard/MyAccount/index.tsx
coreyar: You only have to import on function currently, unless I am missing something. I can change the exports to make it clearer

therealemjy: What I mean is you have to import `calculateYearlyEarningsForAssets` and `calculateDailyEarningsCents`, but because we don't actually use the value returned by `calculateYearlyEarningsForAssets`, we only pass it to `calculateDailyEarningsCents`. So if `calculateDailyEarningsCents` used the assets to call `calculateDailyEarningsCents` from within it, we'd only have to import and call `calculateDailyEarningsCents`

src/utilities/calculateYearlyEarnings.ts
therealemjy: Ok I see. then Perhaps `calculateYearlyEarningsForAssets` should be named to something else, because effectively it's not just calculating the yearly earnings but also returns more data.
I think we have to be very careful with these functions so their logic and names are as clear as water.

src/pages/Dashboard/MyAccount/index.tsx
coreyar: We use the value for return value from `calculateYearlyEarningsForAssets` in our Net APY calculation. I like `calculateDailyEarningsCents` as a separate function that does not call `calculateYearlyEarningsForAssets` because it has a single responsibility

src/utilities/calculateYearlyEarnings.ts
coreyar: Ok, yes I agree that a clear name is important. I'll update to `calculateYearlyEarningsAndSupplyBalanceForAssets` and we can look at a better solution for calculation Supply Balance in the future

src/utilities/calculateYearlyEarnings.spec.ts
therealemjy: So on that, I think we should remove this comment. I understand it as APY is always the interest earned and APR is always the interests owed, which is incorrect

therealemjy (approved):
Really nice 👍
I left one final minor comment about removing the comment that talks about APY and APR.

PR #388, opened 2022-04-28
Feature: add mobile view for tables

baomastr:


src/pages/Dashboard/Markets/BorrowMarket/BorrowMarketTable.tsx
```diff
@@ -64,6 +75,21 @@ const BorrowMarketTable: React.FC<IBorrowMarketTableProps> = ({
...
+        <h4>{t('markets.borrowMarketTableTitle')}</h4>
+        {rows.map(([asset, apy, wallet, liquidity]) => (
+          <AssetCardMobile
+            title={asset.render()}
+            columns={columnsMobile}
+            data={[[apy, wallet, liquidity]]}
+            rowOnClick={rowOnClick}
+          />
+        ))}
+      </>
```
therealemjy, line 90: I'd approach the problem differently. Since all our tables use the card layout on mobile, I think we should update the `Table` component so it automatically switches to the correct layout based on the resolution.

Also, please use CSS as much as possible to handle responsiveness rather than the JS hooks.

therealemjy (commented):
I'd approach the problem differently. Since all our tables use the card layout on mobile, I think we should update the `Table` component so it automatically switches to the correct layout based on the resolution.
If you look at the designs, you can see than the first column is always used as header, then the other ones are displayed bellow. We'll just have to be smart with table that have a lot of columns and find a way to determine how many rows we want to use (see: https://www.figma.com/file/GYJwWWLbC4B8DrJOPcETz3/Venus-2.0?node-id=1867%3A25917 for example).

src/components/v2/Table/index.tsx
```diff
@@ -29,6 +31,7 @@ export interface ITableProps {
   rowOnClick?: (e: React.MouseEvent<HTMLDivElement>, row: ITableRowProps[]) => void;
   className?: string;
   gridTemplateColumns?: string;
+  isMobileView?: boolean;
```
baomastr: This property is adjustable by parent because in different cases we need to switch to a mobile view of tables at different resolutions

src/components/v2/Table/index.tsx
```diff
@@ -1,35 +1,17 @@
 /** @jsxImportSource @emotion/react */
-import React, { useMemo, useState } from 'react';
+import React from 'react';
+import { uid } from 'react-uid';
```
coreyar: I think we should avoid using this library. In this case we are passing cells to it. If the data in the cells change this can cause rendering issues because the key will also change

src/components/v2/Table/index.tsx
```diff
@@ -41,92 +23,88 @@ export const Table = ({
   rowKeyIndex,
   className,
   gridTemplateColumns,
+  gridTemplateRowsMobile = '1fr',
+  isMobileView,
```
coreyar: Do we need this prop if we handle responsive styles with css?

src/components/v2/Table/index.stories.tsx
```diff
@@ -113,3 +118,19 @@ export const WithCustomColumnsWidth = () => (
...
+
+export const MobileViewMultipleRows = () => (
+  <Table
+    columns={columns}
+    data={rows}
+    title="Market Data"
+    rowKeyIndex={0}
+    gridTemplateColumns="1fr 150px"
+    gridTemplateRowsMobile="1fr 1fr"
+    isMobileView
```
coreyar: We take snapshots for a variety of view port sizes so we don't need a specific story to show responsiveness. Stories represent different states of the components

src/components/v2/Table/index.tsx
```diff
@@ -41,92 +23,88 @@ export const Table = ({
...
-            {rows.map(row => (
-              <TableRow
-                hover
-                key={row[rowKeyIndex].value.toString()}
-                onClick={e => rowOnClick && rowOnClick(e, row)}
-                css={styles.getTemplateColumns({ gridStyles })}
+      {isMobileView ? (
+        <>
+          {rows.map(([rowTitle, ...cells]) => (
+            <Paper key={uid(cells)} css={styles.tableWrapperMobile}>
```
coreyar: We can use `rowKeyIndex` here instead of `uid` with cells. This is nicer because we can pick a static column like the symbol or id of the assset to create the key and know that it won't change.

If we serialize all the cell data and some of it changes like collateral enabled, than the key will also change

src/components/v2/Table/index.tsx
```diff
@@ -41,92 +23,88 @@ export const Table = ({
...
+      {title && <h4 css={[styles.title, isMobileView && styles.titleMobile]}>{title}</h4>}
 
-          <TableBody>
-            {rows.map(row => (
-              <TableRow
-                hover
-                key={row[rowKeyIndex].value.toString()}
-                onClick={e => rowOnClick && rowOnClick(e, row)}
-                css={styles.getTemplateColumns({ gridStyles })}
+      {isMobileView ? (
```
coreyar: We could render both tables and hide one or the other with responsive css

coreyar: This could also be hardcoded because our responsive breakpoints for mobile are static

coreyar:
The toggle looks out of place in this snapshot - https://tf-venus-prod-vizreg.s3.amazonaws.com/57ffda6a9445ac6112d88e4afa0d9545b21ee8f7/index.html?id=new-chrome_iphone7_Components_Table_Mobile_View_Multiple_Rows.png#new-chrome_ipad_Components_Table_Mobile_View_Multiple_Rows.png

src/components/v2/Table/index.stories.tsx
```diff
@@ -47,7 +47,7 @@ function createData(asset: TokenId, apy: number, wallet: number, collateral: boo
       ),
     },
     {
-      key: apy.toString(),
+      key: 'APY',
```
therealemjy: Should be written in lowercase

src/components/v2/Table/index.stories.tsx
```diff
@@ -56,9 +56,9 @@ function createData(asset: TokenId, apy: number, wallet: number, collateral: boo
         </div>
       ),
     },
-    { key: wallet.toString(), value: wallet, render: () => `${wallet} ${asset}` },
+    { key: 'Wallet', value: wallet, render: () => `${wallet} ${asset}` },
```
therealemjy: Should be written in lowercase

src/components/v2/Table/index.stories.tsx
```diff
@@ -56,9 +56,9 @@ function createData(asset: TokenId, apy: number, wallet: number, collateral: boo
...
     {
-      key: collateral.toString(),
+      key: 'Collateral',
```
therealemjy: Should be written in lowercase

src/components/v2/Table/index.tsx
```diff
@@ -41,92 +23,88 @@ export const Table = ({
...
-                {row.map(({ key, render }: ITableRowProps) => {
-                  const cellContent = render();
-                  const cellTitle = typeof cellContent === 'string' ? cellContent : undefined;
+                {cells.map(cell => {
+                  const currentColumn = columns.find(column => column.key === cell.key);
                   return (
-                    <TableCell css={styles.cellWrapper} key={uid(key)} title={cellTitle}>
-                      {cellContent}
-                    </TableCell>
+                    <div key={uid(cell)} css={styles.cellMobile}>
```
therealemjy: As Corey mentioned, let's not use `uid`. Keys need to be based on an identifier that stays the same for a given cell between renders

src/pages/Dashboard/Markets/AssetCardMobile/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+        initialOrder={{
+          orderBy: 'apyEarned',
+          orderDirection: 'asc',
+        }}
+        rowOnClick={rowOnClick}
+        rowKeyIndex={0}
+      />
+    </Paper>
+  );
+};
```
therealemjy: This component should be removed since it's now handled from with `Table` directly

src/pages/Dashboard/Markets/BorrowMarket/BorrowMarketTable.tsx
```diff
@@ -75,7 +79,8 @@ const BorrowMarketTable: React.FC<IBorrowMarketTableProps> = ({
       }}
       rowKeyIndex={0}
       rowOnClick={rowOnClick}
-      gridTemplateColumns="120px 1fr 1fr 1fr"
+      gridTemplateColumns={isSmDown ? '1fr 1fr 1fr' : '120px 1fr 1fr 1fr'}
```
therealemjy: This should be imported from the styles, so we keep a unique source of truth

src/pages/Dashboard/Markets/BorrowMarket/BorrowingTable.tsx
```diff
@@ -103,7 +107,8 @@ const BorrowingTable: React.FC<IBorrowingUiProps> = ({
       }}
       rowKeyIndex={0}
       rowOnClick={rowOnClick}
-      gridTemplateColumns="120px 1fr 1fr 1fr"
+      gridTemplateColumns={isSmDown ? '1fr 1fr 1fr' : '120px 1fr 1fr 1fr'}
```
therealemjy: This should be imported from the styles, so we keep a unique source of truth

src/pages/Dashboard/Markets/SupplyMarket/SuppliedTable.tsx
```diff
@@ -86,7 +90,8 @@ export const SuppliedTable: React.FC<ISuppliedTableUiProps> = ({
       }}
       rowOnClick={rowOnClick}
       rowKeyIndex={0}
-      gridTemplateColumns="120px 1fr 1fr 1fr"
+      gridTemplateColumns={isSmDown ? '1fr 1fr 120px' : '120px 1fr 1fr 1fr'}
```
therealemjy: This should be imported from the styles, so we keep a unique source of truth

src/pages/Dashboard/Markets/SupplyMarket/SupplyMarketTable.tsx
```diff
@@ -86,7 +91,8 @@ export const SupplyMarketTable: React.FC<ISupplyMarketTableUiProps> = ({
       }}
       rowOnClick={rowOnClick}
       rowKeyIndex={0}
-      gridTemplateColumns="120px 1fr 1fr 1fr"
+      gridTemplateColumns={isSmDown ? '1fr 1fr 120px' : '120px 1fr 1fr 1fr'}
```
therealemjy: This should be imported from the styles, so we keep a unique source of truth

therealemjy:
@coreyar I agree with you about the positioning of the toggle, however the designs are inconsistent; sometimes the last column is aligned to the right while other times it is aligned to the left.
Screen Shot 2022-05-03 at 12 00 22
Screen Shot 2022-05-03 at 12 00 24

Handling this on a case by case basis will add overhead which I think is unnecessary at this stage.

src/components/v2/Table/index.tsx
therealemjy: Agreed, removed.

src/components/v2/Table/index.stories.tsx
therealemjy: Agreed, removed

src/components/v2/Table/index.tsx
therealemjy: In this case, because we rely on JavaScript anyway to set the grid layout of the table, then I think we using JS to define what layout to display

src/components/v2/Table/index.tsx
therealemjy: Nope, deleted.

src/components/v2/Table/index.tsx
therealemjy: We need to use CSS for that. I made the change

src/components/v2/Table/index.tsx
therealemjy: Agreed, fixed.

coreyar (approved):

PR #389, opened 2022-04-28
Supply withdraw updates

coreyar:
- [x] Show hypothetical borrow limit and daily earnings when user enters a value in input
 - [x] Convert values passed to BorrowBalanceAccountHealth into cents (currently passing dollars) 
 - [x] use XVS lifted state
 - [x] Set default when accessing account (This could be the default value in the context since)
 - [x] Include all assets in calculation of daily earnings
 - [x] Lift theme state above contexts
 - [x] Update validation flow so it doesn't block the user input but instead changes the color of the input to red when the value is higher than the maximum possible (the designs also show an error, but I think we should just ignore it - let me know if you think differently). I've had to integrate the same flow on the borrow/repay modal so have a look if it can help!
 - [x] Pass row onlick as a prop
 - [x] Close modal and display successful transaction modal when transaction is successfully sent

### DIFF
--- a/src/pages/App.tsx
+++ b/src/pages/App.tsx
@@ -40,11 +40,11 @@ const App = () => (
     <Web3Wrapper>
       <QueryClientProvider client={queryClient}>
         <Provider store={store}>
-          <AuthProvider>
+          <MuiThemeProvider>
             <RefreshContextProvider>
-              <VaiContextProvider>
-                <MarketContextProvider>
-                  <MuiThemeProvider>
+              <AuthProvider>
+                <VaiContextProvider>
+                  <MarketContextProvider>
                     <SuccessfulTransactionModalProvider>
                       <BrowserRouter>
                         <ToastContainer
@@ -77,11 +77,11 @@ const App = () => (
                         </Layout>
                       </BrowserRouter>
                     </SuccessfulTransactionModalProvider>
-                  </MuiThemeProvider>
-                </MarketContextProvider>
-              </VaiContextProvider>
+                  </MarketContextProvider>
+                </VaiContextProvider>
+              </AuthProvider>
             </RefreshContextProvider>
-          </AuthProvider>
+          </MuiThemeProvider>
         </Provider>
       </QueryClientProvider>
     </Web3Wrapper>
--- a/src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
+++ b/src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
@@ -5,7 +5,7 @@ import { FormikProps } from 'formik';
 import { Typography } from '@mui/material';
 import toast from 'components/Basic/Toast';
 import { FormValues } from 'containers/AmountForm/validationSchema';
-import { AmountForm, IAmountFormProps } from 'containers/AmountForm';
+import { AmountForm, IAmountFormProps, ErrorCode } from 'containers/AmountForm';
 import {
   TokenTextField,
   Delimiter,
@@ -45,6 +45,7 @@ export const SupplyWithdrawContent: React.FC<
   ISupplyWithdrawFormUiProps & FormikProps<FormValues>
 > = ({
   values,
+  errors,
   setFieldValue,
   asset,
   tokenInfo,
@@ -124,12 +125,14 @@ export const SupplyWithdrawContent: React.FC<
         tokenId={assetId as TokenId}
         value={amountString}
         onChange={amt => setFieldValue('amount', amt, true)}
-        max={maxInput.toFixed()}
+        disabled={isTransactionLoading}
         rightMaxButton={{
           label: t('supplyWithdraw.max').toUpperCase(),
           valueOnClick: maxInput.toFixed(),
         }}
         css={styles.input}
+        // Only display error state if amount is higher than borrow limit
+        hasError={errors.amount === ErrorCode.HIGHER_THAN_MAX}
       />
       <Typography
         component="div"
--- a/src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
+++ b/src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
@@ -14,6 +14,7 @@ import {
   BorrowBalanceAccountHealth,
   ValueUpdate,
 } from 'components';
+import PLACEHOLDER_KEY from 'constants/placeholderKey';
 import { SAFE_BORROW_LIMIT_PERCENTAGE } from 'config';
 import { useTranslation } from 'translation';
 import { Asset, TokenId } from 'types';
@@ -31,7 +32,7 @@ interface ISupplyWithdrawFormUiProps {
   tokenInfo: ILabeledInlineContentProps[];
   userTotalBorrowBalance: BigNumber;
   userTotalBorrowLimit: BigNumber;
-  dailyEarningsCents: BigNumber;
+  dailyEarningsCents: BigNumber | undefined;
   inputLabel: string;
   enabledButtonKey: string;
   disabledButtonKey: string;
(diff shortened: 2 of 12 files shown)


src/components/v2/ProgressBar/AccountHealth/index.tsx
```diff
@@ -49,7 +49,9 @@ export const AccountHealth: React.FC<IAccountHealthProps> = ({
 
   const safeBorrowLimitCents =
     typeof borrowLimitCents === 'number'
-      ? Math.floor((borrowLimitCents * safeBorrowLimitPercentage) / 100)
+      ? Math.floor(
+          ((hypotheticalBorrowBalanceCents || borrowLimitCents) * safeBorrowLimitPercentage) / 100,
```
therealemjy: This formula shouldn't be modified: the safe borrow limit doesn't change in function of the borrow balance.

src/pages/App.tsx
```diff
@@ -40,11 +40,11 @@ const App = () => (
...
         <Provider store={store}>
-          <AuthProvider>
+          <MuiThemeProvider>
             <RefreshContextProvider>
-              <VaiContextProvider>
-                <MarketContextProvider>
-                  <MuiThemeProvider>
+              <AuthProvider>
+                <VaiContextProvider>
+                  <MarketContextProvider>
```
therealemjy, line 47: Do we need to make the same change inside `renderComponent`?

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -77,23 +88,41 @@ export const SupplyWithdrawContent: React.FC<
       const temp = calculateNewBalance(amountInUsd);
       updateBorrowLimit = BigNumber.maximum(temp, 0);
     }
-
-    return [updateBorrowLimit];
+    const updateBorrowLimitCents = updateBorrowLimit?.multipliedBy(100);
+    return [updateBorrowLimitCents];
```
therealemjy: Small thing: since you're returning only one value, no need to wrap it with an array.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -77,23 +88,41 @@ export const SupplyWithdrawContent: React.FC<
...
   }, [amount, asset?.id, userTotalBorrowBalance, userTotalBorrowLimit]);
 
+  const dailyEarningsCents = useMemo(() => {
+    const hypotheticalAssets = [...assets];
+    // Modify asset with hypotheticalBalance
+    if (amount) {
+      const hypotheticalAsset = { ...asset, supplyBalance: asset.supplyBalance.plus(amount) };
```
therealemjy: This will cause an issue in the withdraw tab, as it will consider the `amount` as being additional supply while we're actually withdrawing from the supply.

I think we unfortunately have the same issue with the calculation of `hypotheticalBorrowLimitCents` above.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -105,7 +134,7 @@ export const SupplyWithdrawContent: React.FC<
           components={{
             White: <Typography component="span" variant="small1" css={styles.whiteLabel} />,
           }}
-          values={{ amount: format(maxInput), symbol: assetId?.toUpperCase() }}
+          values={{ amount: format(maxInput.minus(amount)), symbol: assetId?.toUpperCase() }}
```
therealemjy: This value shouldn't scale with `amount`, since it doesn't represent an hypothetical value but a value that's currently accurate. Otherwise it will tell the users their max withdraw amount or wallet balance is different from what it actually is.

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -182,18 +185,19 @@ export const SupplyWithdrawUi: React.FC<ISupplyWithdrawUiProps & ISupplyWithdraw
...
   const { t } = useTranslation();
   const { userTotalBorrowBalance, userTotalBorrowLimit } = useUserMarketInfo({
-    accountAddress: account?.address,
+    accountAddress,
   });
   const { data: vTokenBalance } = useGetVTokenBalance(
-    { account: account?.address || '', vTokenId: asset.id as VTokenId },
-    { enabled: !!account },
+    { account: accountAddress, vTokenId: asset.id as VTokenId },
+    { enabled: !!accountAddress },
```
therealemjy, line 197: Just to illustrate what I meant on my comment in your other PR where I was saying we should avoid assigning a default empty string to `accountAddress` from within the query hook directly.
Here you could remove the `enabled` option, and it wouldn't be obvious anymore that this query will first execute and pass an empty string to the contract call as the account address, then send another query once the account address becomes defined.

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -218,62 +222,96 @@ const SupplyWithdrawModal: React.FC<ISupplyWithdrawUiProps> = props => {
...
-    supply({
-      amount: getBigNumber(value)
+  const onSubmitSupply: IAmountFormProps['onSubmit'] = async value => {
+    try {
+      const withdrawableAmount = new BigNumber(value)
         .times(new BigNumber(10).pow(asset.decimals || 18))
-        .toString(10),
-    });
+        .toString(10);
+      const res = await supply({
```
therealemjy: If `supply` throws an error, two error toasts will display: one from the `catch` block here and one from the `onError` callback. Same thing with the `supply` function.

Also we should follow the same pattern we've agreed upon and handle all UI-related actions from the UI component.

src/components/v2/ProgressBar/AccountHealth/index.tsx
coreyar: Supplying additional coins would raise your borrow limit. That is what I am trying to show here. How the borrow limit would change with the hypothetically supplied coins.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
coreyar: I refactored `calculateNewBalance` to accept an initial value and am using it now here and for the supply Balance. `hypotheticalBorrowLimitCents` should be good because it is already using this function

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -164,10 +200,23 @@ interface ISupplyWithdrawFormProps extends ISupplyWithdrawFormUiProps {
...
+const SupplyWithdrawForm: React.FC<ISupplyWithdrawFormProps> = ({
+  onSubmit,
+  maxInput,
+  ...props
+}) => {
+  const onSubmitHandleError: IAmountFormProps['onSubmit'] = async (value: string) => {
+    try {
+      await onSubmit(value);
+    } catch (err) {
+      toast.error({ title: (err as Error).message });
```
therealemjy, line 217: So to follow our pattern, I think this should throw a `ToastError` which gets caught by `SupplyWithdrawContent` to render the toast.

therealemjy (approved):
One minor comment. We'll also need to update this PR once https://github.com/VenusProtocol/venus-protocol-interface/pull/387 is merged in

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
coreyar: This does follow our pattern in that the error is displayed from a UI component. This is not a container that contains side effects.

therealemjy: You're right 👍

therealemjy (approved):

PR #390, opened 2022-04-28
Add unique query keys for getVTokenBalancesAll

coreyar:
Add a unique key to the `getVTokenBalancesAll` queries so that they use a unique cache

### DIFF
--- a/src/clients/api/queries/useUserMarketInfo.ts
+++ b/src/clients/api/queries/useUserMarketInfo.ts
@@ -31,7 +31,7 @@ const useUserMarketInfo = ({
     { placeholderData: [], enabled: Boolean(accountAddress) },
   );
   const { data: vTokenBalancesAccount = [] } = useGetVTokenBalancesAll(
-    { account: accountAddress, vtAddresses },
+    { account: accountAddress || '', vtAddresses },
     { placeholderData: [], enabled: Boolean(accountAddress) },
   );
   const { data: vTokenBalancesTreasury = [] } = useGetVTokenBalancesAll(
--- a/src/clients/api/queries/getVTokenBalancesAll.ts
+++ b/src/clients/api/queries/getVTokenBalancesAll.ts
@@ -1,6 +1,6 @@
 export interface IGetVTokenBalancesAllInput {
   venusLensContract: $TSFixMe; // @TODO: use contract type (through Typechain?)
-  account: string | undefined | null;
+  account: string;
   vtAddresses: string[];
 }
 
--- a/src/clients/api/queries/useGetVTokenBalancesAll.ts
+++ b/src/clients/api/queries/useGetVTokenBalancesAll.ts
@@ -12,7 +12,7 @@ type Options = QueryObserverOptions<
   Error,
   IGetVTokenBalancesAllOutput,
   IGetVTokenBalancesAllOutput,
-  FunctionKey.GET_V_TOKEN_BALANCES_ALL
+  [FunctionKey.GET_V_TOKEN_BALANCES_ALL, string]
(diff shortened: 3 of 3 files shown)


src/clients/api/queries/useUserMarketInfo.ts
```diff
@@ -16,7 +16,7 @@ import {
 } from 'clients/api';
 
 const useUserMarketInfo = ({
-  accountAddress,
+  accountAddress = '',
```
therealemjy: I think we should avoid this pattern.
I've seen that we're starting to do that on hooks, but it can lead to issues that are hard to debug. Assigning an empty string if the address is `undefined` means devs won't realize they are potentially passing an `undefined` parameter to the hook. However, the query needs an account address to be passed, and passing an empty string will cause errors from the contracts.
Also, beyond generating an error it could waste a request, since it would first make the request with an empty string, then once the account address is defined it would execute the query again; it didn't need to execute the first time.

Requiring `accountAddress` not to be `undefined` means that if a developer wants to use the hook in situations where the address they want to pass might be `undefined`, then they have to use the `enabled` option of the query: it makes it clear to someone that they shouldn't pass an `undefined` value and the only way to do that is to use that option.

Screen Shot 2022-04-28 at 9 29 52 PM

coreyar: I made this change to avoid duplicated the logic you have in the screenshot where we fall back to an empty string in the function call. I don't see a significant difference to falling back in the function call over when accessing the variable. 

Our hooks currently are typed two ways, with `accountAddress` as a `string` and `accountAddress`  as `string | undefined`. I've been moving towards requiring `accountAddress` to be a string because we previously discussed that and because the contract shouldn't be called with it as `undefined`, because the contract can't be called with the address as `undefined`.

Both an empty string and `undefined` are cast to false, thus preventing any unwanted calls. On these requests we are doing [just that](https://github.com/VenusProtocol/venus-protocol-interface/pull/390/files#diff-356fb756083fc74e410b7bcd06e9bf0f89305fa1896889681c5739f6ecf0d8d5R33-R40)

I am fine either way and can update if you feel strongly we should always fall back in the function call. It sounds like you want to see the typescript error when writing the function call and fallback at the call site and have that be the reminder to set enabled? I think a better solution would be to type the `enabled` property as required on these hooks because there is no guarantee that you will also set the enabled option when falling back in the function call

therealemjy: I get your point about avoiding duplication, but I think it's necessary in this case to make things clearer.

You mention in your second paragraph that we're moving towards requiring `accountAddress` to be a string when it's passed to an operation hook: I agree with that approach, but then we shouldn't assign an empty string as default to it as otherwise we allow `accountAddress` to be passed as `undefined`.

I'm feeling strongly about this, because I believe it's easily prone to errors, especially for new comers.
_It sounds like you want to see the typescript error when writing the function call and fallback at the call site and have that be the reminder to set enabled?_
Exactly that yes. TS lets them know the address can not be `undefined` and so passing an empty string to it should make it obvious that unless you pass the `enabled` option too then a call will be made using an empty string when `accountAddress` is `undefined`. The developer using the hook is in full control of the behaviour of it, so there shouldn't be any surprise.

I wouldn't make `enabled` mandatory, because there might be cases where we know the user is authenticated, in which case we know the account address isn't `undefined`. It is indeed not guaranteed that a developer will pass the `enabled` option along side assigning an empty string if `accountAddress` is `undefined`, but it is then much easier to spot in the code.

therealemjy (approved):

PR #392, opened 2022-04-28
Wrap Repay and Mint Vai with Enable Token Component

coreyar:
The PR wrap the the Repay and Mint VAI component in the Enable Token Component so that the user will only see the form if they've enable the token.

If they haven't enable the token they will be prompted to enable it

### DIFF
--- a/src/pages/Dashboard/MintRepayVai/MintVai/index.spec.tsx
+++ b/src/pages/Dashboard/MintRepayVai/MintVai/index.spec.tsx
@@ -10,7 +10,6 @@ import { AuthContext } from 'context/AuthContext';
 import { VaiContext } from 'context/VaiContext';
 import renderComponent from 'testUtils/renderComponent';
 import { assetData } from '__mocks__/models/asset';
-import en from 'translation/translations/en.json';
 import RepayVai from '.';
 
 jest.mock('clients/api');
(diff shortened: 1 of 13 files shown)


src/pages/Dashboard/MintRepayVai/MintVai/index.spec.tsx
```diff
@@ -108,5 +118,32 @@ describe('pages/Dashboard/MintRepayVai/MintVai', () => {
     });
   });
 
+  it('asks the user to enable token if not enabled', async () => {
```
therealemjy: I saw you added this test for the supply/withdraw modal too. I think we should add tests to the `EnableToken` component instead, so that we don't have to duplicate them on every component that uses it.

src/pages/Dashboard/MintRepayVai/RepayVai/index.spec.tsx
```diff
@@ -101,5 +112,32 @@ describe('pages/Dashboard/MintRepayVai/RepayVai', () => {
     });
   });
 
+  it('asks the user to enable token if not enabled', async () => {
```
therealemjy: Same thing here about testing `EnableToken` instead

therealemjy:
I found two issues with this:
- The XVS icon doesn't seem to display in the `EnableToken` component
Screen Shot 2022-05-03 at 17 37 40
- The `EnableToken` component displays when the user is logged out, however in the designs it's the interface with the form and a button indicating "Connect wallet" that should display. To be honest I think that's an inconsistent UX anyway, so I'd propose instead we do like we do for the borrow/supply modals and use the `ConnectWallet` component to handle this case. What do you think?

therealemjy (approved):

PR #398, opened 2022-05-03
Refactor Formik setup

coreyar:
This takes advantage of the Formik context to create components configured to use in the forms. The TokenTextField and SubmitButton are specifically configured so that no extra configuration is needed to make them compatible with Formik.

The advantage of setting things up like this is that it enforces all our our forms and fields have the same behavior. If we want to have a custom form we still can but it would be require a conscious decision not to use our configured Formik components.

### DIFF
--- a/src/components/v2/Form/FormikSubmitButton.tsx
+++ b/src/components/v2/Form/FormikSubmitButton.tsx
@@ -1,0 +1,24 @@
+import React from 'react';
+import { useFormikContext } from 'formik';
+import { Button, IButtonProps } from '../Button';
+
+interface IFormikSubmitButtonProps extends IButtonProps {
+  disabledLabel?: string;
+  enabledLabel: string;
+}
+
+export const FormikSubmitButton = ({
+  disabledLabel,
+  enabledLabel,
+  variant = 'primary',
+  disabled,
+  ...rest
+}: Omit<IFormikSubmitButtonProps, 'type'>) => {
+  const { isValid, dirty } = useFormikContext();
+  const showDisableLabel = !isValid || !dirty; // loading disabled
+  return (
+    <Button variant={variant} {...rest} disabled={disabled || showDisableLabel} type="submit">
+      {(showDisableLabel && disabledLabel) || enabledLabel}
+    </Button>
+  );
+};
--- a/src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
+++ b/src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
@@ -3,7 +3,6 @@ import React, { useContext } from 'react';
 import BigNumber from 'bignumber.js';
 import type { TransactionReceipt } from 'web3-core';
 
-import { TokenId } from 'types';
 import { AuthContext } from 'context/AuthContext';
 import { getToken } from 'utilities';
 import useSuccessfulTransactionModal from 'hooks/useSuccessfulTransactionModal';
@@ -14,19 +13,22 @@ import {
   EnableToken,
   IconName,
   ILabeledInlineContentProps,
-  SecondaryButton,
+  FormikSubmitButton,
   LabeledInlineContent,
-  TokenTextField,
+  FormikTokenTextField,
   ConnectWallet,
 } from 'components';
 import { useVaiUser } from 'hooks/useVaiUser';
 import { useGetVaiTreasuryPercentage, useMintVai } from 'clients/api';
 import toast from 'components/Basic/Toast';
 import { useTranslation } from 'translation';
+import { TokenId } from 'types';
+import PLACEHOLDER_KEY from 'constants/placeholderKey';
 import useConvertToReadableCoinString from '../useConvertToReadableCoinString';
 import { VAI_ID } from '../constants';
-import getReadableFeeVai from './getReadableFeeVai';
 import { useStyles } from '../styles';
+import getReadableFeeVai from './getReadableFeeVai';
+import MintFee from './MintFee';
 
 export interface IMintVaiUiProps {
   disabled: boolean;
@@ -63,21 +65,6 @@ export const MintVaiUi: React.FC<IMintVaiUiProps> = ({
 
   const hasMintableVai = limitWei?.isGreaterThan(0) || false;
 
-  const getReadableMintFee = React.useCallback(
-    (valueWei: string) => {
-      if (!mintFeePercentage) {
-        return '-';
-      }
-
-      const readableFeeVai = getReadableFeeVai({
-        valueWei: new BigNumber(valueWei || 0),
-        mintFeePercentage,
-      });
-      return `${readableFeeVai} (${mintFeePercentage}%)`;
-    },
-    [mintFeePercentage],
-  );
-
   const onSubmit: IAmountFormProps['onSubmit'] = async amountTokens => {
     const amountWei = convertCoinsToWei({
       value: new BigNumber(amountTokens),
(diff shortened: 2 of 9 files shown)


src/components/v2/Button/FormikSubmitButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+export const FormikSubmitButton = ({
+  disabledLabel,
+  enabledLabel,
+  variant = 'primary',
+  disabled = false,
+  loading = false,
+  ...rest
+}: IFormikSubmitButtonProps) => {
+  const { isValid, dirty } = useFormikContext();
+  const disable = !isValid || !dirty;
```
therealemjy: Having `disable` and `disabled` is confusing imo, perhaps it could be renamed?

src/components/v2/Button/FormikSubmitButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <PrimaryButton
+      variant={variant}
+      {...rest}
```
therealemjy: I suggest spreading these last so the default props can be overwritten

src/components/v2/Button/FormikSubmitButton.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { useFormikContext } from 'formik';
+import { PrimaryButton, IButtonProps } from '.';
+
+interface IFormikSubmitButtonProps extends IButtonProps {
+  disabled?: boolean;
+  loading: boolean;
```
therealemjy: Both these props are included in `IButtonProps`, so you can remove them

src/components/v2/Button/FormikSubmitButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+  disabledLabel?: string;
+  enabledLabel: string;
+}
+
+export const FormikSubmitButton = ({
+  disabledLabel,
+  enabledLabel,
+  variant = 'primary',
+  disabled = false,
+  loading = false,
```
therealemjy: Since `loading` is already defined in `IButtonProps` and a default value is assigned to it in the `Button` component, you can remove it here.

src/components/v2/Button/FormikSubmitButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+}: IFormikSubmitButtonProps) => {
+  const { isValid, dirty } = useFormikContext();
+  const disable = !isValid || !dirty;
+  return (
+    <PrimaryButton
+      variant={variant}
+      {...rest}
+      disabled={disabled || disable}
+      type="submit"
+      loading={loading}
```
therealemjy: Following my previous comments, you can then also remove this line since if a a loading prop is passed it would be spread through the `rest` variable.

src/components/v2/TokenTextField/FormikTokenTextField.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <TokenTextField
+      value={value}
+      onChange={onChange}
+      onBlur={onBlur}
+      hasError={!!(error && errorCodes.includes(error))}
+      {...rest}
+    />
+  );
+};
```
therealemjy: I'd move this file inside a separate folder (`form` would work as you suggested yesterday) so it's separate from `TokenTextField` which will be appreciated if get rid of Formik in the future.

src/components/v2/TokenTextField/FormikTokenTextField.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { useField } from 'formik';
+import { TokenTextField, ITokenTextFieldProps } from '.';
+
+interface IFormikTokenTextField extends Omit<ITokenTextFieldProps, 'name' | 'onChange' | 'value'> {
+  name: string;
+  errorCodes?: string[];
+}
+
+export const FormikTokenTextField = ({ name, errorCodes = [], ...rest }: IFormikTokenTextField) => {
+  const [{ value, onBlur }, { error }, { setValue }] = useField(name); // eslint-disable-line  @typescript-eslint/no-unused-vars
```
therealemjy: It looks like we can get rid of the silencer comment, I ran the linter on your branch without it and it didn't return any error
```suggestion
  const [{ value, onBlur }, { error }, { setValue }] = useField(name);
```

src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
```diff
@@ -27,6 +28,35 @@ export interface IMintVaiUiProps {
...
+    <LabeledInlineContent
+      css={styles.getRow({ isLast: true })}
+      iconName="fee"
+      label={t('mintRepayVai.mintVai.mintFeeLabel')}
+    >
+      {getReadableMintFee(value)}
+    </LabeledInlineContent>
+  );
+};
+
```
therealemjy: If we're making this its own component, then it should get its own file (inside the `MintVai` folder)

src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
```diff
@@ -27,6 +28,35 @@ export interface IMintVaiUiProps {
...
 }
 
+const MintFee = ({ mintFeePercentage }: Pick<IMintVaiUiProps, 'mintFeePercentage'>) => {
+  const [{ value }] = useField('amount');
+  const styles = useStyles();
+  const { t } = useTranslation();
+  const getReadableMintFee = React.useCallback(
+    (valueWei: string) => {
+      if (!mintFeePercentage) {
+        return '-';
```
therealemjy: I know that's my code, but now that we have a constant for this character can you update this line please?
```suggestion
        return PLACEHOLDER_KEY;
```

src/components/v2/TokenTextField/FormikTokenTextField.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { useField } from 'formik';
+import { TokenTextField, ITokenTextFieldProps } from '.';
+
+interface IFormikTokenTextField extends Omit<ITokenTextFieldProps, 'name' | 'onChange' | 'value'> {
+  name: string;
+  errorCodes?: string[];
```
therealemjy: I think the name `errorCodes` is confusing in this context, as I'd expect it to correspond to the codes of all the current errors for this component.
Perhaps `displayableErrorCodes` or something like that?

src/components/v2/Button/FormikSubmitButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+    >
+      {(disable && disabledLabel) || enabledLabel}
+    </PrimaryButton>
+  );
+};
```
therealemjy: Same thing as for `FormikTokenTextField`, I'd move this component to a separate `form` folder

src/components/v2/Button/FormikSubmitButton.tsx
coreyar: We can do that but we lose some benefit of having this component which is to strictly enforce our form patterns. Breaking our form UX pattern should be a conscious decision that requires using the primitive button and configuring it or creating a new Form component. Those changes are stronger flags that we are creating a unique form than overriding props.

therealemjy: I get your intention, but in this case we should omit the props that are assigned by this component so they can't be passed. Otherwise we'll end up with a broken implementation. Take this example:
```tsx
<FormikSubmitButton disabled={true} />
```

Here I'd expect the button to always be disabled, however that won't be the case since although it accepts a `disabled` prop it gets overwritten within the component itself.

src/components/v2/Form/FormikSubmitButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+  enabledLabel,
+  variant = 'primary',
+  disabled,
+  ...rest
+}: Omit<IFormikSubmitButtonProps, 'type'>) => {
+  const { isValid, dirty } = useFormikContext();
+  const showDisableLabel = !isValid || !dirty; // loading disabled
+  return (
+    <PrimaryButton
+      variant={variant}
```
therealemjy: Just saw that sorry. Since we're passing the variant anyway, I guess we should use the `Button` component directly.

therealemjy (approved):
Just one small thing

therealemjy (approved):

PR #399, opened 2022-05-03
Add labeled progress bar

coreyar:
This branch was merged but it looks like it got dropped in a rebase


therealemjy (approved):
Good catch 👍

PR #400, opened 2022-05-03
Convert VRT UI

coreyar:
This PR adds the UI for the VRT Screen but not functionality


src/translation/translations/en.json
```diff
@@ -79,6 +79,23 @@
...
+    "dailyLimit": "Daily limit",
+    "progressBar": "daily limit progress bar",
+    "remainingTime": "<Icon/> Remaining time:\u00a0<White>{{time}}</White>",
+    "usedOverTotalVrt": "{{used}}/{{total}} VRT",
+    "vrtEqualsXvs": "1 VRT = {{xvsToVrtRate}} XVS",
+    "withdraw": "Withdraw",
+    "withdrawableAmount": "Withdrawable amount",
+    "withdrawXvs": "Withdraw XVS",
+    "xvsAVailable": "Current XVS Available",
+    "youWillReceive": "You will recieve"
```
therealemjy:
```suggestion
    "youWillReceive": "You will receive"
```

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+  vrtLimitUsed,
+  vrtLimit,
+  vestingTime,
+}) => {
+  const styles = useStyles();
+  const { t, Trans } = useTranslation();
+  const readableXvsAvailable = useConvertToReadableCoinString({
+    valueWei: xvsTotal,
+    tokenId: XVS_ID,
+  });
```
therealemjy: Since this value doesn't change often, it can be memoized

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { format } from 'utilities/common';
+import { ConvertVrtForm } from '../Form';
+import { VRT_ID, XVS_ID } from '../constants';
+import { useStyles } from '../styles';
+
+interface ConvertProps {
+  xvsTotal: BigNumber;
+  xvsToVrtRate: BigNumber;
+  vrtLimitUsed: number;
+  vrtLimit: number;
```
therealemjy: `vrtLimitUsed` and `vrtLimit ` need to use `BigNumber`, JavaScript numbers don't handle big values well.
Also, to be consistent with what we've done so far I'd add the units to each value (for example in this case I don't know if vrtLimitUsed is a percentage or a value in wei or VRT).

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+  vestingTime: number;
```
therealemjy: I'd add the unit it is expressed in (I'm guessing ms?)

src/pages/ConvertVrt/Form/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+      initialValues={{
+        vrt: initialAmount,
+        xvs: initialAmount,
+      }}
+      onSubmit={handleSubmit}
+      validationSchema={getValidationSchema({ maxXvs, maxVrt })}
+      validateOnMount
+      validateOnChange
+    >
+      {children}
```
therealemjy: I'd suggest passing the formik props tot the `children`, you'll likely need them when you do the wire up work

src/pages/ConvertVrt/Form/validationSchema.ts
```diff
@@ -0,0 +1,32 @@
...
+  yup.object({
+    vrt: yup
+      .string()
+      .required()
+      .test('isPositive', ErrorCode.NOT_POSITIVE, value => !!value && +value > 0)
+      .test(
+        'isHigherThanMax',
+        ErrorCode.HIGHER_THAN_MAX,
+        value => !value || !maxVrt || +value <= +maxVrt,
+      ),
```
therealemjy: Seeing we use this rule twice here and also in `AmountForm`, perhaps we could put it in a separate file and import in the forms?

src/pages/ConvertVrt/Withdraw/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  xvsTotal: BigNumber;
+}
+
+const Withdraw: React.FC<WithdrawProps> = ({ xvsTotal }) => {
+  const { t } = useTranslation();
+  const styles = useStyles();
+  const readableXvsAvailable = useConvertToReadableCoinString({
+    valueWei: xvsTotal,
+    tokenId: XVS_ID,
+  });
```
therealemjy: This can be memoized since it won't change often

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+                enabledLabel={t('convertVrt.convertVrttoXvs')}
+              />
+              <Typography css={styles.remainingTime}>
+                <Trans
+                  i18nKey="convertVrt.remainingTime"
+                  components={{
+                    Icon: <Icon name="countdown" />,
+                    White: <span css={styles.whiteLabel} />,
+                  }}
+                  values={{ time: vestingTime }}
```
therealemjy: `vestingTime` will need to be formatted to a readable value here. The design team hasn't specific rules for how this time should be converted, so what do you think about using these ones:
- if time is longer than a day, display whole days only. .e.g: `2 days`
- if time is longer than an hour but shorter than a day, display hours and minutes only. e.g.: `10 hours 30 minutes`
- otherwise display whole minutes only. e.g.: `46 minutes`

src/pages/ConvertVrt/Form/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import getValidationSchema, { FormValues } from './validationSchema';
+
+export * from './validationSchema';
+
+export interface IAmountFormProps
+  extends Omit<FormikConfig<FormValues>, 'onSubmit' | 'initialValues'> {
+  onSubmit: (values: { vrt: string; xvs: string }) => Promise<void> | void;
+  children: (formProps: FormikProps<FormValues>) => React.ReactNode;
+  initialAmount?: string;
```
therealemjy: I'd get rid of this prop and use `initialValues` directly since there's two inputs in this form

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+                </Typography>
+              </div>
+              <div css={styles.progressBar}>
+                <LabeledProgressBar
+                  greyLeftText={t('convertVrt.dailyLimit')}
+                  whiteRightText={t('convertVrt.usedOverTotalVrt', {
+                    used: vrtLimitUsed,
+                    total: vrtLimit,
+                  })}
+                  value={vrtLimitUsed / vrtLimit}
```
therealemjy: I think that's incorrect. If `vrtLimitUsed` is a value in VRT, them the value here would be:
```suggestion
                  value={vrtLimitUsed * 100 / vrtLimit}
```

If `vrtLimitUsed` is the percentage already, then it should be passed as is:
```suggestion
                  value={vrtLimitUsed}
```

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+          {() => (
+            <>
+              <div css={styles.inputSection}>
+                <Typography variant="small2" css={styles.inputLabel}>
+                  {t('convertVrt.convertVrt')}
+                </Typography>
+                <FormikTokenTextField
+                  tokenId={VRT_ID}
+                  name="vrt"
+                  max={xvsTotal.toFixed()}
```
therealemjy: I think this line should be removed and instead we should make the border of the input red if the value is higher than the max, because users are also able to change the value of the XVS input which means they might enter a value that's higher than the max allowed

therealemjy (commented):
A few small things:
- `useConvertToReadableCoinString` should be moved into the `hooks` folder
- The font weight of the labels above the inputs is incorrect
- although the VRT balance displayed is 1000, I'm able to enter an amount higher than that in the VRT input
- maybe you're planning on adding this logic when you do the wire up, but changing the value of one input should update the value of the other
- the submit button on the withdraw tab adds a bottom margin that should be removed (in the designs the spacing between the submit button and the border of the component is 40px, here because of the bottom margin it's 72px)

coreyar:
> * although the VRT balance displayed is 1000, I'm able to enter an amount higher than that in the VRT input
I was planning on working on the interactions when wiring it up

> * maybe you're planning on adding this logic when you do the wire up, but changing the value of one input should update the value of the other

I asked about this in the design channel. It seems this was not the original intention and the designs will be updated to make this clearer.

src/pages/ConvertVrt/Convert/index.tsx
coreyar: I'll check on the units when I wire them up

src/pages/ConvertVrt/Form/index.tsx
coreyar: We don't need them anymore if we use the formik hooks like in the previous PR

src/pages/ConvertVrt/Form/validationSchema.ts
coreyar: I still need to check this when wiring up the functionality. I can do this in this PR but was going to do it in the next.

src/pages/ConvertVrt/Form/index.tsx
therealemjy: In some instances you'll need access to the Formik props without using a hook to read the context, for example to create a function that uses the values, like we do with `calculateDailyEarnings` on the Borrow/Repay form.

coreyar: The props are passed automatically [here](https://github.com/VenusProtocol/venus-protocol-interface/pull/400/files#diff-e7f075c5ecae4dbbe05afc73300eeafbcbb3e9ef4504e8c1f42528da1c25a3aeR53) 

No change is needed

src/pages/ConvertVrt/Form/validationSchema.ts
coreyar: Updated

therealemjy (commented):
I know we don't have designs for this and so you went with your gut feeling, but what do you think about hiding "0 XVS Current XVS available" title on this screen? It will always display a 0 value since no wallet is connected
Screen Shot 2022-05-06 at 10 21 26

Update on this: Pavel updated the designs to add the "Connect wallet" step and he updated the "Enable token" step too.
I also realized that both tokens need to be enabled, I'm not sure how the design team wants to handle that. I've asked Pavel.

I spotted a few CSS issues:
- the spacing between the content and the borders is too high
Screen Shot 2022-05-06 at 10 27 22
- the clock icon isn't vertically aligned
Screen Shot 2022-05-06 at 10 27 27

Also, the time displayed seems off.

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+    }
+  | {
+      relativeTimeTranslationKey: 'convertVrt.remainingTimeMissing';
+      realtiveTimeFormatValues: { count: string };
+    };
+// Don't remove this comment, it allows for translation keys set by this function to be collected
+// t('convertVrt.remainingTimeDays_one') t('convertVrt.remainingTimeDays_other') t('convertVrt.remainingTimeHours_one') t('convertVrt.remainingTimeHours_other')
+// t('convertVrt.remainingTimeHoursAndMinutes') t('convertVrt.remainingTimeMinutes') t('convertVrt.remainingTimeMinutes_other') t('convertVrt.minutes_one')
+// t('convertVrt.minutes_other') t('convertVrt.hours_one') t('convertVrt.hours_other') t('convertVrt.days_one') t('convertVrt.days_other')
+const formatI18nextRelativetimeValues = (
```
therealemjy: Let's move this into its own file so it can be reused as a global util.

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+  vrtConversionEndTime: string | undefined,
+): FormatI18nextRelativetimeValuesReturn => {
```
therealemjy: Why not just take a `Date` param and return the formatted time as a string?

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+      realtiveTimeFormatValues: {
+        hours: Math.floor(vestingTimeRemainingMs / HOUR_IN_MILISECONDS),
+        minutes: Math.floor(
+          ((vestingTimeRemainingMs % HOUR_IN_MILISECONDS) / HOUR_IN_MILISECONDS) * 60,
+        ),
+      },
+      relativeTimeTranslationKey: 'convertVrt.remainingTimeHoursAndMinutes',
+    };
+  }
+  return relativeTimeValues;
```
therealemjy: I saw you marked my previous comment on how to format the time as resolved, but this doesn't follow the rules I wrote:
- if time is longer than a day, display whole days only. .e.g: 2 days
- if time is longer than an hour but shorter than a day, display hours and minutes only. e.g.: 10 hours 30 minutes
- otherwise display whole minutes only. e.g.: 46 minutes

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+                        i18nKey="convertVrt.balance"
+                        components={{
+                          White: <span css={styles.whiteLabel} />,
+                        }}
+                        values={{ amount: format(new BigNumber(vrtLimitWei)) }}
+                      />
+                    }
+                    rightMaxButton={{
+                      label: t('convertVrt.max').toUpperCase(),
+                      valueOnClick: xvsTotalWei.toFixed(),
```
therealemjy, line 85: This needs to be converted to whole tokens. I also think it's the wrong value since this input is for VRT not XVS

src/components/VrtConversion/Convert.tsx
```diff
@@ -102,7 +102,7 @@ export default ({
       </div>
       {/* recieve section */}
       <div className="recieve-xvs">
-        <div className="input-title">You will recieve</div>
+        <div className="input-title">You will receive</div>
```
therealemjy, line 105: Nice catch 👍

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+  vrtLimitUsedWei,
+  vrtLimitWei,
+  vrtConversionEndTime,
+}) => {
+  const styles = useStyles();
+  const { t, Trans } = useTranslation();
+  const readableXvsAvailable = useConvertToReadableCoinString({
+    valueWei: xvsTotalWei,
+    tokenId: XVS_ID,
+  });
```
therealemjy, line 44: This can be memoized since it won't change often

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+      </section>
+      <ConnectWallet message={t('convertVrt.connectWalletToConvertVrtToXvs')}>
+        <AmountForm onSubmit={noop} maxAmount={vrtLimitWei.toFixed()}>
+          {({ values }) => {
+            const xvsValue = useMemo(() => {
+              if (values.amount) {
+                return new BigNumber(values.amount).times(xvsToVrtRate).toFixed();
+              }
+              return '';
+            }, [values.amount, xvsToVrtRate]);
```
therealemjy: There's no point in memoizing this value here, since `values.amount` will change very often you'll actually lengthen the initial rendering time without making subsequent renders faster (given most of the re-renders will come from `values.amount` changing).
You could create a `convertToXvsTokens` (or any name you see fit) function outside of the `return` block and that uses `useCallback` to return a memoized function that accepts the amount as parameter:
```typescript
const convertToXvsTokens = React.useCallback((xvsAmountTokens: string | undefined) =>
  values.amount ?
    new BigNumber(values.amount).times(xvsToVrtRate).dp(VRT_DECIMALS).toFixed() : '', [xvsToVrtRate]);
```

Please note we also should apply format the final value so it doesn't contain more decimals than XVS has.

src/pages/ConvertVrt/styles.ts
```diff
@@ -0,0 +1,73 @@
...
+    submitButton: css`
+      margin-bottom: ${theme.spacing(4)};
+    `,
+    remainingTime: css`
+      width: 100%;
+      display: inline-flex;
+      justify-content: center;
+      svg {
+        margin-right: ${theme.spacing(2)};
+      }
```
therealemjy: I'd avoid doing that and instead apply CSS to the icon directly by passing a `css` prop to it (less prone to confusion if in the future we add another `svg` inside this block).

src/components/v2/TokenTextField/FormikTokenTextField.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <TokenTextField
+      value={value}
+      onChange={onChange}
+      onBlur={onBlur}
+      hasError={!!(error && errorCodes.includes(error))}
+      {...rest}
+    />
+  );
+};
```
therealemjy: This file needs to be removed

src/components/v2/Button/FormikSubmitButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+      variant={variant}
+      {...rest}
+      disabled={disabled || disable}
+      type="submit"
+      loading={loading}
+    >
+      {(disable && disabledLabel) || enabledLabel}
+    </PrimaryButton>
+  );
+};
```
therealemjy: This file needs to be removed

src/pages/ConvertVrt/Convert/index.tsx
coreyar: It is memoized

src/pages/ConvertVrt/Convert/index.tsx
coreyar: It does follow the rules but the math is a bit off for example I assumed milliseconds but it is actually returned as a timestamp. This has been updated in the [PR](https://github.com/VenusProtocol/venus-protocol-interface/pull/422/files#diff-e7f075c5ecae4dbbe05afc73300eeafbcbb3e9ef4504e8c1f42528da1c25a3aeR57) where it wires everything up. 

We can also take those changes here.

src/pages/ConvertVrt/Convert/index.tsx
coreyar: We can do that I started with a string because that is it is returned. I'd like to work through this function on [PR 422](https://github.com/VenusProtocol/venus-protocol-interface/pull/422) because that is where I implemented it based on the unix time stamp and not a relative millisecond time like I originally assumed. In this PR it looks like it picked up some changes but not all.

src/pages/ConvertVrt/Convert/index.tsx
coreyar: This is updated in PR 422

src/components/v2/TokenTextField/FormikTokenTextField.tsx
coreyar: We are using these files to have base interactive form components that conform to our UX form pattern. The advantage of this is that if we implement a form that breaks our UX pattern it will be obvious because we'd have to either implement an input and buttons directly or create a new component

therealemjy (approved):

PR #401, opened 2022-05-04
Fix warnings printed in console

therealemjy:
- disable `@typescript-eslint/explicit-module-boundary-types` eslint rule. We infer most of the return types of our functions already, this rule doesn't do anything for us
- remove unnecessary type assertions
- replace `:first-child` instances with `:first-of-type`. The explanation for this comes from [this post](https://stackoverflow.com/questions/57239807/the-pseudo-class-first-child-is-potentially-unsafe-when-doing-server-side-ren):
> This comes from [@emotion](https://emotion.sh/docs/introduction) through storybook - since emotion (creates CSS through JavaScript) sometimes injects a <style> element in the beginning - which would make that the first child and breaks your CSS. first-of-type [is supported in most browsers these days](https://caniuse.com/?search=first-of-type) and can be used in many cases alternatively.
- Update `ProgressBar` so only recognized props are spread on tags. This still doesn't fix a warning we're getting:
Screen Shot 2022-05-04 at 08 54 40
But since it seems to come from MUI directly, we can't do anything about it other than updating MUI if they've released a fix (I haven't looked into it, since updating the library could break other elements of the app).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -63,6 +63,7 @@ module.exports = {
         },
       },
     ],
+    '@typescript-eslint/explicit-module-boundary-types': 0,
     'react/jsx-wrap-multilines': 0,
     'generator-star-spacing': 0,
   },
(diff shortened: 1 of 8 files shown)


coreyar:
Nice work! Thanks!

coreyar (approved):

PR #402, opened 2022-05-04
Feature: add notice component

baomastr:
https://app.clickup.com/t/296p8ze

### DIFF
--- a/src/components/v2/Notice/index.tsx
+++ b/src/components/v2/Notice/index.tsx
@@ -1,0 +1,53 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+import Paper from '@mui/material/Paper';
+import Typography from '@mui/material/Typography';
+import { Icon } from '../Icon';
+import { NoticeVariant } from './types';
+import { useStyles } from './styles';
+
+interface INoticeProps {
+  className?: string;
+  title?: string;
+  description: string;
+  variant?: NoticeVariant;
+}
+
+const getNoticeIconName = (variant: NoticeVariant): 'info' | 'notice' | 'checkInline' => {
+  switch (variant) {
+    case 'error':
+      return 'notice';
+    case 'success':
+      return 'checkInline';
+    default:
+    case 'info':
+      return 'info';
+  }
+};
+
+export const Notice = ({ className, title, description, variant = 'info' }: INoticeProps) => {
+  const styles = useStyles();
+
+  return (
+    <Paper css={[styles.root, styles.getRootStyles({ variant })]} className={className}>
+      <Icon
+        css={[styles.icon, styles.getIconStyles({ variant })]}
+        name={getNoticeIconName(variant)}
+      />
+      <div css={styles.content}>
+        {title && (
+          <Typography variant="small2" color="text.primary" css={styles.title}>
+            {title}
+          </Typography>
+        )}
+        <Typography variant="small2" color="text.primary">
+          {description}
+        </Typography>
+      </div>
+    </Paper>
+  );
+};
+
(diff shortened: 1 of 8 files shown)


src/components/v2/Notice/styles.ts
```diff
@@ -0,0 +1,49 @@
...
+    `,
+    icon_primary: css`
+      color: ${theme.palette.interactive.primary};
+    `,
+    icon_error: css`
+      color: ${theme.palette.interactive.error};
+    `,
+    icon_success: css`
+      color: ${theme.palette.interactive.success};
+    `,
```
therealemjy: Let's use the pattern we've followed so far and use a function for these

src/components/v2/Notice/styles.ts
```diff
@@ -0,0 +1,49 @@
...
+    content: css`
+      display: flex;
+      flex-direction: column;
+      padding-left: ${theme.spacing(8)};
+    `,
+    text: css`
+      color: ${theme.palette.text.primary};
+      font-size: ${theme.spacing(3.5)};
```
therealemjy: We can use a variant instead of assigning a font size like this

src/components/v2/Notice/index.tsx
```diff
@@ -0,0 +1,41 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+import Paper from '@mui/material/Paper';
+import Typography from '@mui/material/Typography';
+import { Icon } from 'components';
+
+import { useNoticeStyles as useStyles } from './styles';
+
+type NoticeVariant = 'primary' | 'error' | 'success';
```
therealemjy: To be consistent:
```suggestion
type NoticeVariant = 'info' | 'error' | 'success';
```

src/components/v2/Notice/styles.ts
```diff
@@ -0,0 +1,49 @@
+import { css } from '@emotion/react';
+import { alpha, useTheme } from '@mui/material';
+
+export const useNoticeStyles = () => {
```
therealemjy: To be consistent:
```suggestion
export const useStyles = () => {
```

therealemjy (commented):
I spotted a small design issue: the border radius applied on the container should be 16px according to the designs, not 24px.

src/components/v2/Notice/styles.ts
baomastr: done

src/components/v2/Notice/styles.ts
baomastr: done

src/components/v2/Notice/index.tsx
baomastr: done

src/components/v2/Notice/styles.ts
baomastr: done

src/components/v2/Notice/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+      return 'notice';
+    case 'success':
+      return 'checkInline';
+    default:
+    case 'info':
+      return 'info';
+  }
+};
+
+export const Notice = ({ className, title, description, variant = 'info' }: INoticeProps) => {
```
coreyar, line 28: We export our variants similar to how we export variants for buttons

baomastr: added `types.ts`

therealemjy: I think Corey means we should export each variant individually, like we do for the `Button` component

therealemjy (approved):

PR #404, opened 2022-05-04
Remove old Dashboard page

therealemjy:
Replace old dashboard page with the v2 one.


baomastr (approved):

baomastr:
historical moment (:

src/pages/App.tsx
```diff
@@ -14,7 +14,6 @@ import { Layout } from 'components';
 import { init as initTranslationLibrary } from 'translation';
```
coreyar, line 14: We can `components/Dashboard`

coreyar:
We can also remove `components/Dashboard`

therealemjy:
> We can also remove `components/Dashboard`

Good catch, it's done.

PR #406, opened 2022-05-04
Add queries and mutations for VRT Conversion

coreyar:
Adds queries and mutations needed to connect the new VRT Conversion screen


src/clients/api/mutations/useApproveVrt.ts
```diff
@@ -0,0 +1,28 @@
...
+        vrtTokenContract,
+        ...params,
+      }),
+    {
+      ...options,
+    },
+  );
+};
+
+export default useApproveVrt;
```
therealemjy: Can't we use `useApproveToken` directly?

src/clients/api/queries/useGetXvsBalanceOf.ts
```diff
@@ -0,0 +1,31 @@
...
+  const xvsTokenContract = useTokenContract('xvs');
+
+  return useQuery(
+    FunctionKey.GET_XVS_BALANCE_OF,
+    () => getXvsBalanceOf({ xvsTokenContract, accountAddress }),
+    options,
+  );
+};
+
+export default useGetXvsBalanceOf;
```
therealemjy: I'd suggest creating a `useGetTokenBalance` instead, similarly to `useGetVTokenBalance`, that can then be used to get the balances for XVS and VRT (and any other token we support that's not a Venus token).

src/clients/api/queries/useGetVrtAllowance.ts
```diff
@@ -0,0 +1,31 @@
...
+  const vrtTokenContract = useTokenContract('vrt');
+
+  return useQuery(
+    FunctionKey.GET_VRT_TOKEN_ALLOWANCE,
+    () => getVrtAllowance({ vrtTokenContract, accountAddress, spenderAddress }),
+    options,
+  );
+};
+
+export default useGetVrtAllowanceIndex;
```
therealemjy: Same thing here, I'd create a `useGetTokenAllowanceIndex` that's used for all tokens

src/clients/api/mutations/useApproveVrt.ts
coreyar: We could. The difference is the contract it uses

therealemjy (approved):
Thanks for the updates.
I've approved but feel free to message me if you want to discuss why I requested the changes more, since yesterday I asked you if we could talk about this later given we had the release.

coreyar:
> Thanks for the updates. I've approved but feel free to message me if you want to discuss why I requested the changes more, since yesterday I asked you if we could talk about this later given we had the release.

All good. I saw what you were pointing out

PR #412, opened 2022-05-05
Display correct amount in withdrawal modal

coreyar:
This PR displays the correct withdrawal value when withdrawing instead of redeemed tokens

Closes https://app.clickup.com/t/29chx5f


therealemjy (approved):

therealemjy (changes requested):
That fixes one of the issue, but the other one is that the amount itself is incorrect if you withdraw the whole value. In that case, the value shown in the success modal is the redeemed value.
For example:
Screen Shot 2022-05-05 at 07 06 57
But then in the success modal:
Screen Shot 2022-05-05 at 07 07 18

therealemjy (approved):

PR #414, opened 2022-05-05
Feature: add select component

baomastr:


therealemjy (changes requested):
Hey,
Sorry but this doesn't match the designs at all:
Screen Shot 2022-05-05 at 17 06 38
Designs:
Screen Shot 2022-05-05 at 17 06 32

therealemjy:
Much better 👍

src/components/v2/Select/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+  options: Option[];
+}
+
+const DEFAULT_OPTION = {
+  value: '',
+  label: 'All',
+};
+
+export const Select = ({ className, options }: ISelectProps) => {
+  const [selectedValue, setSelectedValue] = React.useState(DEFAULT_OPTION.value);
```
therealemjy: The state and its modifier function should be passed through props, as we'll always need to use the state from the parent component that renders it

src/components/v2/Select/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+interface ISelectProps {
+  className?: string;
+  options: Option[];
+}
+
+const DEFAULT_OPTION = {
+  value: '',
+  label: 'All',
+};
```
therealemjy: I'd remove that to make it more flexible (not all selects will have a default "All" option).

src/components/v2/Select/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+export const Select = ({ className, options }: ISelectProps) => {
```
therealemjy: Please use `React.FC` to type it like we do with other components

therealemjy (commented):
It displays a title on mobile:
Screen Shot 2022-05-08 at 22 05 57

Since we don't have mobile designs for this component I think it's fine to use MUI's default one, but let's remove the title.

src/components/v2/Select/index.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { withCenterStory } from 'stories/decorators';
+import { Select } from '.';
+
+export default {
+  title: 'Components/Select',
+  component: Select,
+  decorators: [withCenterStory({ width: 200 })],
+  parameters: {
+    backgrounds: {
+      default: '#1F2028',
```
coreyar: Ideally this value comes from the enum of options we have set `Primary` `Paper` `White`. They are set in `.storybook/preview.js` if not we can grab them from the Palette constant

baomastr:
@therealemjy 

> Since we don't have mobile designs for this component I think it's fine to use MUI's default one, but let's remove the title.

here are mobile designs:

image

therealemjy:
Ah right, nice 👍
Looks good to me.
@coreyar Thanks for updating the PR

therealemjy (approved):

src/components/v2/Select/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+      anchorOrigin: {
+        vertical: 'top',
+        horizontal: 'center',
+      },
+    };
+    return {
+      PaperProps: {
+        sx: styles.menuWrapper,
+      },
+      ...(isSmDown ? mobileStyles : {}),
```
therealemjy, line 63: Actually, I just saw that.

We want to move away from using JS for responsiveness: let's use CSS directly here (you can render both mobile and desktop instances of the component and choose which one to render based on the screen size using CSS directly).
Also, please don't copy and paste the whole code: you can make it so only the style prop differs from one version to the other.

therealemjy (changes requested):

therealemjy:
Actually never mind, we can't use CSS only here because of how MUI renders the dropdown menu.
Looks all good to me then 👍

therealemjy (approved):

PR #415, opened 2022-05-05
Display hypothetical supply balance on SupplyWithdrawForm

therealemjy:
- update `ValueUpdate` so its types are better inferred
- add `ValueUpdate` component on `SupplyWithdrawForm` to display hypothetical supply balance

### DIFF
--- a/src/components/v2/ValueUpdate/index.tsx
+++ b/src/components/v2/ValueUpdate/index.tsx
@@ -6,32 +6,24 @@ import { formatCentsToReadableValue } from 'utilities/common';
 import { Icon } from '../Icon';
 import { useStyles } from './styles';
 
-interface INumberValueProps {
-  original: number | undefined;
-  update: number | undefined;
-  format?: (value: number | undefined) => string;
-}
-
-interface IBigNumberValueProps {
-  original: BigNumber | undefined;
-  update: BigNumber | undefined;
-  format?: (value: BigNumber | undefined) => string;
-}
-
-interface ValueUpdateCommonProps {
+export interface IValueUpdateProps<T> {
+  original: T;
+  update: T;
+  format?: (value: T) => string;
   className?: string;
   positiveDirection?: 'asc' | 'desc';
 }
 
-export type ValueUpdateProps = ValueUpdateCommonProps & (INumberValueProps | IBigNumberValueProps);
-
-export const ValueUpdate: React.FC<ValueUpdateProps> = ({
+export function ValueUpdate<T>({
   className,
   original,
   update,
-  format = (value: ValueUpdateProps['original']) => formatCentsToReadableValue({ value }),
+  format = (value: T) =>
+    formatCentsToReadableValue({
+      value: value instanceof BigNumber || typeof value === 'number' ? value : undefined,
+    }),
   positiveDirection = 'asc',
-}) => {
+}: React.PropsWithChildren<IValueUpdateProps<T>>) {
   let isImprovement = false;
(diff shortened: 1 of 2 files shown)


src/components/v2/ValueUpdate/index.tsx
```diff
@@ -6,32 +6,24 @@ import { formatCentsToReadableValue } from 'utilities/common';
...
+  update: T;
+  format?: (value: T) => string;
   className?: string;
   positiveDirection?: 'asc' | 'desc';
 }
 
-export type ValueUpdateProps = ValueUpdateCommonProps & (INumberValueProps | IBigNumberValueProps);
-
-export const ValueUpdate: React.FC<ValueUpdateProps> = ({
+export function ValueUpdate<T>({
```
coreyar, line 17: This could remain an arrow function and still be made generic
```suggestion
export const ValueUpdate = <T,>({
```

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -81,8 +81,10 @@ export const SupplyWithdrawUi: React.FC<ISupplyWithdrawUiProps & ISupplyWithdraw
       ]
     : [];
 
-  const calculateNewSupplyAmount = (initial: BigNumber, amount: BigNumber) => initial.plus(amount);
-  const calculateNewBorrowAmount = (initial: BigNumber, amount: BigNumber) => initial.minus(amount);
+  const calculateNewSupplyAmount =
```
coreyar: I think this is more clearly written as two separate functions. Writing it as two functions introduces unnecessary logic and arguments. Was there an issue with using the individual functions. 

`calculateNewBorrowAmount` probably should be named `calculateNewWithdrawBalance`

therealemjy: In both cases it's the supply balance you're calculating, which goes up if you're supplying more tokens or goes up if you're withdrawing some

src/components/v2/ValueUpdate/index.tsx
therealemjy: TypeScript doesn't seem to like that. I haven't seen that notation before 
Screen Shot 2022-05-05 at 14 17 07

coreyar: Interesting. I didn't receive any errors when I tested it. That syntax generally works but the compiler can get tripped up in `.tsx` files

src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
coreyar: Yes, you are right, I was confused there. I still think they are nicer as two functions opposed to one function with a boolean flag

therealemjy: Can do. What would you name them though?

coreyar: `calculateAdditionalSupplyAmount` and `calculateWithdrawalSupplyAmount`

coreyar: Looks like this is updated in in PR 425

therealemjy: Indeed yeah, sorry this got messy. I merged a PR I shouldn't have and ended up messing up the PR order. It's sorted now.

coreyar (approved):

PR #418, opened 2022-05-05
Feature: add useCardLayoutFrom prop for Tables

baomastr:
https://app.clickup.com/t/2gvefed

### DIFF
--- a/src/components/v2/Table/Head.tsx
+++ b/src/components/v2/Table/Head.tsx
@@ -7,7 +7,7 @@ import TableRow from '@mui/material/TableRow';
 import Box from '@mui/material/Box';
 import { visuallyHidden } from '@mui/utils';
 import { Icon } from '../Icon';
-import { useStyles } from './styles';
+import { useHeadStyles as useStyles } from './styles';
 
 interface IHeadProps<C extends { key: string; label: string; orderable: boolean }[]> {
   columns: C;
@@ -24,7 +24,7 @@ function Head<C extends { key: string; label: string; orderable: boolean }[]>({
   onRequestOrder,
   className,
 }: IHeadProps<C>) {
-  const styles = useStyles({});
+  const styles = useStyles();
   return (
     <TableHead>
       <TableRow className={className}>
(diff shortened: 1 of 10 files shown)


src/components/v2/Table/Head.tsx
```diff
@@ -24,7 +24,7 @@ function Head<C extends { key: string; label: string; orderable: boolean }[]>({
   onRequestOrder,
   className,
 }: IHeadProps<C>) {
-  const styles = useStyles();
+  const styles = useStyles({});
```
therealemjy: I'd make the param optional instead. Passing an empty object doesn't make sense

src/components/v2/Table/index.tsx
```diff
@@ -97,17 +100,17 @@ export const Table = ({
...
+  }, [columns, gridTemplateColumns, isCardLayout]);
 
   return (
-    <Paper css={[styles.root, isSmDown && styles.rootMobile]} className={className}>
-      {title && <h4 css={[styles.title, isSmDown && styles.titleMobile]}>{title}</h4>}
+    <Paper css={[styles.root, isCardLayout && styles.rootMobile]} className={className}>
+      {title && <h4 css={[styles.title, isCardLayout && styles.titleMobile]}>{title}</h4>}
 
-      {isSmDown ? (
+      {isCardLayout ? (
```
therealemjy, line 113: As we spoke about earlier this week, we want to move away from using Javascript and use CSS instead. You can render both variants and use a CSS media query to hide one or the other using `display: none`.

src/components/v2/Table/index.tsx
```diff
@@ -7,8 +7,9 @@ import TableContainer from '@mui/material/TableContainer';
 import TableRow from '@mui/material/TableRow';
 import Paper from '@mui/material/Paper';
 import Typography from '@mui/material/Typography';
+import { Breakpoint } from '@mui/material';
```
therealemjy: This should use the properties of our `BREAKPOINT` constant from the theme instead

therealemjy (changes requested):

src/components/v2/Table/index.tsx
baomastr: i think, hook use is justified here.
we have different layouts for this cases. that's such hooks are made for.

src/components/v2/Table/index.tsx
baomastr: done

src/components/v2/Table/Head.tsx
baomastr: done

PR #420, opened 2022-05-05
Apply missing styles when tooltip is not rendered

coreyar:
Apply component with passed prop styles when not rendering track tooltip


coreyar:
Something is wrong this should have failed storybooks

therealemjy (approved):
Good catch 👍

PR #421, opened 2022-05-05
Add disabled input styles

coreyar:
Add styles for disabled input.

### DIFF
--- a/src/components/v2/TextField/index.tsx
+++ b/src/components/v2/TextField/index.tsx
@@ -9,7 +9,7 @@ import { useStyles } from './styles';
 export interface ITextFieldProps extends InputHTMLAttributes<HTMLInputElement> {
   className?: string;
   label?: string;
-  description?: string;
+  description?: string | React.ReactElement;
   hasError?: boolean;
   leftIconName?: IconName;
   rightAdornment?: React.ReactElement;
@@ -26,6 +26,7 @@ export const TextField: React.FC<ITextFieldProps> = ({
   max,
   min,
   type,
+  disabled,
   ...inputProps
 }) => {
   const styles = useStyles();
@@ -58,7 +59,7 @@ export const TextField: React.FC<ITextFieldProps> = ({
         </Typography>
       )}
 
-      <Box css={styles.getInputContainer({ hasError })}>
+      <Box css={styles.getInputContainer({ hasError, disabled })}>
         {!!leftIconName && (
           <Icon name={leftIconName} size={styles.theme.spacing(6)} css={styles.leftIcon} />
         )}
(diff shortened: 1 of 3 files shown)


baomastr (approved):

src/components/v2/TextField/index.tsx
```diff
@@ -26,6 +26,7 @@ export const TextField: React.FC<ITextFieldProps> = ({
   max,
   min,
   type,
+  disabled,
```
therealemjy, line 29: Because the prop is now extracted from here, you also need to pass it to the `input` directly (at the moment the input doesn't get disabled because of that)

src/components/v2/TextField/styles.ts
```diff
@@ -12,14 +12,26 @@ export const useStyles = () => {
...
       border-radius: ${theme.spacing(3)};
-      border: ${theme.spacing(0.5)} solid
-        ${hasError ? theme.palette.interactive.error : 'transparent'};
-      background-color: ${theme.palette.background.default};
+      ${hasError
+        ? `border: ${theme.spacing(0.5)} solid ${theme.palette.interactive.error}`
+        : 'transparent'};
+      ${disabled
+        ? `border: ${theme.spacing(0.5)} solid ${theme.palette.secondary.light}`
+        : 'transparent'};
```
therealemjy: 'transparent' gets directly printed in the CSS if `hasError` or `disabled` are `false`. Since only the border color is variable, we can avoid having to redeclare the whole style:
```suggestion
border: ${theme.spacing(0.5)} solid transparent;
${(hasError || disabled) &&
  css`
    border-color: ${hasError ? theme.palette.secondary.error : theme.palette.interactive.light};
  `};
```

baomastr: for me first variant is more clear to read

coreyar: I agree with Dmitriy, I prefer not to nest unary and ternary statements

therealemjy: We can keep the first syntax, but the update you made changes the UI incorrectly: the component should always have a border regardless of it state. It's transparent by default, grey if disabled and red if `hasError` is true.

src/components/v2/TextField/styles.ts
```diff
@@ -12,21 +11,38 @@ export const useStyles = () => {
...
-        ${hasError ? theme.palette.interactive.error : 'transparent'};
-      background-color: ${theme.palette.background.default};
+    getInputContainer: ({
+      hasError,
+      disabled,
+    }: {
+      hasError: boolean;
+      disabled: boolean | undefined;
+    }) => {
+      let borderColor: undefined | string = 'transparent';
```
therealemjy, line 21: Nit: `borderColor` is never undefined

therealemjy (approved):

src/components/v2/TextField/styles.ts
coreyar: According to the theme typing `theme.palette.interactive.error` could be undefined

PR #422, opened 2022-05-06
Wire up convert vrt

coreyar:


therealemjy (commented):
The Convert VRT link in the sidebar is only present on the mainnet, it looks like Dmitriy updated that a while back but it should display on all environment.

src/clients/api/queries/getVrtConversionEndTime.ts
```diff
@@ -4,11 +4,13 @@ export interface IGetVrtConversionEndTimeInput {
...
+export type GetVrtConversionEndTimeOutput = Date;
 
-const getVrtConverstionEndTime = ({
+const getVrtConverstionEndTime = async ({
   vrtConverterContract,
-}: IGetVrtConversionEndTimeInput): Promise<GetVrtConversionEndTimeOutput> =>
-  vrtConverterContract.methods.conversionEndTime().call();
+}: IGetVrtConversionEndTimeInput): Promise<GetVrtConversionEndTimeOutput> => {
+  const resp = await vrtConverterContract.methods.conversionEndTime().call();
+  return new Date(+resp * 1000);
```
therealemjy, line 14: It would be nice to add a comment explaining the value retrieved is in seconds and so you convert it into milliseconds.

src/components/v2/SuccessfulTransactionModal/index.tsx
```diff
@@ -40,17 +42,19 @@ export const SuccessfulTransactionModal: React.FC<ISuccessfulTransactionModalPro
         <h3 css={styles.title}>{title}</h3>
 
         <div css={styles.messageContainer}>
-          <Typography variant="small1" component="p">
-            {message}
-          </Typography>
-
+          {message && (
```
therealemjy: Message could be an empty string:
```suggestion
          {!!message && (
```

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -1,141 +1,218 @@
...
+            <Typography variant="small2" css={[styles.fontWeight600, styles.successMessage]}>
+              {xvsAmountWei &&
+                convertWeiToCoins({
+                  valueWei: xvsAmountWei,
+                  tokenId: XVS_ID,
+                  returnInReadableFormat: true,
+                })}
+            </Typography>
+          </div>
+        ),
```
therealemjy: This will create some conflicts if `message` and `amount` are also passed to the function.
How about renaming `message` to `content` inside `SuccessfulTransactionModal` and make it so it can be a string or a React element?
We'd be able to do what you're doing here and avoid potential visual glitches from passing too many props.

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -1,141 +1,218 @@
...
-                        values={{ amount: format(new BigNumber(vrtLimitWei)) }}
-                      />
+        <EnableToken
+          title={t('convertVrt.enableVrt')}
+          assetId={VRT_ID}
+          isEnabled={userVrtEnabled}
+          tokenInfo={[
+            {
+              iconName: 'xvs',
+              label: 'VRT Conversion Ratio',
```
therealemjy: This needs to be internationalized

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -1,141 +1,218 @@
...
+              children: xvsToVrtConversionRatio?.toFixed(6),
+            },
+            { iconName: 'vrt', label: 'Current VRT Balance', children: readableUserVrtBalance },
```
therealemjy: The label needs to be internationalized

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -1,141 +1,218 @@
...
+        >
+          <AmountForm onSubmit={onSubmit} maxAmount={userVrtBalance} css={styles.form}>
+            {({ values }) => {
+              const xvsValue =
+                values.amount && xvsToVrtConversionRatio
+                  ? new BigNumber(values.amount)
+                      .times(xvsToVrtConversionRatio)
+                      .dp(VRT_DECIMAL)
+                      .toFixed()
+                  : '';
```
therealemjy: How about making this into a function wrapped with `React.useCallback`, outside the render block, that can then be called to get the converted XVS amount?

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -1,141 +1,218 @@
...
   walletConnected,
 }) => {
   const styles = useStyles();
   const { t, Trans } = useTranslation();
+  const { openSuccessfulTransactionModal } = useSuccessfulTransactionModal();
+  const vrtConverterProxyAddress = getContractAddress('vrtConverterProxy');
+
   const readableXvsAvailable = useConvertToReadableCoinString({
-    valueWei: xvsTotalWei,
+    valueWei: userVrtBalanceWei?.times(xvsToVrtConversionRatio || 1),
```
therealemjy: I don't think we should default the conversation ratio to 1, as this could lead to displaying incorrect data. We should display the placeholder key if xvsToVrtConversionRatio is undefined.

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -1,141 +1,218 @@
...
+                        <Trans
+                          i18nKey="convertVrt.balance"
+                          components={{
+                            White: <span css={styles.whiteLabel} />,
+                          }}
+                          values={{ amount: readableUserVrtBalance }}
+                        />
+                      }
+                      rightMaxButton={{
+                        label: t('convertVrt.max').toUpperCase(),
```
therealemjy: I'd avoid modifying texts, as we want translators to have as much control on sentences as much as possible. Instead the translation can be upper cased by default.
This way you also won't need a test ID to target the button, you can use the translation key directly.

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -1,141 +1,218 @@
...
+                        valueOnClick: userVrtBalance || '',
```
therealemjy: If a user clicks on the max button and it doesn't nothing, they'd think it's broken. I think we should just hide the max button if `userVrtBalance` is undefined

therealemjy (commented):
This input uses a different amount of decimals than the balance displayed, they should both use the same decimals:
Screen Shot 2022-05-09 at 17 35 03

src/hooks/useConvertToReadableCoinString.ts
```diff
@@ -15,11 +16,11 @@ const useConvertToReadableCoinString = ({
...
       valueWei
         ? convertWeiToCoins({
-            value: valueWei,
+            valueWei,
             tokenId,
             returnInReadableFormat: true,
-          }).toString()
-        : '-',
+          })
+        : PLACEHOLDER_KEY,
```
therealemjy, line 23: Nice

src/pages/ConvertVrt/Withdraw/index.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+      </AuthContext.Provider>,
+    );
+    const submitButton = getByText(en.convertVrt.withdrawXvs).closest(
+      'button',
+    ) as HTMLButtonElement;
+    expect(submitButton).toBeEnabled();
+    await act(async () => {
+      await waitFor(() => fireEvent.click(submitButton));
+    });
+    await waitFor(() => expect(withdrawXvs).toHaveBeenCalledTimes(1));
```
therealemjy, line 54: It would be good to check it's called with the correct params

src/pages/ConvertVrt/Withdraw/index.tsx
```diff
@@ -1,35 +1,72 @@
...
 import { useTranslation } from 'translation';
 import { XVS_ID } from '../constants';
 import { useStyles } from '../styles';
 
-interface WithdrawProps {
-  xvsTotal: BigNumber;
+export interface IWithdrawProps {
+  xvsWithdrawableAmount: BigNumber | undefined;
+  withdrawXvs: () => Promise<string>;
+  xvsWithdrawlLoading: boolean;
```
therealemjy: How about using the convention `is<NAME>Loading` for boolean values like that?

src/pages/ConvertVrt/index.tsx
```diff
@@ -53,15 +75,82 @@ export const ConvertVrtUi = ({
...
+        vrtConversionLoading={vrtConversionLoading}
+        userVrtEnabled={userVrtEnabled}
+        convertVrt={handleConvertVrt}
+        withdrawXvs={handleWithdrawXvs}
+        xvsWithdrawlLoading={xvsWithdrawlLoading}
+        xvsWithdrawableAmount={xvsWithdrawableAmount}
+      />
+    );
+  }
+  return <LoadingSpinner />;
```
therealemjy: I know we don't have any designs for this yet, but I think we should add a todo to display critical errors in the future.

therealemjy (commented):
The success modal displays the token symbols twice:
Screen Shot 2022-05-09 at 17 28 06

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -1,141 +1,218 @@
...
+  };
+  const userVrtBalance =
+    userVrtBalanceWei &&
+    convertWeiToCoins({ valueWei: userVrtBalanceWei, tokenId: VRT_ID }).toFixed();
   return (
     <div css={styles.root}>
       {walletConnected ? (
         <section css={styles.title}>
-          <Typography variant="h3">{readableXvsAvailable}</Typography>
+          <Typography variant="h3">{readableXvsAvailable || PLACEHOLDER_KEY}</Typography>
```
therealemjy: I guess `|| PLACEHOLDER_KEY` shouldn't be needed since `readableXvsAvailable` uses `useConvertToReadableCoinString` which already defaults to the placeholder key

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -1,141 +1,218 @@
...
+                  <div css={styles.inputSection}>
+                    <Typography variant="small2" css={[styles.inputLabel, styles.fontWeight600]}>
+                      {t('convertVrt.youWillReceive')}
+                    </Typography>
+                    <TokenTextField
+                      tokenId={XVS_ID}
+                      name="xvs"
+                      css={styles.input}
+                      description={t('convertVrt.vrtEqualsXvs', {
+                        xvsToVrtConversionRatio: xvsToVrtConversionRatio?.toFixed(6),
```
therealemjy: Did you set it to 6 decimals because it shows 6 in the designs?
I think we should display it as it is returned to us by the smart contract

src/pages/ConvertVrt/Convert/index.tsx
coreyar: I am going to remove this. I added this before the designs were complete for this view.

src/pages/ConvertVrt/Convert/index.tsx
coreyar: I am going to remove this. I added this before the designs were complete for this view.

src/pages/ConvertVrt/Convert/index.tsx
coreyar: I used 6 decimals because the current view uses 6 decimals

src/pages/ConvertVrt/Convert/index.tsx
coreyar: I made one but it got dropped here during a rebase it looks like.

src/pages/ConvertVrt/Withdraw/index.spec.tsx
coreyar: It doesn't get called with any params

src/pages/ConvertVrt/Convert/index.tsx
therealemjy: Got you. I think we should use whatever is returned from the smart contract since that's the rate we use. I'll ask Pavel to confirm

therealemjy: Just got the confirmation from Pavel:
Screen Shot 2022-05-10 at 15 48 05

src/pages/ConvertVrt/Withdraw/index.spec.tsx
therealemjy: Ah right, my bad! I got mixed up.

src/pages/ConvertVrt/index.tsx
```diff
@@ -53,15 +75,83 @@ export const ConvertVrtUi = ({
...
+    return res.transactionHash;
+  };
+
+  const conversionRatio = useMemo(() => {
+    if (vrtConversionRatio) {
+      return new BigNumber(vrtConversionRatio).div(CONVERSION_RATIO_DECIMAL);
+    }
+    return undefined;
+  }, [vrtConversionRatio]);
+  if (conversionRatio && userVrtBalanceWei && vrtConversionEndTime && xvsVestedBalanceWei) {
```
therealemjy: In the case where no wallet is connected, the loading spinner will display indefinitely because `xvsVestedBalanceWei` and `userVrtBalanceWei` are `undefined`.

I think all the side effects should be handled by the individual tabs, which would solve that issue since they're the ones displaying `EnableToken` and `ConnectWallet`.

src/pages/ConvertVrt/index.tsx
```diff
@@ -53,15 +75,83 @@ export const ConvertVrtUi = ({
...
+    }
+    const res = await withdrawXvs({
+      accountAddress,
+    });
+    return res.transactionHash;
+  };
+
+  const conversionRatio = useMemo(() => {
+    if (vrtConversionRatio) {
+      return new BigNumber(vrtConversionRatio).div(CONVERSION_RATIO_DECIMAL);
```
therealemjy, line 130: Rather than having a constant written in a separate file, we should use our main constant file as source of truth and which contains all the decimals of all the supported tokens.
You can use `getToken('xvs').decimals` to get the decimals of the XVS.

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -1,50 +1,134 @@
...
+      userVrtBalanceWei &&
+      convertWeiToCoins({
+        valueWei: userVrtBalanceWei,
+        tokenId: VRT_ID,
+      });
+
+    return formatCoinsToReadableValue({
+      value: userVrtBalanceCoins,
+      tokenId: VRT_ID,
+      shorthand: true,
```
therealemjy: I know this isn't well described in the designs (which makes us waste a bit of time on things like this :/), but just to be consistent with the rest of the app (Borrow/Supply forms for example) I think we should not use the shorthand option here.

src/pages/ConvertVrt/Convert/index.tsx
```diff
@@ -1,50 +1,134 @@
...
+    });
+  }, [userVrtBalanceWei?.toString()]);
```
therealemjy: So after digging yesterday, I found out we should avoid using `toString` because it returns large values using exponentials (1e22). This means that in this case some updates might not happen because two large different numbers with the same amount of exponentials and the same first digit will be returned the same.
We should use `toFixed` to avoid this issue.

src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
```diff
@@ -52,7 +52,7 @@ export const MintVaiUi: React.FC<IMintVaiUiProps> = ({
 
   const vaiToken = getToken(VAI_ID);
   const limitTokens = useMemo(
-    () => (limitWei ? convertWeiToCoins({ value: limitWei, tokenId: VAI_ID }).toString() : '0'),
+    () => (limitWei ? convertWeiToCoins({ valueWei: limitWei, tokenId: VAI_ID }).toString() : '0'),
     [limitWei?.toString()],
```
therealemjy: Following my previous comment:
```suggestion
    () => (limitWei ? convertWeiToCoins({ valueWei: limitWei, tokenId: VAI_ID }).toFixed() : '0'),
    [limitWei?.toFixed()],
```

src/pages/Dashboard/MintRepayVai/RepayVai/index.tsx
```diff
@@ -55,7 +55,7 @@ export const RepayVaiUi: React.FC<IRepayVaiUiProps> = ({
         ? BigNumber.minimum(userBalanceWei, userMintedWei)
         : new BigNumber(0);
 
-    return convertWeiToCoins({ value: limitWei, tokenId: VAI_ID }).toString();
+    return convertWeiToCoins({ valueWei: limitWei, tokenId: VAI_ID }).toString();
   }, [userBalanceWei?.toString(), userMintedWei?.toString()]);
```
therealemjy:
```suggestion
    return convertWeiToCoins({ valueWei: limitWei, tokenId: VAI_ID }).toFixed();
  }, [userBalanceWei?.toFixed(), userMintedWei?.toFixed()]);
```

therealemjy:
A few comments following updates and findings from yesterday (that we need to use `toFixed` instead of `toString` with bignumber values), but feature-wise it's perfect.

src/pages/Dashboard/MintRepayVai/RepayVai/index.tsx
coreyar: solved with rebase

src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
coreyar: solved with rebase

therealemjy (approved):

therealemjy (approved):

PR #426, opened 2022-05-06
Run tests before commit and display coverage files

coreyar:
This PR runs tests before committing and shows coverage on committed files.

While this will add a bit of time to the commit checks I think this is nice because the coverage output even ignoring files that will be removed is really long and not easy to digest. At some point we could add a full report to PRs and enforce a level of coverage with a badge on the README.

My thought is viewing coverage on changed will help guide writing tests as changes are made.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,1 +1,2 @@
 yarn tsc
+yarn test --watchAll=false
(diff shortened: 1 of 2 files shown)


therealemjy:
I don't recommend doing either of those things.
Running the tests before any push slows down the process a whole lot; the tests should be run on PRs only.

Same goes for the coverage. I'm OK with displaying the coverage on PRs but we don't need to show it before every commit.

PR #433, opened 2022-05-06
Set AuthContext value from renderComponent in tests

coreyar:
To set the AuthContext in tests we are wrapping the tested component directly, this is cumbersome to just set a single value on the context.
It also results in the component getting wrapped twice in the AuthContext.

The same refactor has been made for VaiContext


therealemjy (commented):
Nice, I think that's very useful 👍
You might just need to rebase with develop (it looks like the branch comes from another one where you have the `Spinner` component).

therealemjy (approved):

therealemjy (approved):

PR #436, opened 2022-05-08
Update supply form tests

therealemjy:
- Update mutations used by `SupplyWithdraw` component so `amount` param uses `BigNumber` to be consistent with other mutations

Tests added:
- Wallet balance displayed is correct
- Supply balance displayed is correct
- Pressing on “MAX” button updates input value to wallet balance
- User can’t supply amount that’s higher than their wallet balance
- User can supply amount that is less or equal to their wallet balance

### DIFF
--- a/src/clients/api/mutations/redeemUnderlying.spec.ts
+++ b/src/clients/api/mutations/redeemUnderlying.spec.ts
@@ -1,4 +1,10 @@
+import BigNumber from 'bignumber.js';
+
+import fakeAccountAddress from '__mocks__/models/address';
+import { VBep20 } from 'types/contracts';
 import redeemUnderlying from './redeemUnderlying';
+
+const fakeAmount = new BigNumber(10000000000000000);
 
 describe('api/mutation/redeemUnderlying', () => {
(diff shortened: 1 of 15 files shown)


src/clients/api/mutations/redeemUnderlying.spec.ts
```diff
@@ -38,18 +41,18 @@ describe('api/mutation/redeemUnderlying', () => {
...
         redeemUnderlying: redeemUnderlyingMock,
       },
-    } as unknown as any;
+    } as unknown as VBep20;
 
     const response = await redeemUnderlying({
       tokenContract: fakeContract,
-      amount: fakeAmountWei,
-      account: fakeFromAccountsAddress,
+      amount: fakeAmount,
```
coreyar: Should we label these amounts as `wei`?

therealemjy: Good idea 👍

therealemjy: Updated

coreyar (approved):

PR #437, opened 2022-05-09
Setup spinner with Lottie animation

coreyar:
Adds lottie animation support and a loading spinner

Includes a commit updating timer setup per the `react-testing-library` docs


therealemjy (approved):
So should we use this spinner for the loading state of buttons too?

PR #439, opened 2022-05-09
Center success modal over content

coreyar:
Offset the success modal by the rendered sidebar
Center title text on mobile


therealemjy (approved):
Just saw Pavel's message confirming that's the UI we want. Looks perfect in that case 👍

PR #444, opened 2022-05-10
Create ApyChart component

therealemjy:
- update Rechart to 2.1.9
- add `tiny` variant to theme (recently added to designs: https://www.figma.com/file/GYJwWWLbC4B8DrJOPcETz3/Venus-2.0?node-id=7%3A203)
- update theme type to make it stricter
- Create `ApyChart` component
Screen Shot 2022-05-10 at 16 32 51
Screen Shot 2022-05-10 at 16 32 54


src/components/v2/ApyChart/formatToReadableDate.ts
```diff
@@ -0,0 +1,6 @@
+import moment from 'moment';
```
coreyar: `moment` isn't recommended anymore. We should look at using `luxon` or `date-fns`

src/components/v2/ApyChart/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+                        }),
+                      }}
+                    />
+                  </div>
+                </div>
+              ) : (
+                // Although the type definition of "content" allows an undefined
+                // value to be returned, it actually creates a runtime error if
+                // we do so. For that reason we return a fragment instead
+                <></>
```
coreyar: null should be returned in place of rendering `undefined`. We can also remove the comment if we return null

src/components/v2/ApyChart/formatToReadableDate.ts
therealemjy: Works for me 👍

src/components/v2/ApyChart/index.tsx
therealemjy: That's much nicer 👍

coreyar (approved):

PR #458, opened 2022-05-11
Format percent without decimals on borrow table

coreyar:
Format percent without decimals on borrow table using utility function


therealemjy (approved):
Nice 👍

PR #459, opened 2022-05-11
Refactor Table and Card styles

coreyar:
[4afd52b](https://github.com/VenusProtocol/venus-protocol-interface/pull/459/commits/4afd52b185cb624cb014973943285c0816a25002) Hide and show table/cards responsively with media queries

### DIFF
--- a/src/pages/Dashboard/Markets/SupplyMarket/SupplyMarketTable.tsx
+++ b/src/pages/Dashboard/Markets/SupplyMarket/SupplyMarketTable.tsx
@@ -5,7 +5,7 @@ import { formatCoinsToReadableValue, formatToReadablePercentage } from 'utilitie
 import { Asset, TokenId } from 'types';
 import { Table, ITableProps, Token, Toggle } from 'components';
 import { useTranslation } from 'translation';
-import { useIsSmDown, useIsLgDown } from 'hooks/responsive';
+import { useStyles as useSharedStyles } from '../styles';
 import { useStyles } from './styles';
 
 export interface ISupplyMarketTableUiProps {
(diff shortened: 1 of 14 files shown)


src/pages/Dashboard/Markets/SupplyMarket/SupplyMarketTable.tsx
```diff
@@ -91,7 +89,13 @@ export const SupplyMarketTable: React.FC<ISupplyMarketTableUiProps> = ({
...
       rowOnClick={rowOnClick}
       rowKeyIndex={0}
-      gridTemplateColumns={styles.getGridTemplateColumns({ isCardLayout: isSmDown })}
+      gridTemplateColumnsCards={styles.getGridTemplateColumns}
+      tableClassName={TABLE_CLASS_NAME}
+      cardClassName={CARD_CLASS_NAME}
+      css={sharedStyles.marketTable({
+        tableClassName: TABLE_CLASS_NAME,
+        cardClassName: CARD_CLASS_NAME,
+      })}
```
therealemjy: Since all these classes do is define at what breakpoint each layout should be shown or hidden, why not follow the original idea we had of adding a prop to specify that breakpoint directly?
https://app.clickup.com/t/2gvefed

coreyar: We can. Passing as styles opens up the door to customize more than just the breakpoint to show cards or the table. Eventually, I'd like to break these out into completely separate components that share context for pagination, ordering, filtering ect. In that scenario there would not be a breakpoint prop.

therealemjy (approved):

PR #461, opened 2022-05-11
Numerator should be token value in dollars

coreyar:
Numerator should be token value in dollars, previously it was the token amount


therealemjy (approved):
Right, makes sense

PR #462, opened 2022-05-12
Create InterestRateChart component

therealemjy:
- create `charts` folder inside v2 components directory to hold all chart components
- create `InterestRateChart` component
Screen Shot 2022-05-12 at 12 59 32
- reuse styles and `TooltipContent` between chart components


src/components/v2/charts/InterestRateChart/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+export const InterestRateChart: React.FC<IInterestRateChartProps> = ({
+  className,
+  currentUtilizationRate,
+  data,
+}) => {
+  const sharedStyles = useSharedStyles();
+  const localStyles = useLocalStyles();
+  const styles = {
+    ...sharedStyles,
```
coreyar: The risk here is that we over ride styles with the same key, which could be either a feature or a bug. Either way it is clearer where the styles come from if we don't spread them.

src/components/v2/charts/ApyChart/index.tsx
```diff
@@ -11,29 +11,36 @@ import {
...
+  data: IApyChartItem[];
   type: 'supply' | 'borrow';
   className?: string;
 }
 
 export const ApyChart: React.FC<IApyChartProps> = ({ className, data, type }) => {
-  const styles = useStyles();
+  const sharedStyles = useSharedStyles();
+  const localStyles = useLocalStyles();
+  const styles = {
```
coreyar: Same as previous comment about spreading styles

src/components/v2/charts/InterestRateChart/index.tsx
therealemjy: In this case it would indeed be a feature. The idea is that `sharedStyles` represent the base styles used by all components importing it. If one of the component wishes to override of the element, it can do so by redeclaring it inside its own local styles. In practice, we haven't had to do that so far, but it's an option.

That said, separating them works for me too. If you feel it's clearer to you, I'll make the change 👍

coreyar (approved):

PR #468, opened 2022-05-12
Market details - UI

therealemjy:
- create `Card` component to be used for each section on the Market Details page
- create UI of Market Details page


src/pages/MarketDetails/Card/index.tsx
```diff
@@ -1,19 +1,61 @@
...
+
+      {stats.length > 0 && (
+        <div css={styles.row}>
+          {stats.map(stat => (
+            <div css={styles.stat} key={`card-${title}-legend-${stat.label}`}>
+              <Typography variant="small2" component="div" css={styles.statLabel}>
+                {stat.label}
+              </Typography>
+
+              <Typography variant="h4" component="div" css={styles.statValue}>
```
coreyar: I think there is usually a more descriptive element to use than `div` for displaying text. I usually fallback to span if nothing works well. H4 seems like a possible option here for example

therealemjy: In this case an `h4` wouldn't be correct, as it's a value we're displaying not a title. I can use a `span` here if you prefer

coreyar (approved):

PR #472, opened 2022-05-13
Always enable treasury balance query

coreyar:
Always enable treasury balance query because it doesn't require a conn…ected wallet


therealemjy:
Good point  👍

therealemjy (approved):

PR #491, opened 2022-05-16
Use notice component for banner message

baomastr:
- added warning style for notice
- notice is used for banner message


src/components/v2/Icon/icons/index.tsx
```diff
@@ -34,6 +34,7 @@ export { ReactComponent as checkInline } from './checkInline.svg';
 export { ReactComponent as mark } from './mark.svg';
 export { ReactComponent as arrowShaft } from './arrowShaft.svg';
 export { ReactComponent as notice } from './notice.svg';
+export { ReactComponent as warning } from './warning.svg';
```
coreyar: I think we currently have this SVG because it is being rendered with the name `attention`

baomastr: removed warning icon, used attention.

therealemjy (approved):

PR #492, opened 2022-05-16
Add XVS header with ellipse text component

coreyar:
Add XVS header UI
Add a responsive component for adding ellipse to text

### DIFF
--- a/src/components/v2/ProgressBar/LabeledProgressBar/index.tsx
+++ b/src/components/v2/ProgressBar/LabeledProgressBar/index.tsx
@@ -9,6 +9,7 @@ interface ILabeledProgressBar extends IProgressBarProps {
   whiteLeftText?: string;
   greyRightText?: string;
   whiteRightText?: string;
+  className?: string;
 }
 
 export const LabeledProgressBar: React.FC<ILabeledProgressBar> = ({
@@ -16,12 +17,13 @@ export const LabeledProgressBar: React.FC<ILabeledProgressBar> = ({
   whiteRightText,
   greyLeftText,
   whiteLeftText,
+  className,
   ...progressBarProps
 }) => {
   const styles = useStyles();
   return (
-    <div>
-      <div css={styles.topProgressBarLegend}>
+    <>
+      <div className={className} css={styles.topProgressBarLegend}>
         <div css={styles.inlineContainer}>
           {greyLeftText && (
             <Typography component="span" variant="small2" css={styles.inlineLabel}>
@@ -50,6 +52,6 @@ export const LabeledProgressBar: React.FC<ILabeledProgressBar> = ({
         </div>
       </div>
       <ProgressBar {...progressBarProps} />
-    </div>
+    </>
   );
 };
--- a/src/pages/Xvs/Header/index.tsx
+++ b/src/pages/Xvs/Header/index.tsx
@@ -1,11 +1,13 @@
 /** @jsxImportSource @emotion/react */
-import React, { useMemo } from 'react';
+import React, { useContext, useMemo } from 'react';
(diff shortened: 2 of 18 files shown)


src/components/v2/ProgressBar/LabeledProgressBar/index.tsx
```diff
@@ -9,19 +10,21 @@ interface ILabeledProgressBar extends IProgressBarProps {
   whiteLeftText?: string;
   greyRightText?: string;
   whiteRightText?: string;
+  customCss?: SerializedStyles;
```
therealemjy: Can we not use `className` like we have so far with other components?

src/pages/Xvs/Header/index.tsx
```diff
@@ -1,13 +1,107 @@
...
+  venusVaiVaultRate,
+  totalXvsDistributedWei,
+}) => {
+  const styles = useStyles();
+  const xvsAddress = getToken('xvs').address;
+  const copyAddress = () => copy(xvsAddress);
+  const { t } = useTranslation();
+
+  const readableDailyDistribution = useMemo(() => {
+    const dailyDistribution = dailyVenus.div(new BigNumber(10).pow(18)).plus(venusVaiVaultRate);
```
therealemjy: Let's use `getToken` to get the token decimals rather than inputting `18` directly

src/pages/Xvs/constants.ts
```diff
@@ -0,0 +1,2 @@
+// @TODO magic number
+export const MINTED_XVS_WEI = 23700000;
```
therealemjy: It looks like this value is expressed in tokens (23.7 million XVS) rather than in wei.
Converting this value to wei should fix the progress bar which displays full:
Screen Shot 2022-05-17 at 15 53 03

src/pages/Xvs/Header/index.tsx
```diff
@@ -1,13 +1,107 @@
...
+
+  return (
+    <Paper className={className} css={styles.headerRoot}>
+      <EllipseText css={styles.address} text={xvsAddress}>
+        <div css={styles.xvsIconContainer}>
+          <Icon name="xvs" size={styles.iconSize} />
+        </div>
+        <Typography
+          className="ellipse-text"
+          href={generateBscScanUrl('xvs')}
```
therealemjy, line 74:
```suggestion
          href={generateBscScanUrl('xvs', 'token')}
```

therealemjy (approved):

therealemjy (changes requested):
Mistakenly approved sorry

src/components/v2/ProgressBar/LabeledProgressBar/index.tsx
coreyar: I am not passing this to the parent which is why it needed to be passed as serialized styles.

src/pages/Xvs/constants.ts
coreyar: I think there is something else going on there that I need to follow up on when I wire everything up. It was originally set in tokens and I wanted to use it in WEI so I renamed the variable but the value still needs to be verified hence the TODO

therealemjy: Ah fair enough, fine for me then 👍

src/components/v2/ProgressBar/LabeledProgressBar/index.tsx
therealemjy: Makes sense. I saw you changed the structure of the component and now this is passed through the `className` prop since it's applied to the root div 👍

therealemjy (approved):

PR #493, opened 2022-05-16
Add XVS Table UI

coreyar:
Adds XVS Table UI
Includes updates to fetching markets to sum total daily XVS

### DIFF
--- a/src/clients/api/queries/useUserMarketInfo.ts
+++ b/src/clients/api/queries/useUserMarketInfo.ts
@@ -147,6 +147,9 @@ const useUserMarketInfo = ({
         collateral,
         percentOfLimit,
         hypotheticalLiquidity: ['0', '0', '0'] as [string, string, string],
+        xvsPerDay: new BigNumber(market.supplierDailyVenus)
+          .plus(new BigNumber(market.borrowerDailyVenus))
+          .div(new BigNumber(10).pow(getToken('xvs').decimals)),
       };
       // user totals
       const borrowBalanceUsdCents = asset.borrowBalance.times(asset.tokenPrice).times(100);
@@ -166,6 +169,8 @@ const useUserMarketInfo = ({
       );
       acc.treasuryTotalAvailableLiquidityUsdBalanceCents =
         acc.treasuryTotalAvailableLiquidityUsdBalanceCents.plus(asset.liquidity.times(100));
+      // total distributed Xvs
+      acc.totalXvsDistributed = acc.totalXvsDistributed.plus(market.totalDistributed);
 
       acc.totalXvsDistributedWei = acc.totalXvsDistributedWei.plus(
         new BigNumber(market.totalDistributed).times(
--- a/src/clients/api/queries/useUserMarketInfo.ts
+++ b/src/clients/api/queries/useUserMarketInfo.ts
@@ -170,8 +170,6 @@ const useUserMarketInfo = ({
       acc.treasuryTotalAvailableLiquidityUsdBalanceCents =
         acc.treasuryTotalAvailableLiquidityUsdBalanceCents.plus(asset.liquidity.times(100));
       // total distributed Xvs
-      acc.totalXvsDistributed = acc.totalXvsDistributed.plus(market.totalDistributed);
-
       acc.totalXvsDistributedWei = acc.totalXvsDistributedWei.plus(
         new BigNumber(market.totalDistributed).times(
           new BigNumber(10).pow(getToken('xvs').decimals),
--- a/src/context/MarketContext.tsx
+++ b/src/context/MarketContext.tsx
@@ -192,6 +192,9 @@ const MarketContextProvider = ({ children }: $TSFixMe) => {
             collateral,
             percentOfLimit,
             hypotheticalLiquidity: ['0', '0', '0'] as [string, string, string],
+            xvsPerDay: new BigNumber(market.supplierDailyVenus)
+              .plus(new BigNumber(market.borrowerDailyVenus))
+              .div(new BigNumber(10).pow(getToken('xvs').decimals)),
           };
         });
 
--- a/src/pages/Xvs/Table/index.tsx
+++ b/src/pages/Xvs/Table/index.tsx
@@ -1,17 +1,30 @@
 /** @jsxImportSource @emotion/react */
 import React, { useContext, useMemo } from 'react';
+import BigNumber from 'bignumber.js';
 import { useHistory } from 'react-router-dom';
 import { Typography } from '@mui/material';
-import { useUserMarketInfo } from 'clients/api';
+import {
+  useUserMarketInfo,
+  useGetVenusVaiVaultRate,
+  useGetBalanceOf,
+  useGetMarkets,
+} from 'clients/api';
 import { Token, Table, ITableProps } from 'components';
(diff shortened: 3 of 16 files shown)


src/clients/api/queries/getMarkets.ts
```diff
@@ -20,19 +21,20 @@ const getMarkets = async (): Promise<GetMarketsOutput> => {
...
+      .map(item => {
+        return response.data?.data.markets.find(
           (market: Market) => market.underlyingSymbol.toLowerCase() === item.toLowerCase(),
         );
-      }
-      return undefined;
-    })
-    .filter(notUndefined);
-  return data;
+        return undefined;
```
therealemjy: This code is unreachable since you return the previous line already

src/clients/api/queries/useUserMarketInfo.ts
```diff
@@ -111,7 +116,7 @@ const useUserMarketInfo = ({
           isEnabled = toDecimalAmount(wallet.tokenAllowance).isGreaterThan(walletBalance);
         }
       }
-
+      console.log(market.supplierDailyVenus, market.borrowerDailyVenus);
```
therealemjy: To be removed :)

src/clients/api/queries/useUserMarketInfo.ts
```diff
@@ -142,6 +147,9 @@ const useUserMarketInfo = ({
         collateral,
         percentOfLimit,
         hypotheticalLiquidity: ['0', '0', '0'] as [string, string, string],
+        xvsPerDay: new BigNumber(market.supplierDailyVenus)
+          .plus(new BigNumber(market.borrowerDailyVenus))
+          .div(1e18),
```
therealemjy: Just so we don't have constants in our code, let's get the XVS token and use its decimals instead:
```
.div(new BigNumber(10).pow(getToken('xvs').decimals))
```

src/context/MarketContext.tsx
```diff
@@ -192,6 +192,9 @@ const MarketContextProvider = ({ children }: $TSFixMe) => {
             collateral,
             percentOfLimit,
             hypotheticalLiquidity: ['0', '0', '0'] as [string, string, string],
+            xvsPerDay: new BigNumber(market.supplierDailyVenus)
+              .plus(new BigNumber(market.borrowerDailyVenus))
+              .div(1e18),
```
therealemjy: Same thing here about getting XVS and using its decimals

src/pages/Xvs/Table/index.tsx
```diff
@@ -1,9 +1,103 @@
...
+      key: 'asset',
+      render: () => <Token symbol={asset.symbol as TokenId} />,
+      value: asset.id,
+    },
+    {
+      key: 'xvsPerDay',
+      render: () => (
+        <Typography variant="small1" css={styles.whiteText}>
+          {formatCoinsToReadableValue({
+            value: asset.xvsPerDay.dp(2, 1),
```
therealemjy: `formatCoinsToReadableValue` already takes care of the decimal places.
I guess you did that to follow the designs, but they haven't been updated following some of the updates we've made regarding value formatting so I'd rather stay consistent with the rest of the app and use the `shorthand` option here.

src/pages/Xvs/Table/index.tsx
```diff
@@ -1,9 +1,103 @@
...
+        </Typography>
+      ),
+      value: asset.xvsPerDay.toFixed(),
+    },
+    {
+      key: 'supplyApy',
+      render: () => (
+        <Typography variant="small1" css={styles.whiteText}>
+          {formatToReadablePercentage(
+            asset.xvsSupplyApy.isLessThan(0.01) ? new BigNumber(0.01) : asset.xvsSupplyApy,
```
therealemjy: I'd avoid doing that, even if it means we'll sometimes display `0`. I think it's less misleading than displaying `0.01` when `xvsSupplyApy` is actually 0 for example

src/pages/Xvs/Table/index.tsx
```diff
@@ -1,9 +1,103 @@
...
+        </Typography>
+      ),
+      value: asset.xvsSupplyApy.toFixed(),
+    },
+    {
+      key: 'borrowApy',
+      render: () => (
+        <Typography variant="small1" css={styles.whiteText}>
+          {formatToReadablePercentage(
+            asset.xvsBorrowApy.isLessThan(0.01) ? new BigNumber(0.01) : asset.xvsBorrowApy,
```
therealemjy: Same thing here about not applying a default value

src/pages/Xvs/constants.ts
```diff
@@ -0,0 +1,6 @@
+import BigNumber from 'bignumber.js';
+import { getToken } from 'utilities';
+
+export const XVS_DECIMALS = getToken('xvs').decimals;
+export const MINTED_XVS = '23700000';
```
therealemjy, line 5: I think this constant should be made global to the app, in case other pages need it too since this is true for the entire app.

therealemjy (commented):
I wrote in Slack that we seemed to have lost quite a lot of fixes that were done on the `Table` component and as a result this page is affected too:
Screen Shot 2022-05-17 at 08 43 05

- the column headers shouldn't be uppercased
- the texts inside cells shouldn't be bold

src/pages/Xvs/Table/index.tsx
coreyar: Ok, This came from the current site

src/pages/Xvs/constants.ts
coreyar: There is an open thread with @kkirka about where this value came from. I am not sure if it is relevant to the other pages of the app and we could move it at that time if it is. First, I'd like to understand where it came from.
https://github.com/VenusProtocol/venus-protocol-interface/blob/develop/src/containers/Main/XVS/index.tsx#L199

therealemjy: Got you 👍

src/pages/Xvs/Table/index.tsx
```diff
@@ -1,9 +1,99 @@
...
+  const rows: ITableProps['data'] = assets.map(asset => [
+    {
+      key: 'asset',
+      render: () => <Token symbol={asset.symbol as TokenId} />,
+      value: asset.id,
+    },
+    {
+      key: 'xvsPerDay',
+      render: () => (
+        <Typography variant="small1" css={[styles.whiteText, styles.fontWeight400]}>
```
therealemjy, line 58: It doesn't look like `fontWeight400` is getting applied:
Screen Shot 2022-05-17 at 15 48 07

Seeing all the tables display their cell content with a font-weight of 400, I think we should update the table component directly so it applies globally

coreyar: I don't believe that you have the latest code because I see it applied. Applying it to the table would not be correct because we cannot predict what styles each column will require. The 600 font weight it is applied in the column render function as you can see so customizing those styles belongs here where they base style was applied.

therealemjy: Yes you're right, I didn't have your last commit. Sorry for that.
Looks good to me!

therealemjy (approved):

therealemjy (approved):

PR #499, opened 2022-05-17
Market Details page wire up - Supply/Borrow info

therealemjy:
- add `useGetMarketHistory` query hook to API client
- redirect user to market page if `vTokenId` passed through route params is invalid
- fetch and display supply and borrow info on `MarketDetails` page


.eslintrc.js
```diff
@@ -72,6 +72,7 @@ module.exports = {
     '@typescript-eslint/explicit-module-boundary-types': 0,
     'react/jsx-wrap-multilines': 0,
     'generator-star-spacing': 0,
+    'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
```
coreyar: This rule is configured on line 44

src/pages/App.tsx
```diff
@@ -72,10 +73,25 @@ const App = () => (
...
+                              render={props => {
+                                // Redirect to market page if vTokenId is
+                                // invalid
+                                if (
+                                  !Object.prototype.hasOwnProperty.call(
+                                    VBEP_TOKENS,
+                                    props.match.params.vTokenId,
+                                  )
+                                ) {
+                                  return <Redirect to="/market" />;
```
coreyar: I think Redirect logic is better organized in the MarketDetails page

src/pages/MarketDetails/Card/index.tsx
```diff
@@ -16,13 +16,20 @@ export interface ICardProps {
   legends?: ILegend[];
   stats?: IStat[];
   className?: string;
+  'data-testid'?: string;
```
coreyar: nit: I think the alternative of keeping all props camelCased and using them ass attributes inline is preferable

src/pages/MarketDetails/index.spec.tsx
```diff
@@ -1,9 +1,122 @@
...
+  supplyVenusApy: '1.14',
+  exchangeRate: '0.00000000020011429546',
+  priceUSD: '10',
+  totalBorrow: '1000',
+  totalSupply: '10000',
+  createdAt: '2022-05-17T05:44:01.000Z',
+  updatedAt: '2022-05-17T06:00:06.000Z',
+};
+
+const fakeMarket: Market = {
```
coreyar: There are market mocks to help avoid recreating this data for every test

src/pages/MarketDetails/index.spec.tsx
```diff
@@ -1,9 +1,122 @@
...
+import { waitFor } from '@testing-library/react';
+
+import { Market, MarketSnapshot } from 'types';
+import { getMarkets, getMarketHistory } from 'clients/api';
 import renderComponent from 'testUtils/renderComponent';
 import MarketDetails from '.';
 
+const fakeVTokenId = 'aave';
+
+const fakeMarketSnapshot: MarketSnapshot = {
```
coreyar: We can we add some marketSnapshot to the mocks so they can be used in all tests

src/pages/MarketDetails/index.spec.tsx
```diff
@@ -1,9 +1,122 @@
...
+          },
+          isExact: true,
+          path: '/:vTokenId',
+          url: '',
+        }}
+      />,
+    );
+
+    // Check supply info displays correctly
+    await waitFor(() => expect(getByTestId('market-details-supply-info')).toMatchSnapshot());
```
coreyar, line 57: These tests are going to be really brittle. I think it is better to check that the correct values are visible because the test breaking will be more indicative of an important issue.

therealemjy: That's what these tests check. Checking the values one by one wouldn't make the tests stronger in this case, because each of the values are formatted differently so changing any of the formatting functions would break the tests the same.

.eslintrc.js
therealemjy: Ah right, I forgot to rebase. Thanks

src/pages/App.tsx
therealemjy: I don't think that's ideal. The router should be responsible of handling redirections, while the page should only care about rendering its content

src/pages/MarketDetails/Card/index.tsx
therealemjy: Works for me

src/pages/MarketDetails/index.spec.tsx
therealemjy: Good idea👍

src/pages/MarketDetails/index.spec.tsx
coreyar: ahh I see you are snapshotting specific elements. Looks good!

src/pages/App.tsx
coreyar: I don't think that is always going to be possible. There will be business logic required at times where it won't make sense to lift that logic/ state to the router. This page is redirecting because it can't render its content because it received an invalid route but we may redirect for many other reasons from with components/ side effect containers.

Additionally, If we include redirect logic inline for every page in the router this component will unwieldily. At that point we'd like want a special component to redirect. 

I don't feel strongly about this unless this component becomes messy

therealemjy: Ok I see, your point about this file becoming messy makes sense to me.
I've made the update; the Market Details page now handle the redirection.
I also moved all the route paths to a global `Path` enum so that they are all defined from a single source of truth.

coreyar (approved):

PR #500, opened 2022-05-17
Fix: multiple row gaps for table card view

baomastr:


coreyar:
I don't see what this PR does by looking at the new story. The new story is identical to With Custom column width story

src/components/v2/Table/index.stories.tsx
```diff
@@ -38,6 +38,15 @@ const useStyles = () => {
         display: initial;
       }
     `,
+
+    /* multiple rows styles */
+
+    cardContentGrid: `
```
coreyar, line 44: This style should not be global to the table. It should be implemented by the consumer of the table component

baomastr: it was in previous iteration, but was updated last week. 
so how do you see the implementation of this?

coreyar: Table responsive styles are passed down to the table when implementing the table. This is done and you can see examples in the market tables

baomastr: the discussion is currently in opened issue: https://app.clickup.com/t/2amjufw

baomastr:
@coreyar @therealemjy the updates for Markets tables are ready for review

therealemjy:
Looks good. I pushed a small fix for the padding of the table card container.

therealemjy (approved):

PR #503, opened 2022-05-17
Restore table titles after refactoring

coreyar:
Restore table titles so table responsiveness is correct after refactoring


therealemjy:
Very nice! 👍

therealemjy (approved):

PR #506, opened 2022-05-18
Add support for Market Details page to Layout header

therealemjy:
This PR adds support for displaying a back button inside the `Layout` header when navigating to the Market Details page.


src/components/v2/Layout/Header/index.tsx
```diff
@@ -19,19 +24,35 @@ const Header = () => {
...
+        </Link>
+      );
+    }
+
     const currentItem = menuItems.find(item => item.href === pathname);
     const currentItemKey = currentItem?.i18nTitleKey || currentItem?.i18nKey;
+
     if (!currentItemKey) {
-      console.error(`missed key for translation: ${pathname}`);
+      console.error(`Missing translation for key: ${pathname}`);
```
coreyar: We could leave a comment here to add error reporting which would help us in the future when we add better error reporting

src/components/v2/Layout/Header/index.tsx
```diff
@@ -19,19 +24,35 @@ const Header = () => {
   const isOnMobile = useIsMdDown();
 
   const title = useMemo(() => {
+    // Handle special case of Market Details page
```
coreyar: If we write this as a component not in the render scope we wouldn't need to memoize it.

src/components/v2/Layout/Header/index.tsx
therealemjy: Good idea.

src/components/v2/Layout/Header/index.tsx
therealemjy: Yeah.
There's a few improvements that can be done here. I'll update

coreyar (approved):

PR #507, opened 2022-05-18
Decouple logic from MarketDetails page container

therealemjy:
The file of the `MarketDetails` page is getting quite big, so before I add even more logic I thought I'd decouple the code a bit to make it easier to digest.

### DIFF
--- a/src/pages/MarketDetails/useGetMarketData.ts
+++ b/src/pages/MarketDetails/useGetMarketData.ts
@@ -14,8 +14,8 @@ const useGetMarketData = ({
   vTokenId: IVBepToken['id'];
   vTokenAddress: IVBepToken['address'];
 }) => {
-  const { data: markets = [] } = useGetMarkets({ placeholderData: [] });
-  const assetMarket = markets.find(
+  const { data: getMarketData } = useGetMarkets();
+  const assetMarket = (getMarketData?.markets || []).find(
     market => market.address.toLowerCase() === vTokenAddress.toLowerCase(),
   );
 
--- a/src/pages/MarketDetails/useGetMarketData.ts
+++ b/src/pages/MarketDetails/useGetMarketData.ts
@@ -1,0 +1,113 @@
+import React from 'react';
+import BigNumber from 'bignumber.js';
+
+import { IVBepToken } from 'types';
+import { getToken } from 'utilities';
+import { convertWeiToCoins } from 'utilities/common';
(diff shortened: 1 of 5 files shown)


src/pages/MarketDetails/useGetMarketData.ts
```diff
@@ -0,0 +1,113 @@
...
+import { useGetMarkets } from 'clients/api';
+
+const useGetMarketData = ({
+  vTokenId,
+  vTokenAddress,
+}: {
+  vTokenId: IVBepToken['id'];
+  vTokenAddress: IVBepToken['address'];
+}) => {
+  const { data: getMarketData } = useGetMarkets();
```
coreyar, line 17: I would put this in the hooks directory. Currently we merge markets and assets in the `useUserMarketHook`.

I think it would make sense to include logic for setting units in the `useGetMarkets` hook so we sanitize and format data the way we want to use it before it is ever used. 

How we handle / merge assets and markets could probably be further optimized and ties into an on going discussion with Lakshmi

therealemjy: This hook will only be relevant to this component, so I don't think it should be put in the `hooks` directory (since it won't be reused).

About setting the units in the `useGetMarkets` hook, for now I wouldn't spend time on it since this is bound to be replaced once we have the subgraph in place and we move away from using the backend at all.

Yeah I agree. For now I think we've been following the right approach and I'd continue that way: we use React Query to refactor the existing fetching functions, so that we benefit from caching and all the rest, but don't actually modify the data model until we are done with the UI refresh. What do you think?

coreyar: We have been making minor changes to the data model for example with units and adding to the userMarket hook.

I'm fine with leaving this as is for now but I think it will get refactored as you said and we should look at how useUserMarkets is accessing this same information.

therealemjy: Ok I see what you mean. We could update `useGetMarkets` so as you said it formats the data as we want and adds the units to the names, but that would imply refactoring existing parts of the app that rely on it and right now I think we should avoid that because it also opens the gates to more refactors (`useUserMarketInfo` for example).

Since this component is brand new, I wanted to follow the patterns we've established by adding the units to the variable names so that when we do refactor how we fetch the data we'll have this in place already.

coreyar (approved):

PR #508, opened 2022-05-18
Add pagination

baomastr:
https://app.clickup.com/t/2au9k3t


baomastr:
The styles import for pagination is broken. Gotta fix it

src/components/v2/Pagination/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+            variant="text"
+            css={[classes.button, classes.getButtonStyles({ isActive: index === activePageIndex })]}
+            key={page}
+            onClick={() => setCurrentPageIndex(index)}
+          >
+            {page}
+          </Button>
+        );
+      })}
+    </div>
```
therealemjy: That's a lot of duplicated code. Only some of the DOM changes for each of these variants so you could avoid a lot of that duplicating

src/components/v2/Pagination/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+    : activePageIndex + halfOfPagesCount;
+
+  if (pagesCount <= 1) {
+    return null;
+  }
+
+  return (
+    <div css={[classes.root, className]}>
+      {pagesArray.map((page, index) => {
+        const iconProps: IIconProps = { size: '20px', name: 'arrowRight', color: 'inherit' };
```
therealemjy: We've established we want to avoid using inline styles: you can set the size (width + height) from the styles file.

src/components/v2/Pagination/styles.ts
```diff
@@ -0,0 +1,42 @@
...
+    getButtonStyles: ({ isActive }: { isActive: boolean }) => css`
+      color: ${isActive ? theme.palette.text.primary : theme.palette.text.secondary};
+    `,
+    arrowButtonWrapper: css`
+      display: flex;
+      align-items: center;
+      line-height: 1px;
+    `,
+    iconReverted: css`
+      transform: rotate(180deg);
```
therealemjy, line 46: Let's create a `leftArrow` icon instead, this way we make sure it respects the designs' dimensions and placement

src/components/v2/Pagination/index.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <Table
+        columns={columns}
+        data={currentPageData}
+        title="Market Data"
+        minWidth="650px"
+        rowKeyIndex={0}
+        tableCss={styles.tableCss}
+        cardsCss={styles.cardsCss}
+        css={styles.table}
+      />
```
therealemjy: Just to avoid confusion, let's only include the `Pagination` component in the story.

src/components/v2/Pagination/index.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <Box display="flex" justifyContent="flex-end" alignItems="center" mt={4}>
+        <Typography mr={2}>{itemsCountString}</Typography>
+        <Pagination
+          activePageIndex={currentPageIndex}
+          pagesCount={pagesCount}
+          setCurrentPageIndex={setCurrentPageIndex}
+        />
+      </Box>
```
therealemjy: This should all be a single component, so that we can handle responsiveness from `Pagination` directly

therealemjy (commented):
Looks good.
Two things:
- you need to make the `Pagination` component responsive
Screen Shot 2022-05-24 at 11 42 49
- did you figure out a solution with Pavel about the component jumping when changing page?

src/components/v2/Pagination/styles.ts
baomastr: but that makes a one more file to load. the arrow in design is the same but rotated. why do you think the current realisation is not suitable?

src/components/v2/Pagination/index.stories.tsx
baomastr: but that's the way to see how it works with data because the hook usePagination is actually part of the component

baomastr: I can leave some comments in code for clarifying if this would solve the issue

src/components/v2/Pagination/index.stories.tsx
baomastr: done

src/components/v2/Pagination/index.tsx
baomastr: there are props, but not inline styles

baomastr: and if we want to avoid the `size` prop, we should remove it from `IIconProps`

src/components/v2/Pagination/index.tsx
baomastr: done

src/components/v2/Pagination/usePagination.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  // const isFirstPage = currentPageIndex === 0;
+  const isLastPage = activePageIndex === pagesCount - 1;
+  const currentPageFirstIndex = activePageIndex * rowsPerPageCount;
+  const currentPageLastIndex = isLastPage ? data.length : currentPageFirstIndex + rowsPerPageCount;
+  const firstItemNumberInRow = currentPageFirstIndex + 1;
+  const isSingleItemOnPage = firstItemNumberInRow === currentPageLastIndex;
+  const itemsCountString = isSingleItemOnPage
+    ? `${currentPageLastIndex} item of ${data.length}`
+    : `${firstItemNumberInRow} - ${currentPageLastIndex} items of ${data.length}`;
```
therealemjy: Texts need to be internationalized

src/components/v2/Pagination/index.stories.tsx
therealemjy: `Pagination` doesn't need to be coupled to the table component. In the future we might want to use pagination without rendering a table, and so the component should be more modular than that.
Pagination is also handled by the backend: its goal is so that we don't have to load all of the data, but we can load it page by page. In your current implementation, the full data would be loaded then divided into pages which will present performance issues.

Perhaps the specs needed to be more explicit, I didn't think it would cause issues.
I'll update the ticket then.

therealemjy: I updated the ticket with more specs. Let me know if it's still unclear

src/components/v2/Pagination/index.tsx
therealemjy: We can yeah

src/components/v2/Pagination/styles.ts
therealemjy: Ok I see. In some instances, the icon could be placed differently in the viewbox depending on the direction of the arrow. If you're sure the left arrow corresponds exactly to the right right with a 180 degree rotation, let's go for it.

src/components/v2/Pagination/index.stories.tsx
baomastr: done

src/components/v2/Pagination/usePagination.ts
baomastr: done

src/components/v2/Pagination/index.tsx
baomastr: we can do it as separate feature task, i think. because there are a lot of places which will affect other components:

image

baomastr:
moved to https://github.com/VenusProtocol/venus-protocol-interface/pull/547

PR #509, opened 2022-05-18
Dynamically set svg ids

coreyar:
- Transforms svg files to react components using svgr
`yarn svgr -- src/components/v2/Icon/icons --out-dir src/components/v2/Icon/icons  --filename-case camel --typescript --no-dimensions --no-index`

- dynamically set id on svg via a ref
- use svg from assets in css


src/components/v2/Icon/index.tsx
```diff
@@ -13,6 +14,11 @@ export interface IIconProps {
 
 export const Icon: React.FC<IIconProps> = ({ name, size, color, ...otherProps }) => {
   const theme = useTheme();
+  const idRef = useRef<string>();
+
+  if (idRef.current === undefined) {
+    idRef.current = uniqueId();
+  }
```
therealemjy: You can directly initialize the red with the ID:
```suggestion
const idRef = useRef<string>(uniqueId());
```

And so seeing we apply this at the component level, in case we'd render SVGs that don't fall in the icon category should we create a `Svg` component (or rename `Icon` to `Svg`)?

therealemjy:
That's nice, I like the approach 👍
I just had one question regarding rendering other SVGs that aren't icons, but other than that it looks good to me.

src/components/v2/Icon/index.tsx
coreyar: I don't have any strong opinions about the name. I imagine if it is not an icon the underlying file will not always be an SVG and it might be applied using the `img` element or css

therealemjy: Yeah, but if we call it `Svg` I think we can assume only SVG will be accepted.
We can leave this as it is, then the day we need to render SVGs that aren't icons (for this for example: https://www.figma.com/file/GYJwWWLbC4B8DrJOPcETz3/Venus-2.0?node-id=3021%3A37178) we can think of creating an `Svg` component.

therealemjy (approved):

therealemjy (approved):

PR #512, opened 2022-05-19
Add action to report code coverage

coreyar:
Adds test coverage comparing diff in coverage between base branch and PR branch. 
[Action](https://github.com/ArtiomTr/jest-coverage-report-action)

Adds exporting coverage to a file from current test run.
Skips steps in action to install and run tests
Action will still run tests against base branch and post results/ diff to PR.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -88,6 +88,8 @@ jobs:
 
   test:
     runs-on: ubuntu-20.04
+    container:
+      image: node:14.15.1-buster
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


therealemjy (commented):
Nice :) Yeah I personally prefer that to running tests on every commit. We still get the information, but it doesn't come in the way of the development flow.

coreyar:
> Nice :) Yeah I personally prefer that to running tests on every commit. We still get the information, but it doesn't come in the way of the development flow.

therealemjy (approved):

PR #515, opened 2022-05-19
Prepare 0.4.0

therealemjy:
- bump version to 0.4.0
- remove top banner
- disable supply tab for LUNA and UST
- disable borrow tab for LUNA and UST
- prevent `Table` component from overflowing

### DIFF
--- a/src/pages/Dashboard/Modals/SupplyWithdraw/index.spec.tsx
+++ b/src/pages/Dashboard/Modals/SupplyWithdraw/index.spec.tsx
@@ -3,6 +3,7 @@ import BigNumber from 'bignumber.js';
 import { act, fireEvent, waitFor } from '@testing-library/react';
 
 import fakeTransactionReceipt from '__mocks__/models/transactionReceipt';
+import { DISABLED_TOKENS } from 'utilities';
 import fakeAccountAddress from '__mocks__/models/address';
 import { assetData } from '__mocks__/models/asset';
 import renderComponent from 'testUtils/renderComponent';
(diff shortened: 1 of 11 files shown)


src/pages/Dashboard/Modals/SupplyWithdraw/index.tsx
```diff
@@ -201,6 +187,24 @@ export const SupplyWithdrawUi: React.FC<ISupplyWithdrawUiProps & ISupplyWithdraw
     },
   ];
 
+  // Prevent user from being able to supply UST or LUNA
+  if (assetId !== 'ust' && assetId !== 'luna') {
```
coreyar: I took a slightly different approach here - https://github.com/VenusProtocol/venus-protocol-interface/pull/514

We don't have to take it but the idea was to start setting up feature flags so in the future we don't only have to add tokens to the disabled token list we don't have to change other application code too

therealemjy: Sorry for not seeing you opened a PR for that already.

I updated this PR to follow the same logic as yours :)

src/pages/Dashboard/Modals/SupplyWithdraw/index.spec.tsx
```diff
@@ -83,6 +83,33 @@ describe('pages/Dashboard/SupplyWithdrawUi', () => {
   });
 
   describe('Supply form', () => {
+    it.each(['ust', 'luna'])('does not display supply tab when asset is %s', async tokenId => {
```
coreyar: nit: We could import `DISABLED_TOKENS` here to make this test more resilient

coreyar (approved):

src/pages/Dashboard/Modals/SupplyWithdraw/index.spec.tsx
therealemjy: Ah yeah of course, good point

coreyar (approved):

PR #516, opened 2022-05-19
Check transaction events for errors

coreyar:
Check and throw errors if a transaction event fails.
Errors are decoded using enums declared in the contracts - https://github.com/VenusProtocol/venus-protocol/blob/develop/contracts/ErrorReporter.sol
We'll need to update our errors when this enum changes because it is indexed by order


src/utilities/errors.ts
```diff
@@ -21,3 +31,53 @@ export class InternalError extends Error {
...
+
+export const checkForTokenTransactionError = (receipt: TransactionReceipt) =>
+  checkForTransactionError(receipt, TokenErrorReporterError, TokenErrorReporterFailureInfo);
+
+export const checkForVaiControllerTransactionError = (receipt: TransactionReceipt) =>
+  checkForTransactionError(
+    receipt,
+    VAIControllerErrorReporterError,
+    VAIControllerErrorReporterFailureInfo,
+  );
```
therealemjy, line 83: Why not use `checkForTransactionError` directly if only the type changes?

therealemjy (approved):
Really nice work 👍
Just one minor comment

src/utilities/errors.ts
coreyar: It is actually 2 and could be 3. The error is returned with 3 arguments that differ by contract `error`, `info` and `detail`. Rather have import 3 or four and risk mixing them up when using `checkForTransactionError` this seemed more convenient

therealemjy: Ok got you 👍

PR #519, opened 2022-05-20
Approve VAI with Unitroller as spender

coreyar:
We were enabling VAI with the token contract as the spender but it needs to be enabled with the unitroller as spender


therealemjy (approved):
Good catch 👍

PR #522, opened 2022-05-20
Add tooltip explaining minted VAI is included in borrow limit used

coreyar:
This will help a user understand why this percentage is different than the sum of the borrowed percentages in the borrowed market

### DIFF
--- a/src/components/v2/ProgressBar/AccountHealth/index.tsx
+++ b/src/components/v2/ProgressBar/AccountHealth/index.tsx
@@ -5,7 +5,10 @@ import PLACEHOLDER_KEY from 'constants/placeholderKey';
 import { formatCentsToReadableValue, formatToReadablePercentage } from 'utilities/common';
 import calculatePercentage from 'utilities/calculatePercentage';
 import { useTranslation } from 'translation';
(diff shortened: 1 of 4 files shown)


src/components/v2/ProgressBar/AccountHealth/index.tsx
```diff
@@ -70,7 +74,14 @@ export const AccountHealth: React.FC<IAccountHealthProps> = ({
...
         }
         whiteLeftText={
-          variant === 'borrowBalance' ? readableBorrowBalance : readableBorrowLimitUsedPercentage
+          <span css={styles.textWithTooltip}>
+            {variant === 'borrowBalance'
+              ? readableBorrowBalance
+              : readableBorrowLimitUsedPercentage}
+            <Tooltip css={styles.tooltip} title={t('myAccount.includingMintedVai')}>
+              <Icon css={styles.infoIcon} name="info" />
+            </Tooltip>
```
therealemjy, line 83: The tooltip should only be displayed when the variant isn't "borrowBalance"

therealemjy (approved):

PR #525, opened 2022-05-23
Show negative net APY in red

coreyar:
When a user's net APY is negative, we display it in red.

### DIFF
--- a/src/pages/Dashboard/MyAccount/MyAccountUi.tsx
+++ b/src/pages/Dashboard/MyAccount/MyAccountUi.tsx
@@ -93,7 +93,15 @@ export const MyAccountUi = ({
           </Tooltip>
         </div>
 
-        <Typography variant="h1" color="interactive.success" component="span">
+        <Typography
+          variant="h1"
+          color={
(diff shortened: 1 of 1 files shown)


src/pages/Dashboard/MyAccount/MyAccountUi.tsx
```diff
@@ -93,7 +93,13 @@ export const MyAccountUi = ({
           </Tooltip>
         </div>
 
-        <Typography variant="h1" color="interactive.success" component="span">
+        <Typography
+          variant="h1"
+          color={
+            netApyPercentage && netApyPercentage > 0 ? 'interactive.success' : 'interactive.error'
```
therealemjy: I think we should display 0 values in green. That's the default value for new users who haven't connected their wallet or supplied/borrowed any tokens yet, so the green color would be a bit more welcoming.
```suggestion
            netApyPercentage && netApyPercentage < 0 ? 'interactive.error' : 'interactive.success'
```

therealemjy (approved):

PR #529, opened 2022-05-24
Add market detail navigation 

coreyar:
- Add market detail navigation 
- Add hover styles to clickable cards

### DIFF
--- a/src/components/v2/Table/styles.ts
+++ b/src/components/v2/Table/styles.ts
@@ -22,9 +22,16 @@ export const useStyles = () => {
         padding: 0;
       }
     `,
-    tableWrapperMobile: css`
+    tableWrapperMobile: ({ clickable }: { clickable: boolean }) => css`
       margin-top: ${theme.spacing(6)};
       padding: ${theme.spacing(4, 0)};
+      ${clickable &&
+      `
+        cursor: pointer;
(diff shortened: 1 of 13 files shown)


src/components/v2/Table/styles.ts
```diff
@@ -22,9 +22,16 @@ export const useStyles = () => {
...
+    tableWrapperMobile: ({ clickable }: { clickable: boolean }) => css`
       margin-top: ${theme.spacing(6)};
       padding: ${theme.spacing(4, 0)};
+      ${clickable &&
+      `
+        cursor: pointer;
+        :hover {
+          background-color: ${theme.palette.interactive.hover};
+        }
+      `}
```
therealemjy, line 34: If it's only relevant to mobile devices, we don't need any hover effect since there's no hover state on mobile (it would only briefly show on press)

coreyar: Cards are not exclusively shown on mobile or tables but are also displayed on smaller desktops.
I think this style should remain even so, because resizing the browser can cause cards to be displayed and we would have a gap in styles under that situation.

therealemjy: Got you

therealemjy (approved):

PR #530, opened 2022-05-24
Market Details screen - Wire up Interest Model chart

therealemjy:
- create `getVTokenApySimulations` query function and `useGetVTokenApySimulations` hook
- create `getVTokenCash` query function and `useGetVTokenCash` hook
- create `getVTokenInterestRateModel` query function and `useGetVTokenInterestRateModel` hook
- wire up interest rate model chart on Market Details screen


src/clients/api/queries/getVTokenApySimulations.ts
```diff
@@ -0,0 +1,86 @@
...
+  vTokenInterestModelContract,
+  vTokenReserveFactorMantissa,
+}: IGetVTokenInterestRatesInput): Promise<GetVTokenApySimulationsOutput> => {
+  const promises: Promise<VTokenApySnapshot>[] = [];
+
+  for (let u = 1; u <= 100; u++) {
+    const utilizationRate = u / 100;
+
+    const getRates = async () => {
+      const borrowRate = await vTokenInterestModelContract.methods
```
coreyar: I think this should be hook and each of these queries should be setup individually. This keeps our setup consistent in how we define and combine queries. It is a bit of boilerplate but I think it is an unavoidable consequence of following a pattern. If I recall correctly, we aren't currently following standard organization because these aren't being used in other places but we can't be certain of that in the future.

It also opens up disambiguity on where our layers lie, for example the layer where we fetch data, the layer were we validate and shape the data

src/clients/api/queries/getVTokenInterestRateModel.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+      call: callMock,
+    }));
+
+    const fakeContract = {
+      methods: {
+        interestRateModel: interestRateModelMock,
+      },
+    } as unknown as VBep20;
+
+    const response = await getInterestRateModel({
```
coreyar, line 40: This will need to be wrapped in our error new handling to check for the failure event

src/clients/api/queries/getVTokenCash.ts
```diff
@@ -0,0 +1,17 @@
...
+export interface IGetVTokenCashInput {
+  vTokenContract: VBep20 | VBnbToken;
+}
+
+export type GetVTokenCashOutput = BigNumber;
+
+const getVTokenCash = async ({
+  vTokenContract,
+}: IGetVTokenCashInput): Promise<GetVTokenCashOutput> => {
+  const res = await vTokenContract.methods.getCash().call();
```
coreyar, line 13: This will need to be wrapped in our error new handling to check for the failure event

src/clients/api/queries/getVTokenInterestRateModel.ts
```diff
@@ -0,0 +1,14 @@
...
+export interface IGetVTokenInterestRateModelInput {
+  vTokenContract: VBep20 | VBnbToken;
+}
+
+export type GetVTokenInterestRateModelOutput = string;
+
+const getVTokenInterestRateModel = async ({
+  vTokenContract,
+}: IGetVTokenInterestRateModelInput): Promise<GetVTokenInterestRateModelOutput> =>
+  vTokenContract.methods.interestRateModel().call();
```
coreyar, line 12: This will need to be wrapped in our error new handling to check for the failure event

src/clients/api/queries/useGetVTokenApySimulations.ts
```diff
@@ -0,0 +1,52 @@
...
+const useGetVTokenApySimulations = (
+  {
+    vTokenId,
+    vTokenReserveFactorMantissa,
+  }: { vTokenId: VTokenId; vTokenReserveFactorMantissa?: BigNumber },
+  options?: Options,
+) => {
+  const web3 = useWeb3();
+  const { data: interestRateModelContractAddress } = useGetVTokenInterestRateModel({ vTokenId });
+  const vTokenInterestModelContract = interestRateModelContractAddress
```
coreyar: We could add this as a hook and access it like other contracts

src/clients/api/queries/getVTokenApySimulations.ts
therealemjy: Doing that would make this hook a lot more complicated, because if some of these queries fail and we have to retry them then we'll have to track which ones failed and refetch these specifically. It gets even worse when we start thinking about cache invalidation.

The pattern we have in place is to cache every query that needs to be cached. In this case only the resulting data from fetching all the queries should be cached, not the individual queries.

src/clients/api/queries/useGetVTokenApySimulations.ts
therealemjy: There is a hook to get this contract instance, but in this case we can't use it: because it's a hook, it needs to be always rendered, however `interestRateModelContractAddress` will be undefined until the query has fetched successfully. This means we'd try to instantiate the contract with an `undefined` address, which isn't accepted by web3.

src/clients/api/queries/getVTokenCash.ts
therealemjy: That's a view function, there's no transaction receipt returned

src/clients/api/queries/getVTokenInterestRateModel.ts
therealemjy: That's a view function, there's no transaction receipt returned

src/clients/api/queries/getVTokenInterestRateModel.spec.ts
therealemjy: That's a view function, there's no transaction receipt returned

src/clients/api/queries/getVTokenApySimulations.ts
coreyar: What I'm suggesting tracks with the cache invalidation strategy. The keys are set in the hooks so if you want to group and invalidate queries together that can be configured in the hook

therealemjy: I see what you mean, that would solve the cache invalidation issue. I still think it isn't a good idea to separate all these queries into separate queries for the other reason I mentioned however.

coreyar: I think we gain a lot from keeping code organized in a logical way that aligns with layers of how data flows through our application. I think that setting these up as queries should give us more insight into failures not less.

therealemjy: I don't think we'd achieve what you'd expect, because it would lead to quite a messy code.
The query function would look something like that:
```
const useQueryFn = (uRates: number[]) =>
  useQueries(uRates.map(uRate => ({ queryKey: [QUERY_KEY_BASE, uR], queryFn: fetchUr })))
```

And `useQueryFn` would return an array of query results, which we'd then have to go through to find any potential errors and identify what queries need to be retried. We end up with a lot more complexity than if we treat all the queries as being only one single query.

coreyar: How is that different than troubleshooting the array of promises we have now? With queries we would get the added benefit of caching and graceful error handling, pending. I don't see disadvantages to always calling contracts the same way.

At minimum this should be a hook, We should differentiate between queries which is shaping to be the layer where data is ingested and validated and hooks where data is transformed. This is doing more than just querying and validating data so it should be organized with the hooks.

therealemjy: With the array of promises we currently have, if any of the promises throws an error then the whole process is cancelled. The first error is returned and it becomes easy to retry the request.

I'll make the change you're requesting and I think you'll see what I mean. I think it will give us more ground to discuss that.

And about separating the layer that fetches the data from hooks that shape it, this isn't the route we've taken so far. A lot of our query hooks format the data they receive from the contracts before returning. Look at `getXvsWithdrawableAmount` and `getVTokenBalancesAll` for exemple.

coreyar: > With the array of promises we currently have, if any of the promises throws an error then the whole process is cancelled. The first error is returned and it becomes easy to retry the request.

That makes sense, I'd think this would be something reconfigurable in `react-query`?

> And about separating the layer that fetches the data from hooks that shape it, this isn't the route we've taken so far. A lot of our query hooks format the data they receive from the contracts before returning. Look at getXvsWithdrawableAmount and getVTokenBalancesAll for exemple.

Yes, I agree we have a clear distinction currently, which is part of the motivation for the comment. To ask about / propose a definition for creating a distinction for those layers

therealemjy: > That makes sense, I'd think this would be something reconfigurable in react-query?

Not as far as I can see unfortunately. `useQueries` would be use in this case and it basically just returns the query responses in an array.

I'll look into this more to see if I can find a compromise that works. I understand why you want to use React Query in this case.

> Yes, I agree we have a clear distinction currently, which is part of the motivation for the comment. To ask about / propose a definition for creating a distinction for those layers

Sure, that's a good idea. I'm down to discuss this through messages on Slack tomorrow (so we can focus on the 0.4.0 release today) if you're free. Then we could document our conclusions in ClickUp.

coreyar: Works for me

coreyar (approved):

PR #535, opened 2022-05-25
Automatically scroll to the top of the page on route change

therealemjy:
Create a `ResetScrollOnRouteChange` component that wraps the root router so page is scrolled to the top on route change.


baomastr (approved):

src/components/v2/ResetScrollOnRouteChange/index.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { useEffect } from 'react';
+import { useLocation } from 'react-router';
+
+export const ResetScrollOnRouteChange: React.FC = ({ children }) => {
+  const location = useLocation();
+
+  useEffect(() => {
+    window.scrollTo(0, 0);
+  }, [location]);
```
baomastr, line 9: what about cases when we need to scroll down to any anchor that is placed like hash param? wouldn't this case affect the following effect?

therealemjy: We currently don't support this feature anywhere, but we'd probably have to handle this manually even without the `ResetScrollOnRouteChange` component in place

PR #536, opened 2022-05-25
Fix: toast with long messages

baomastr:
got an issue with toast displaying while using the interface.

fix for cases when notification message has a lot of text:
image

the way how it was displayed before fix:
image


therealemjy (approved):
Nice

PR #537, opened 2022-05-25
Feat: add VaultItemUi component

baomastr:


src/components/v2/VaultItemUi/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+type VaultAssetType = 'vai' | 'xvs' | 'vrt';
+
+export interface IVaultItemUiProps {
+  stakeTokenName: VaultAssetType;
+  earnTokenName: VaultAssetType;
+  rewardValue?: number;
+  stakingValue: number;
+  stakingAprValue: number;
+  dailyEmissionValue: number;
+  totalStakedValue: number;
```
therealemjy: To be consistent with the rest of the app:
- values should be suffixed with their unit
- wei values need to use `BigNumber`
- I think `stakingValue` should be more descriptive to explain that it corresponds to the amount staked by the user

```suggestion
  rewardWei?: BigNumber;
  userStakedWei: BigNumber;
  stakingAprPercentage: number;
  dailyEmissionWei: BigNumber;
  totalStakedWei: BigNumber;
```

src/components/v2/VaultItemUi/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { formatCommaThousandsPeriodDecimal, formatToReadablePercentage } from 'utilities/common';
+import { Icon } from '../Icon';
+import { Button } from '../Button';
+import { useStyles } from './styles';
+
+type VaultAssetType = 'vai' | 'xvs' | 'vrt';
+
+export interface IVaultItemUiProps {
+  stakeTokenName: VaultAssetType;
+  earnTokenName: VaultAssetType;
```
therealemjy: We should use the token IDs here (we don't need to restrict the possible values to just the three we're currently displaying in the desigsns since in the future we might support more vaults with different tokens):
```suggestion
tokenId: TokenId;
rewardTokenId: TokenId;
```

src/components/v2/VaultItemUi/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+    [stakeTokenName, earnTokenName, stakingAprValue, dailyEmissionValue, totalStakedValue],
+  );
+
+  return (
+    <Paper css={classes.container} className={className}>
+      <div css={classes.header}>
+        <div css={classes.title}>
+          <Icon css={classes.tokenIcon} name={stakeTokenName} />
+          <Typography variant="h4" css={classes.text}>
+            {stakeTokenName.toUpperCase()}
```
therealemjy: We should use `getToken` with the token ID passed through props (see previous comment) so that we can retrieve the token data and render its symbol here.

therealemjy (commented):
I spotted a small issue on mobile where the top row content breaks a bit:
Screen Shot 2022-05-25 at 14 57 36

Let's change "Available reward" for "Reward", that should solve the issue.

src/components/v2/VaultItemUi/index.tsx
baomastr: done

src/components/v2/VaultItemUi/index.tsx
baomastr: done

src/components/v2/VaultItemUi/index.tsx
baomastr: done

baomastr:
new merge request is opened for this feature: https://github.com/VenusProtocol/venus-protocol-interface/pull/545

PR #544, opened 2022-05-25
Prepare version 0.5.0

coreyar:
- Bump dev version to 0.5.0
- Remove old VRTConversion page


therealemjy (approved):
Nice 👍
I propose we try to release this on Monday (we'll need approval from Sergio and Brad of course). What do you think?

PR #545, opened 2022-05-26
Feature: Add Vault item UI component

baomastr:


src/components/v2/VaultItemUi/styles.ts
```diff
@@ -0,0 +1,112 @@
...
+      justify-content: space-between;
+      margin-bottom: ${theme.spacing(6)};
+    `,
+    title: css`
+      display: flex;
+      align-items: center;
+    `,
+    tokenIcon: css`
+      width: 20px;
+      height: 20px;
```
therealemjy, line 23: We should use `theme.iconSize` for these

src/components/v2/VaultItemUi/styles.ts
```diff
@@ -0,0 +1,112 @@
...
+      margin-right: ${theme.spacing(1)};
+    `,
+    tokenIconLarge: css`
+      width: 32px;
+      height: 32px;
```
therealemjy, line 28: We should use `theme.iconSize` for these

src/components/v2/VaultItemUi/styles.ts
```diff
@@ -0,0 +1,112 @@
...
+    textAligned: css`
+      display: inline-flex;
+      align-items: center;
+    `,
+    rewardWrapper: css`
+      display: flex;
+      align-items: center;
+
+      ${theme.breakpoints.down('sm')} {
+        font-size: 14px;
```
therealemjy, line 57: We can get this font size from the typography variants in the theme (`small1` or `small2` would work).

src/components/v2/VaultItemUi/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+      {
+        title: t('vaultItemUi.stakingApr', { stakeTokenName: getToken(tokenId).symbol }),
+        value: formatToReadablePercentage(stakingAprPercentage),
+      },
+      {
+        title: t('vaultItemUi.dailyEmission'),
+        value: (
+          <>
+            <Icon css={classes.tokenIcon} name={rewardTokenId} />
+            {formatCommaThousandsPeriodDecimal(
```
therealemjy, line 61: You don't need to use `formatCommaThousandsPeriodDecimal` here, `convertWeiToCoins` already uses it internally.

src/components/v2/VaultItemUi/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+            )}
+          </>
+        ),
+      },
+      {
+        title: t('vaultItemUi.totalStaked'),
+        value: (
+          <>
+            <Icon css={classes.tokenIcon} name={tokenId} />
+            {formatCommaThousandsPeriodDecimal(
```
therealemjy, line 76: You don't need to use `formatCommaThousandsPeriodDecimal` here, `convertWeiToCoins` already uses it internally.

therealemjy (commented):
Nice, it looks good 👍
A few more comments and we should be good

src/components/v2/VaultItemUi/styles.ts
baomastr: this component has custom sizes that don't fit our typography styles.

src/components/v2/VaultItemUi/styles.ts
baomastr: done

src/components/v2/VaultItemUi/styles.ts
baomastr: done

baomastr:
moved to https://github.com/VenusProtocol/venus-protocol-interface/pull/548 due to rebase conflicts

PR #549, opened 2022-05-26
Feature: Add pagination

baomastr:


src/components/v2/Pagination/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+    minPageIndexToShow,
+    maxPageIndexToShow,
+  } = usePagination({
+    itemsCount,
+    onChange,
+    initialPageNumber,
+    itemsPerPageCount,
+  });
+
+  const classes = useStyles();
```
therealemjy: To be consistent with the rest of the codebase:
```suggestion
  const styles = useStyles();
```

src/components/v2/Pagination/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  if (pagesCount <= 1) {
+    return null;
+  }
+
+  const iconProps: IIconProps = { size: '25', name: 'arrowRight', color: 'inherit' };
+
+  return (
+    <div css={[classes.root, className]}>
```
therealemjy: `className` needs to be passed as a `className`:
```suggestion
    <div css={classes.root} className={className}>
```

src/components/v2/Pagination/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+    itemsPerPageCount,
+  });
+
+  const classes = useStyles();
+
+  if (pagesCount <= 1) {
+    return null;
+  }
+
+  const iconProps: IIconProps = { size: '25', name: 'arrowRight', color: 'inherit' };
```
therealemjy: As discussed previously (I think on another PR), the size needs to be defined through CSS

src/components/v2/Pagination/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  return (
+    <div css={[classes.root, className]}>
+      {itemsCountString && (
+        <Typography css={classes.itemsCountString}>{itemsCountString}</Typography>
```
therealemjy: As we discussed during our call, `itemsCountString` should be defined by this component directly

src/translation/translations/en.json
```diff
@@ -529,5 +529,9 @@
     "dailyDistribution": "Daily Distribution:",
     "progressBar": "Xvs Distribution progress",
     "remaining": "Remaining:"
+  },
+  "pagination": {
+    "itemOf": "item {{currentPageLastIndex}} out of {{itemsCount}}",
+    "itemsOf": "items {{firstItemNumber}} - {{currentPageLastIndex}} out of {{itemsCount}}"
```
therealemjy, line 545: @coreyar I'd like your opinion on this one. I asked Dmitriy to write the content this way instead of like in the designs.
The designs write "25 - 36 items of 150" but this seems incorrect to me in English. I suggested Dmitriy to write "items 25 - 36 out of 150" instead, but I'm also not 100% sure on that.
What do you think?

src/components/v2/Pagination/index.tsx
baomastr: it's counted inside of component, but just moved into the hook for separating logic from layouts

therealemjy: Sorry my bad, I didn't realize `usePagination` is imported from the component now. Great then!

therealemjy (commented):
Looks good to me. I'd just like Corey's opinion on the string displayed next to the component, but other than that I'm happy to approve 👍

src/translation/translations/en.json
coreyar: They both seem fine to me. I think the way you have it works.  `out of` sounds better

therealemjy (approved):

PR #552, opened 2022-05-26
Update transaction error translations

coreyar:
[Load translations in render](https://github.com/VenusProtocol/venus-protocol-interface/pull/552/commits/1a528e7cffe6a913cf0577aaf19d784ef4820fc2)
[Add error translations to SupplyWithdrawForm](https://github.com/VenusProtocol/venus-protocol-interface/pull/552/commits/a09e6d1c09ad858b2b39dd4081aed5747cdad7ae)


src/pages/Dashboard/MintRepayVai/MintVai/index.tsx
```diff
@@ -194,12 +194,16 @@ const MintVai: React.FC = () => {
...
         throw new UiError(
-          VAIControllerTransactionErrorsError[
-            err.error as keyof typeof VAIControllerTransactionErrorsError
+          vaiControllerTransactionErrorsError[
+            err.error as keyof typeof vaiControllerTransactionErrorsError
           ],
-          VAIControllerTransactionErrorsFailureInfo[
-            err.info as keyof typeof VAIControllerTransactionErrorsFailureInfo
+          vaiControllerTransactionErrorsFailureInfo[
+            err.info as keyof typeof vaiControllerTransactionErrorsFailureInfo
```
therealemjy: I can see we now have a lot of different ways of defining and shaping errors.
In order to streamline everything, we could create a function that takes any error and returns an instance of one of our error classes, containing the corresponding translations. The returned error can then be directly given to `toast.error` to be rendered on the UI, or handled differently based on what the UI wants to do with it.

coreyar: In the current pattern there are 2 situations I can think of where we reshape errors - we throw an error and then catch it to re-throw a UI error, or catch an unanticipated runtime error and re-throw a UI error.

I am not really a fan of having a function that is responsible for handling all errors and generically reshaping it because I think our error handling should be more granular and be able to differentiate between errors we anticipated and errors that we never anticipated. 

I've been re-throwing UI Errors to always have views catch this error but maybe that isn't the best approach. In the case of `TransactionError` it isn't even necessary because it also has the same properties as `UIError`. 

I do agree Dealing with translations for the `TransactionError` is a bit awkward.

therealemjy: I think we can simplify things a lot.

The error messages are currently divided into four functions (`loadComptrollerTransactionErrorsErrorTranslations`, `loadComptrollerTransactionErrorsFailureInfoTranslations` `loadTokenTransactionErrorsErrorTranslations` and `loadTokenTransactionErrorsFailureInfoTranslations`), but I think we should consolidate them into twos variable that contains all the translations.
Each error code is unique, whether it comes from the `error` or `info` field from a transaction receipt, and it seems we only display one error message to the users, I think we should be smart with what message we display to them. It seems the error codes from the `info` field give more details than the ones from the `error` field. Perhaps we could display the error messages that corresponds to the code in the `info` field then. I don't know if `info` can be `undefined`, but if yes then we could fallback to `error` when that's the case.

As stated in my previous comment, I'd only store the translation keys so that we don't run into initialization issues with i18next (and can use an object to store them rather than having to call a function that will retrieve all the error messages every time one is used):
```typescript
const transactionErrorMessages = {
  INTERNAL_ERROR: 'transactionError.internalError',
  NO_ERROR: 'transactionError.noError',
  UNAUTHORIZED: 'transactionError.unauthorized',
  ...
};

const transactionErrorInfoMessages = {
  ACCEPT_ADMIN_PENDING_ADMIN_CHECK: 'transactionErrorInfo.acceptAdminPendingAdminCheck',
  MINT_FEE_CALCULATION_FAILED: 'transactionErrorInfo.mintFeeCalculationFailed',
  ...
};
```

We can also consolidate all our error classes into one. We currently have three classes, but they old more or less the same data only using different prop names. I'd simply use an `isInternal` prop to differentiate internal (aka unexpected) errors from expected errors.
The error class would look like that (I prefixed it with "V" for "Venus", we could name it anything else that makes sense):

```typescript
export type TransactionErrorCode = keyof typeof transactionErrorMessages;
export type TransactionErrorInfoCode = keyof typeof transactionErrorInfoMessages;

interface IVErrorParams<T = undefined> {
  message: string;
  additionalInfo?: T; // This would be used to pass any additional information that might be relevant (token ID, transaction receipt etc.)
  errorCode?: TransactionErrorInfoCode | TransactionErrorCode; // We can extend this field to handle other errors which aren't related to transactions
  isInternal?: boolean;
}

export class VError<T = undefined> extends Error {
  message: IVErrorParams<T>['message'];
  additionalInfo?: IVErrorParams<T>['additionalInfo'];
  errorCode?: IVErrorParams<T>['errorCode'];
  isInternal?: IVErrorParams<T>['isInternal'];

  constructor(error: IVErrorParams<T>) {
    this.message = message;
    this.additionalInfo = additionalInfo;
    this.errorCode = errorCode;
    this.isInternal = isInternal;
  }
}
```

Doing that opens up to further simplification, because we can then have only one function to check for errors from transactions rather than the four we currently have (`checkForTransactionError`, `checkForComptrollerTransactionError`, `checkForTokenTransactionError` and `checkForVaiControllerTransactionError`):

```typescript
const handleTransactionResult = (transactionReceipt: TransactionReceipt) => {
  if (receipt.events?.Failure) {
    const { error, info } = receipt.events?.Failure.returnValues;
    
    let isInternal = false;
    let message; // Default to internal error message
    if (info && Object.prototype.hasOwnProperty(transactionErrorInfoMessages, info)) {
      message = transactionErrorInfoMessages[info];
    } else if (Object.prototype.hasOwnProperty(transactionErrorMessages, error)) {
      message = transactionErrorMessages[error];
    } else {
      message = transactionErrorMessages.INTERNAL_ERROR;
      isInternal = true;
    }

    throw new VError({
      message,
      errorCode: error,
      isInternal,
    });
  }

  return receipt;
};
```

This means we can then wrap all our mutation functions with `handleTransactionResult`, while UIs won't have to manipulate errors when they want to display them since they can then just pass them to `toast.error`. When we need to handle edge cases, we can use the error object to gather additional information and act accordingly.

coreyar: > As stated in my previous comment, I'd only store the translation keys so that we don't run into initialization issues with i18next (and can use an object to store them rather than having to call a function that will retrieve all the error messages every time one is used):

I choose the path I did to avoid having to manage comments. There are other ways to load these translations besides calling a function. From the docs it looks like we could namespace them and load the namespace when initing.

coreyar: > Doing that opens up to further simplification, because we can then have only one function to check for errors from transactions rather than the four we currently have (`checkForTransactionError`, `checkForComptrollerTransactionError`, `checkForTokenTransactionError` and `checkForVaiControllerTransactionError`):

This will still be necessary because these contracts throw errors with the same key but we may need to translate / display them differently

coreyar: > Each error code is unique, whether it comes from the `error` or `info` field from a transaction receipt, and it seems we only display one error message to the users, I think we should be smart with what message we display to them. It seems the error codes from the `info` field give more details than the ones from the `error` field. Perhaps we could display the error messages that corresponds to the code in the `info` field then. I don't know if `info` can be `undefined`, but if yes then we could fallback to `error` when that's the case.

This gets more difficult with the strategy outlined of always throwing a single error because we'd need to add further flags to know when we differ to additional information and when we show the message. Could also be set when throwing the error but that is a bit confusing with transactions because we'd be swapping the arguments but may be acceptable
Edit: the later is how your example is written

coreyar: In your example we drop information from the returned error event and I don't think this is ideal for troubleshooting.
I think we should handle all the information returned by the error as a separate concern than how we display that error to the user. The current pattern supports this. We throw the error with its complete details and then we decide as a separate concern how to show a toast to the user ( or take another action)

therealemjy: > I choose the path I did to avoid having to manage comments. There are other ways to load these translations besides calling a function. From the docs it looks like we could namespace them and load the namespace when initing.

If a name space works, sure. I just want to store all of them in the same object for both performance and usability  reasons (an object can be easily manipulated, like I did in my example to extract the error codes for example).

> This will still be necessary because these contracts throw errors with the same key but we may need to translate / display them differently

We currently don't, so we should go with the simplest solution for the current problem.

> This gets more difficult with the strategy outlined of always throwing a single error because we'd need to add further flags to know when we differ to additional information and when we show the message. Could also be set when throwing the error but that is a bit confusing with transactions because we'd be swapping the arguments but may be acceptable
Edit: the later is how your example is written

You wouldn't have this issue because our current strategy is that we always display errors (when they are actively handled), which makes sense since otherwise users don't get any feedback of what happened with their transaction.
As you highlighted in the second part of your message, you wouldn't lose the context since any additional information is passed to the `additionalInfo` prop. You'll also see in my example that this prop is set using a generic, which means we'll be able to type it accordingly.

Having one error for the entire codebase will simplify things a lot and make us more consistent. We currently hide a lot of errors when they happen, which isn't a great UX. Streamlining our error handling means errors will always be displayed to users without us having to add boilerplate code.

> In your example we drop information from the returned error event and I don't think this is ideal for troubleshooting.
I think we should handle all the information returned by the error as a separate concern than how we display that error to the user. The current pattern supports this. We throw the error with its complete details and then we decide as a separate concern how to show a toast to the user ( or take another action)

This joins what I just mentioned: having one error class for the whole codebase makes us more consistent.
If typed well, you also won't lose any context. We just have to be smart about what we pass to the error itself.
For example, say a withdraw transaction fails because there isn't enough funds on the contract, the error could look like that:
```typescript
const error = new VError<{ tokenId: TokenId }>({
  message: "Insufficient liquidity",
  errorCode: INSUFFICIENT_LIQUIDITY,
  additionalInfo: {
    tokenId: 'doge'
  }
});
```

coreyar: I understand the motivation but I still don't think this is the best idea. There will be scenarios where we have a async function that could throw errors for different reasons, each error could have a different set of additional information and messages. It will be come complicated figuring out how to format that information correctly where as checking the instance of the error will tell us exactly what type of error we ended up with.

therealemjy: You'd want to use `errorCode` to distinguish the errors from one another in this case. I don't think the current error classes give us a lot of extra context:
- `UiError` sets a title and a description, but the toasts only render one message string so using any of the other two instance classes would work the same
- `InternalError` only contains a message string. If we wanted to handle a specific edge case, we wouldn't have any information other than the string itself available to us
- `TransactionError` is the most complete class and I think it's a good base, which is what I used to give the example of the `VError` class

src/translation/transactionErrors.ts
```diff
@@ -1,484 +1,252 @@
...
+  MARKET_ALREADY_LISTED: t('transactionErrors.marketAlreadyListed'),
+  MATH_ERROR: t('transactionErrors.mathError'),
+  NONZERO_BORROW_BALANCE: t('transactionErrors.nonzeroBorrowBalance'),
+  PRICE_ERROR: t('transactionErrors.priceError'),
+  REJECTION: t('transactionErrors.rejection'),
+  SNAPSHOT_ERROR: t('transactionErrors.snapshotError'),
+  TOO_MANY_ASSETS: t('transactionErrors.tooManyAssets'),
+  TOO_MUCH_REPAY: t('transactionErrors.tooMuchRepay'),
+  INSUFFICIENT_BALANCE_FOR_VAI: t('transactionErrors.insufficientBalanceForVai'),
+  ACCEPT_ADMIN_PENDING_ADMIN_CHECK: t('transactionErrors.acceptAdminPendingAdminCheck'),
```
therealemjy: Although this looks clean, it unfortunately won't work.
i18next can't be used before it's been initialized, which means that if this is imported before the library was initialized then `t` will return the translation keys rather than the translation strings.
I don't think we can get around storing the translation keys and adding the comments to make sure they're picked up by the parser. To make it easier to track the keys and make sure none is forgotten, we can add each comment next to its corresponding key (rather than listing all the keys then commenting all of them in one block):
```typescript
const transactionErrorTranslations = {
  // Translation key, do not remove - t('transactionErrors.noError')
  NO_ERROR: 'transactionErrors.noError',
  // Translation key, do not remove - t('transactionErrors.comptrollerMismatch')
  UNAUTHORIZED: 'transactionErrors.comptrollerMismatch',
  ...
};
```

src/pages/Dashboard/MintRepayVai/RepayVai/index.tsx
```diff
@@ -91,7 +88,14 @@ export const RepayVaiUi: React.FC<IRepayVaiUiProps> = ({
...
     } catch (error) {
-      toast.error({ message: (error as Error).message });
+      let { message } = error as Error;
+      if (error instanceof VError && error.errorCode === 'transactions') {
+        message =
+          transactionErrorTranslations[error.message as keyof typeof transactionErrorTranslations];
+      }
+      toast.error({
+        message,
+      });
```
therealemjy: We can reduce the boilerplate code and create a util that takes an error param of the type `VError` and displays an error toast using that information

therealemjy (commented):
I like your new approach. The `data` field on `VError` based on another param is nice. I have a just a few suggestions.

The `message` param from the default `Error` class `VError` extends seems to have two possible usages: it either corresponds to the error message itself, or an enum from one of the error enums. I think that can get confusing (it took me a minute to understand how it was used in the code, comparing how an error is instantiated in `src/pages/ConvertVrt/Convert/index.tsx` to how it is instantiated in `checkForTransactionFailureEvent`).

The `code` passed to `VError` in this instance is rather a type of error ("transaction", "unexpected" or "interaction") in my opinion, which is valuable information.

`transactionErrorTranslations` consolidates all the error messages into one object, which I think is the right approach. Given that, I think `ComptrollerErrorReporterError`, `ComptrollerErrorReporterFailureInfo`, `TokenErrorReporterError`, `TokenErrorReporterFailureInfo`, `VAIControllerErrorReporterError`, `VAIControllerErrorReporterFailureInfo` should also be consolidated into one `TransactionErrorCode` type which can actually be defined using `transactionErrorTranslations`:
```typescript
type TransactionErrorCode = keyof typeof transactionErrorTranslations;
```
We could later on have other types to map other types of errors (e.g.: `InteractionErrorCode`).

With these things in mind, the `VError` class would be a bit different and look something like that:
```typescript
type ErrorType = keyof IVErrorParamsMap;
type ErrorCode = TransactionErrorCode; // This type can be extended using a union to map other error types

export class VError<E extends ErrorType> extends Error {
  type: E;
  code: ErrorCode;
  data: IVErrorParamsMap[E] | undefined;

  constructor({
    type,
    code,
    data,
  }: {
    type: E;
    code: ErrorCode;
    data: IVErrorParamsMap[E];
  }) {
    super(error);
    this.type = type;
    this.code = code;
    this.data = data;
  }
}
```

- the error and how it is displayed are separated like you wanted (`VError` doesn't contain any error message or anything specific to the UI)
- a util can then be created to display a `VError` instance using a toast
- a small thing: I'd rename `errorCode` to `code`, since you're already in the context of the error (otherwise `data` and `type` would also need to be renamed `errorData` and `errorType` which is unnecessary in my opinion)

One last thing: I think we should put all the files responsible for the general error handling logic inside the same folder, just like we did with `translation`. Perhaps an `errorHandling` folder at the of `src` directory. What do you think?

therealemjy (approved):

PR #556, opened 2022-05-26
Refactor table to set alignment per column

coreyar:
The History table differs in column alignment pattern from the rest of the tables. Our custom css was causing conflicts trying  to introduce a new alignment pattern.

The TableCell component takes a prop to set alignment which his PR uses to set custom alignments per column

### DIFF
--- a/src/components/v2/Table/Head.tsx
+++ b/src/components/v2/Table/Head.tsx
@@ -9,7 +9,14 @@ import { visuallyHidden } from '@mui/utils';
 import { Icon } from '../Icon';
 import { useStyles } from './styles';
 
-interface IHeadProps<C extends { key: string; label: string; orderable: boolean }[]> {
+interface IColProps {
+  key: string;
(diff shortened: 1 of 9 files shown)


therealemjy:
Could we not target the specific columns that need different alignment rules in CSS?
If we set these through JavaScript we move away from trying to have styles separated from the business logic.

coreyar:
> Could we not target the specific columns that need different alignment rules in CSS? If we set these through JavaScript we move away from trying to have styles separated from the business logic.

This is getting set in css but leveraging the styles in the table. Internally the component is processing the styles in JS and writing css. Because we are using the material MUI I believe this is a clean way to handle per column alignment. Having defaults and trying to override them become increasingly complex and verbose.

therealemjy:
Ok I see, I forgot the MUI part of this. I think that's the right approach then 👍

therealemjy (approved):

PR #557, opened 2022-05-26
Add history page UI

coreyar:
This PR adds the skeleton for the history page and the table
- Includes mock transactions
- Configured i18next to format dates including an argument to format the relative distance
- Adds an argument to format coins without the symbol

This PR was rebased with [PR 556](https://github.com/VenusProtocol/venus-protocol-interface/pull/556) which refactors the tables to set the alignment per column


src/pages/History/HistoryTable/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+    ],
+    [],
+  );
+
+  const cardColumns = useMemo(() => {
+    const newColumns = [...columns];
+    const [liquidityCol] = newColumns.splice(5, 1);
+    newColumns.splice(3, 0, liquidityCol);
+    return newColumns;
+  }, [columns]);
```
therealemjy: I think this makes it hard to understand what the final card columns look like.
For simplicity I'd just declare the columns directly (like you did just above with the `columns` variable).

src/types/index.ts
```diff
@@ -187,3 +187,38 @@ export interface MarketSnapshot {
   totalSupply: string;
   updatedAt: string;
 }
+
+export type TransactionEvents =
```
therealemjy: The type name should not be pluralized
```suggestion
export type TransactionEvent =
```

src/types/index.ts
```diff
@@ -187,3 +187,38 @@ export interface MarketSnapshot {
...
+  category: 'vtoken' | 'vai' | 'vote';
+  createdAt: Date;
+  event: TransactionEvents;
+  from: string;
+  id: number;
+  timestamp: string | null;
+  to: string;
+  transactionHash: string;
+  updatedAt: string;
+  vTokenAddress: string;
```
therealemjy, line 223: Can we get the vToken ID too, or does the backend/smart contracts only return the address?

src/pages/History/HistoryTable/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+      value: txn.id,
+      align: 'left',
+    },
+    {
+      key: 'type',
+      render: () => (
+        <Token
+          variant="small2"
+          css={styles.whiteText}
+          symbol={getTokenIdFromVAddress(txn.vTokenAddress) as TokenId}
```
therealemjy: The Token ID might not always correspond to its symbol, for example the casing could be different

src/pages/History/HistoryTable/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+      align: 'right',
+    },
+    {
+      key: 'createdAt',
+      render: () => (
+        <Typography variant="small1" css={styles.whiteText}>
+          {t('history.createdAt', { date: txn.createdAt })}
+        </Typography>
+      ),
+      value: txn.createdAt.toDateString(),
```
therealemjy: Since this column needs to be orderable:
```suggestion
      value: txn.createdAt.getTime(),
```

src/pages/History/HistoryTable/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+    {
+      key: 'type',
+      render: () => (
+        <Token
+          variant="small2"
+          css={styles.whiteText}
+          symbol={getTokenIdFromVAddress(txn.vTokenAddress) as TokenId}
+        />
+      ),
+      value: txn.vTokenAddress,
```
therealemjy, line 73: I think the designs for this are incorrect. The name of the column is "Type" yet we only display the token logo and name for it.
In the current app, we display the token logo and the transaction type:
Screen Shot 2022-05-27 at 10 08 24

I'll ask the design team about that

src/translation/index.tsx
```diff
@@ -20,6 +22,19 @@ export const init = () =>
...
       escapeValue: false,
+      format: (value, format, lng) => {
+        const locales = { en: enUS };
+
+        if (isDate(value)) {
+          const locale = lng && lng in locales ? locales[lng as keyof typeof locales] : locales.en;
+          if (format === 'distanceToNow') {
+            return formatDistanceToNowStrict(value, { locale });
+          }
+          console.log({ value, format, lng, locale });
```
therealemjy: This needs to be removed

src/utilities/getTokenIdFromVAddress.ts
```diff
@@ -0,0 +1,14 @@
+import vBepTokens from 'constants/contracts/addresses/vBepTokens.json';
+import { TokenId } from 'types';
+
+const getTokenFromVAddress = (address: string) => {
+  const token = Object.entries<{ [key: string]: string }>(vBepTokens).find(
+    ([, value]) =>
+      value[process.env.REACT_APP_CHAIN_ID || '56'].toLowerCase() === address.toLowerCase(),
```
therealemjy, line 7: We should use the `isOnTestnet` config variable instead.

src/utilities/getTokenIdFromVAddress.ts
```diff
@@ -0,0 +1,14 @@
+import vBepTokens from 'constants/contracts/addresses/vBepTokens.json';
+import { TokenId } from 'types';
+
+const getTokenFromVAddress = (address: string) => {
```
therealemjy: You are returning the vToken ID so I'd rename the function accordingly:
```suggestion
const getVTokenIdFromVAddress = (address: string) => {
```

therealemjy:
There's some differences with the designs:
- the texts inside cells of the "Type" column should be white
- the texts inside cells of the "Block" column should have a line-height of 1.5 and a font size of 14px
- none of the texts should be bold

Regarding the last point, we had that issue with the market page already. I really think we need to change the Table component directly so the cells do not render with bold text by default.

src/pages/History/HistoryTable/index.tsx
coreyar: I think the logic is pretty easy to understand but maybe it isn't clear why. Adding a second definition would introduce duplicated code. How about a comment explaining why this is happening?

The purpose is to rearrange the columns for the mobile cards

src/types/index.ts
coreyar: I think `TransactionEvents` is more accurate. because it is a union of all possible events
`TransactionEvent` would be one element from the union

src/types/index.ts
coreyar: This is returned from the backend api. We could look at updating the payload

src/pages/History/HistoryTable/index.tsx
coreyar: This  prop is mislabeled and probably didn't get updated when we established using symbol to display and id to identify. This component is manipulating the casing on this prop to use it as both

src/utilities/getTokenIdFromVAddress.ts
coreyar: This is not returning the vTokenID. It is returning the tokenId. e.g. `bnb`. 
[Source](https://github.com/VenusProtocol/venus-protocol-interface/blob/develop/src/constants/contracts/addresses/vBepTokens.json)

src/utilities/getTokenIdFromVAddress.ts
coreyar: I don't think that makes sense here. The unary is here because  `process.env.REACT_APP_CHAIN_ID` could be undefined

src/types/index.ts
therealemjy: Even if it's a union, at the end it only corresponds to one possible values out of all the options it accepts

src/types/index.ts
therealemjy: Got you 👍
I think for now let's leave it as is

src/pages/History/HistoryTable/index.tsx
therealemjy: Ok I see. I think we should update `Token`, so that it accepts a token ID and retrieves the corresponding token using `getToken` to get the symbol.
There might be cases where the symbol doesn't correspond to the ID uppercased (that's the case with the VBEP20 tokens for example: vXVS, vBNB etc.)

src/utilities/getTokenIdFromVAddress.ts
therealemjy: Got you. Then it should be named `getTokenIdFromVAddress`

src/utilities/getTokenIdFromVAddress.ts
therealemjy: Ah right my bad, I read this wrong! You're right

src/pages/History/HistoryTable/index.tsx
coreyar: Ticketed - https://app.clickup.com/t/2vdgmtp

therealemjy: Thanks 👍
Just make sure to add the story points, tags and acceptance criteria. Sergio will thank you for that

src/pages/History/HistoryTable/index.tsx
therealemjy: I saw you made the change. I think that's a good call anyway (otherwise naming the column "Type" doesn't make sense). Just one thing: the token logo and the text need to be vertically aligned:
Screen Shot 2022-05-27 at 17 11 35

src/pages/History/HistoryTable/index.tsx
therealemjy: I don't find it easy to read, but it's not a big deal to be honest. The comments will help for sure 👍

Just one thing I noticed now, we have the same thing we have with the cards on the market table: the rows aren't spaced enough.

App:
Screen Shot 2022-05-27 at 17 14 08

Designs:
Screen Shot 2022-05-27 at 17 16 40

Dmitriy fixed this by adding `row-gap: ${theme.spacing(5)};` in the right place.

therealemjy (approved):

PR #558, opened 2022-05-26
Add checkbox

coreyar:
Adds a checkbox component needed for the History screen

### DIFF
--- a/src/components/v2/Checkbox/index.tsx
+++ b/src/components/v2/Checkbox/index.tsx
@@ -1,0 +1,28 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+import MuiCheckbox from '@mui/material/Checkbox';
+import { Icon } from 'components';
+import { useStyles } from './styles';
+
+export interface ICheckboxProps {
+  value: boolean;
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+  className?: string;
+}
+
+export const Checkbox = ({ value, onChange, className }: ICheckboxProps) => {
+  const styles = useStyles();
+  return (
+    <MuiCheckbox
+      className={className}
+      css={styles.root}
+      checked={value}
+      onChange={onChange}
+      icon={<Icon name="checkboxBorder" />}
+      checkedIcon={<Icon name="checked" />}
+      disableRipple
+    />
+  );
+};
+
+export default Checkbox;
--- a/src/components/v2/Icon/icons/checkboxBorder.tsx
+++ b/src/components/v2/Icon/icons/checkboxBorder.tsx
@@ -1,0 +1,10 @@
+import * as React from 'react';
+import { SVGProps } from 'react';
+
+const SvgCheckboxBorder = (props: SVGProps<SVGSVGElement>) => (
+  <svg viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg" {...props}>
+    <rect x="0.5" y="0.5" width="23" height="23" rx="5.5" stroke="currentColor" />
+  </svg>
+);
+
+export default SvgCheckboxBorder;
--- a/src/components/v2/Checkbox/index.stories.tsx
+++ b/src/components/v2/Checkbox/index.stories.tsx
@@ -1,0 +1,44 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+import { ComponentMeta, Story } from '@storybook/react';
+import { css } from '@emotion/react';
+import { withCenterStory, withOnChange } from 'stories/decorators';
+import { noop } from 'lodash';
+import { Checkbox, ICheckboxProps } from '.';
+
+export default {
+  title: 'Components/Checkbox',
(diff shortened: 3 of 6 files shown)


src/components/v2/Checkbox/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { useStyles } from './styles';
+
+export interface ICheckboxProps {
+  value: boolean;
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+}
+
+export const Checkbox = ({ value, onChange }: ICheckboxProps) => {
+  const styles = useStyles();
+  console.log(styles);
```
therealemjy: This needs to be removed

src/components/v2/Checkbox/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <MuiCheckbox
+      css={styles.root}
+      checked={value}
+      onChange={onChange}
+      icon={<Icon name="checkboxBorder" />}
+      checkedIcon={<Icon name="checked" />}
+    />
```
therealemjy: I don't think we should use MUI for that, because with their implementation we don't have control over the state. `checked` is only used to initialize the state of the component, after that the component handles its own state which means we'd have to react to that and effectively duplicates states. You can see that because in the story you pass `noop` to the `onChange` handler, yet clicking on the checkbox changes its state.
Given how simple the design of the component is, I'd just build it with plain HTML and CSS. It will also make it simpler to correspond to our design (no ripple effect, correct colors etc...).

src/components/v2/Icon/icons/checkboxBorder.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+import { SVGProps } from 'react';
+
+const SvgCheckboxBorder = (props: SVGProps<SVGSVGElement>) => (
+  <svg viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg" {...props}>
+    <rect x="0.5" y="0.5" width="23" height="23" rx="5.5" stroke="currentColor" />
+  </svg>
+);
```
therealemjy, line 8: To continue on why I think we shouldn't use MUI; we wouldn't need to use an svg to set the border.
If we scale the checkbox, the SVG will scale with it and so the border width won't be constant.

src/components/v2/Checkbox/index.tsx
coreyar: That is not correct. `noop` is passed for completeness when creating the args for the story. The prop used in the story comes from the withState decorator

src/components/v2/Icon/icons/checkboxBorder.tsx
coreyar: I don't see an issue here. Commonly you style a checkbox using :after and an adjacent element similar to this where it is using an `SVG` (also no uncommon). In either case we would be sizing the input and the styles

src/components/v2/Checkbox/index.tsx
therealemjy: Ah right! My bad.

If you want to keep using MUI for that, then these need to be addressed:
- the border color when the checkbox isn't hovered is incorrect (it should be grey, not blue)
- the ripple effect should be disabled

src/components/v2/Icon/icons/checkboxBorder.tsx
therealemjy: Using an SVG is fine for the check icon, but as I mentioned using an SVG for the border means it won't scale properly if the checkbox width and height change.
This isn't an issue for now since we only use one size of checkbox, but it's common to have different checkbox size. If that time comes, we can make the update then.

src/components/v2/Checkbox/index.tsx
coreyar: The hover state is blue (rgba(58, 120, 255, 1)) according to the designs which is currently set

coreyar: When the checkbox isn't hovered the border color is grey `rgb(149, 151, 161)`

src/components/v2/Icon/icons/checkboxBorder.tsx
coreyar: This component does scale. We can either use their provided size prop or we can use css. I pushed an example

src/components/v2/Checkbox/index.tsx
therealemjy: Great 👍

src/components/v2/Icon/icons/checkboxBorder.tsx
therealemjy: Yeah, so if you look at the border width of the scaled checkbox, it is thinner than the one of the "normal" sized checkbox: that was my point. If we ever have checkboxes which are smaller in the app, it is very likely designers will make them have the same border width.

As I said though, that's not the case now so I'm happy to approve the PR

therealemjy (approved):

PR #559, opened 2022-05-26
Refactor asset fetching strategy

therealemjy:
We're currently experiencing performance issues with the dashboard, which this PR fixes for the most part. In a nutshell, the biggest issue comes from the `useUserMarketInfo` (renamed `useGetUserMarketInfo` for consistency) which recreates the list of assets every time it re-renders.
The solution this PR provides is that it memoizes the list so it is only recalculated when values change (updated data is fetched, cache is updated etc.).

We also fetch all of the `hypotheticalLiquidity` field values of the tokens using separate queries (we have to since they come from a contract). When each query is successful, the list of asset is recalculated.
The solution this PR provides is that the hypothetical liquidity of a token is asynchronously fetched when attempting to remove it from collaterals (which is the only existing case for this value), which removes as many queries as there are tokens.


src/pages/Market/Header/index.tsx
```diff
@@ -65,18 +65,23 @@ export const HeaderUi: React.FC<IHeaderProps> = ({
...
+  });
+
   return (
     <HeaderUi
-      totalSupplyCents={treasuryTotalSupplyUsdBalanceCents}
-      totalBorrowCents={treasuryTotalBorrowUsdBalanceCents}
-      availableLiquidityCents={treasuryTotalAvailableLiquidityUsdBalanceCents}
-      totalTreasuryCents={treasuryTotalUsdBalanceCents}
+      totalSupplyCents={
+        getUserMarketInfoData?.treasuryTotalSupplyUsdBalanceCents || new BigNumber(0)
```
coreyar: Default values can also be handled with placeholder data via react query

src/pages/Dashboard/index.tsx
```diff
@@ -85,22 +85,24 @@ const DashboardUi: React.FC<IDashboardUiProps> = ({
...
   return (
     <DashboardUi
       accountAddress={account?.address || ''}
-      assets={assets}
-      userTotalBorrowLimitCents={userTotalBorrowLimitCents}
-      userTotalBorrowBalanceCents={userTotalBorrowBalanceCents}
-      userTotalSupplyBalanceCents={userTotalSupplyBalanceCents}
+      assets={getUserMarketInfoData?.assets || []}
+      userTotalBorrowLimitCents={
+        getUserMarketInfoData?.userTotalBorrowLimitCents || new BigNumber(0)
```
coreyar: Default values can also be handled with placeholder data via react query

src/pages/Dashboard/Modals/BorrowRepay/AccountData/index.tsx
```diff
@@ -34,13 +34,14 @@ const AccountData: React.FC<IAccountDataProps> = ({
   const styles = useStyles();
   const { account } = React.useContext(AuthContext);
 
-  const { assets, userTotalBorrowBalanceCents, userTotalBorrowLimitCents } = useUserMarketInfo({
+  // TODO: handle loading state
+  const { data: getUserMarketInfoData } = useGetUserMarketInfo({
```
coreyar: Many of the changes in this PR are checking if the value returned exists. We could also leverage react query using placeholders

coreyar (changes requested):

src/pages/Dashboard/Modals/BorrowRepay/AccountData/index.tsx
therealemjy: That's a good point yeah. I originally removed the placeholders to see if that causing issues, but now they can be put back for sure. I'll update the PR

src/clients/api/queries/useGetUserMarketInfo.ts
```diff
@@ -0,0 +1,305 @@
...
+  isLoading: boolean;
+  data: IData;
+}
+
+const vTokenAddresses: string[] = Object.values(VBEP_TOKENS).reduce(
+  (acc, item) => (item.address ? [...acc, item.address] : acc),
+  [],
+);
+
+// TODO: decouple, this hook handles too many things
```
baomastr: shall we make a ticket for this `TODO`?

src/pages/Dashboard/Modals/BorrowRepay/AccountData/index.tsx
```diff
@@ -34,13 +34,14 @@ const AccountData: React.FC<IAccountDataProps> = ({
   const styles = useStyles();
   const { account } = React.useContext(AuthContext);
 
-  const { assets, userTotalBorrowBalanceCents, userTotalBorrowLimitCents } = useUserMarketInfo({
+  // TODO: handle loading state
```
baomastr: add a ticket here too pls

src/pages/Dashboard/Modals/BorrowRepay/Borrow/index.tsx
```diff
@@ -205,23 +206,27 @@ const Borrow: React.FC<IBorrowProps> = ({ asset, onClose, isXvsEnabled }) => {
   // Calculate maximum and safe maximum amount of coins user can borrow
   const [limitTokens, safeLimitTokens] = React.useMemo(() => {
     // Return 0 values if borrow limit has been reached
-    if (userTotalBorrowBalanceCents.isGreaterThan(userTotalBorrowLimitCents)) {
+    if (
+      getUserMarketInfoData.userTotalBorrowBalanceCents.isGreaterThan(
```
baomastr: could this constant also been decomposed like `const { userTotalBorrowBalanceCents, ... } = getUserMarketInfoData`?

src/pages/Dashboard/Modals/BorrowRepay/AccountData/index.tsx
therealemjy: Sure 👍

src/pages/Dashboard/Modals/BorrowRepay/Borrow/index.tsx
therealemjy: Definitely 👍

src/clients/api/queries/useGetUserMarketInfo.ts
therealemjy: Yeah that's my bad, it's done

src/clients/api/queries/useGetUserMarketInfo.ts
```diff
@@ -0,0 +1,305 @@
...
+  isLoading: boolean;
+  data: IData;
+}
+
+const vTokenAddresses: string[] = Object.values(VBEP_TOKENS).reduce(
+  (acc, item) => (item.address ? [...acc, item.address] : acc),
+  [],
+);
+
+// TODO: decouple, this hook handles too many things (see https://app.clickup.com/t/2d4rfx6)
```
coreyar, line 41: nit: This name doesn't completely make  sense because there is not `getUserMarketInfo` query. This could be improved when it is refactored per the todo

src/pages/Dashboard/Markets/SupplyMarket/index.tsx
```diff
@@ -101,49 +110,75 @@ const SupplyMarket: React.FC<
...
+    const vTokenContract = getVTokenContract(asset.id as VTokenId, web3);
+
+    let assetHypotheticalLiquidity;
+    try {
+      const vTokenBalanceOf = await getVTokenBalanceOf({
+        tokenContract: vTokenContract,
+        account: accountAddress,
+      });
+
+      assetHypotheticalLiquidity = await getHypotheticalAccountLiquidity({
```
coreyar, line 165: Any reason to not use a `react-query` query?

src/constants/functionKey.ts
```diff
@@ -4,7 +4,6 @@ enum FunctionKey {
   GET_MARKETS = 'GET_MARKETS',
   GET_MARKET_HISTORY = 'GET_MARKET_HISTORY',
   GET_ASSETS_IN_ACCOUNT = 'GET_ASSETS_IN_ACCOUNT',
-  GET_HYPOTHETICAL_LIQUIDITY = 'GET_HYPOTHETICAL_LIQUIDITY',
```
coreyar, line 7: Don't we still need this for the query?

src/pages/Dashboard/Markets/SupplyMarket/index.tsx
coreyar: This is a nice update to reduce the number of times we make this query

src/clients/api/queries/useGetUserMarketInfo.ts
therealemjy: We haven't really discussed this before, but I don't think using `useGetSomething` needs to mean that it calls a function that is called `getSomething`. I've used the prefix `useGet` to indicate it's a hook ("use") and that it's a query ("get")

src/pages/Dashboard/Markets/SupplyMarket/index.tsx
therealemjy: I'd like to, but we unfortunately can't because we need to call the function and get its result within the same execution.
Although it is a constraint and not a choice, I actually think it's the right approach too because the hypothetical account liquidity for the token could have changed from the time the page is rendered to when a user try to decollateralize an asset

src/constants/functionKey.ts
therealemjy: No because we no longer use React Query to call `getHypotheticalAccountLiquidity`

coreyar (approved):
Looks good just need tests to pass

PR #560, opened 2022-05-27
Refactor EnableToken

therealemjy:
- refactor `EnableToken` so it handles its own side-effect and displays a loading state. This also fixes an issue where the wrong address was being past as spender in the Mint VAI section of the dashboard.
- remove `isEnabled` prop from `Asset` type
- create `withEnabledToken` decorator for stories


src/pages/Dashboard/Modals/SupplyWithdraw/index.stories.tsx
```diff
@@ -42,20 +42,6 @@ DisconnectedSupply.args = {
   isWithdrawLoading: false,
 };
 
-export const DisabledSupply = Template.bind({});
```
coreyar, line 45: Why are we removing this story?

src/clients/api/mutations/useApproveToken/getSpenderAddress.ts
```diff
@@ -0,0 +1,16 @@
+import { TokenId } from 'types';
```
coreyar: nit: Scattering utility functions like this doesn't seem ideal, regardless of whether they are only used once I think we should organize them in the `utilities` dir. It keeps everything organized in one place, you know where to look if you need a utility function, it decreases the possibility that it gets rewritten. It also means `useApproveToken` doesn't need to get moved to a directory

src/pages/Dashboard/Modals/SupplyWithdraw/index.stories.tsx
therealemjy: Because I forgot to make this update ahah
This story is broken after the changes I made, I need to mock `getAllowance` so in one case the token is considered enabled and in the other it's considered disabled. I'll make the update

src/clients/api/mutations/useApproveToken/getSpenderAddress.ts
therealemjy: I don't think that's the right approach here. The `utilities` folder should contain reusable utils, that's not the case here because it is specific to `useApproveToken`. It is the same thing as having a component that's only rendered in a specific page: we'd include it in the folder of that page only rather than making it a reusable component.

coreyar (approved):

src/pages/Dashboard/Modals/SupplyWithdraw/index.stories.tsx
therealemjy: It's updated. I've created a `withEnabledToken` story so we can wrap individual stories with it to set a token as enabled

src/clients/api/mutations/useApproveToken/index.ts
```diff
@@ -4,17 +4,18 @@ import { TokenId } from 'types';
 import { queryClient, approveToken, IApproveTokenInput, ApproveTokenOutput } from 'clients/api';
 import FunctionKey from 'constants/functionKey';
 import { useTokenContract } from 'clients/contracts/hooks';
+import setCachedTokenAllowanceToMax from './setCachedTokenAllowanceToMax';
 
 const useApproveToken = (
-  { assetId }: { assetId: TokenId },
+  { tokenId }: { tokenId: TokenId },
```
coreyar, line 10: nit: Noticing this name change, I think we should choose between asset and token for referencing assets. I have been going with asset recently because that is what they are called on the frontend but token seems to be preferred in this file

therealemjy: Since we've called our main type `Asset` but the smart contracts use the term "token", I don't think we can use only one for all cases. And I actually think having both makes sense, because `Asset` is a special type we've created and that contains more than the traditional token data.

In this instance, `useApproveToken` calls `approveToken` which calls the `setAllowance` method of a token smart contract, which in turn expects a token address. Because this is directly related to a smart contract call that uses the term "token", I think calling this param "tokenId" is ideal. What do you think?

Another solution would be to rename the `id` prop of `Asset` to `tokenId`, then replace `assetId` for `tokenId` everywhere in the codebase.

src/clients/api/mutations/useApproveToken/getSpenderAddress.ts
therealemjy: Actually, you're right. I didn't realize that although this util is used only once, it is shared by two hooks: `useGetAllowance` and `useApproveToken`, so it should be a reusable hook. I'll make the change

src/clients/api/mutations/useApproveToken/index.ts
coreyar: Having two names  makes sense if they have two different definitions. I don't think extending the returned token data is enough to warrant a new name or definition. 

We should define a layer where we validate and format data we receive from the backend. This layer would be where we would modify any naming conventions for example switching from camel to snake case for example. 

To me token and asset are interchangeable and refer to the same thing so I would pick one but if we can define what they each are then we can use both.

coreyar (approved):

PR #562, opened 2022-05-27
Add history filters

coreyar:
This PR adds the UI for filters on the history page.
It also includes some import fixes for the checkbox and updates the TransactionEvent to be an enum

### DIFF
--- a/src/pages/History/Filters/index.tsx
+++ b/src/pages/History/Filters/index.tsx
@@ -1,0 +1,63 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+import { Paper, Typography } from '@mui/material';
+import { Checkbox, Select } from 'components';
+import { useTranslation } from 'translation';
+import { TransactionEvent } from 'types';
+import { useStyles } from './styles';
+
+export const ALL_VALUE = 'All';
+
+export interface IFilterProps {
+  eventType: TransactionEvent | typeof ALL_VALUE;
+  setEventType: (eventType: TransactionEvent | typeof ALL_VALUE) => void;
+  showOnlyMyTxns: boolean;
+  setShowOnlyMyTxns: (showOnlyMyTxns: boolean) => void;
+}
+
+export const Filters: React.FC<IFilterProps> = ({
+  eventType,
+  setEventType,
+  showOnlyMyTxns,
+  setShowOnlyMyTxns,
+}) => {
+  const { t } = useTranslation();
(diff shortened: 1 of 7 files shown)


src/types/index.ts
```diff
@@ -186,25 +186,26 @@ export interface MarketSnapshot {
...
+  MintVAI = 'MintVAI',
+  Withdraw = 'Withdraw',
+  RepayVAI = 'RepayVAI',
+  Deposit = 'Deposit',
+  VoteCast = 'VoteCast',
+  ProposalCreated = 'ProposalCreated',
+  ProposalQueued = 'ProposalQueued',
+  ProposalExecuted = 'ProposalExecuted',
+  ProposalCanceled = 'ProposalCanceled',
+}
```
therealemjy: Just curious: is there an advantage in using an enum rather than a list of strings?

For enums, I think we should follow the pattern you followed with the errors and uppercase + snake case each item.

src/pages/History/Filters/index.tsx
```diff
@@ -0,0 +1,65 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+import { Paper, Typography } from '@mui/material';
+import { Checkbox, Select } from 'components';
+import { useTranslation } from 'translation';
+import { TransactionEvent } from 'types';
+import { useStyles } from './styles';
+
+export enum AllValue {
+  All = 'All',
```
therealemjy: Same thing here about uppercasing + snake casing

src/types/index.ts
coreyar: I went with an enum because places where the values were referenced  as plain strings it kept throwing a type error that string wasn't equal to a value in the union although they were the same

src/pages/History/Filters/index.tsx
```diff
@@ -0,0 +1,63 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+import { Paper, Typography } from '@mui/material';
+import { Checkbox, Select } from 'components';
+import { useTranslation } from 'translation';
+import { TransactionEvent } from 'types';
+import { useStyles } from './styles';
+
+export const AllValue = 'All';
```
therealemjy: That's a minor thing, but since it's a constant: 
```suggestion
export const ALL_VALUE = 'All';
```

therealemjy (approved):

therealemjy (approved):

PR #563, opened 2022-05-27
Add transaction query

coreyar:
- Extends restService to format params from an object
- Adds a query to fetch transactions
- Hooks query up to History page skeleton
- Adds a model for Transactions
- Adds pagination component to History Table
- Use placeholder key when to field is missing from transaction

This PR contains [d0aa393](https://github.com/VenusProtocol/venus-protocol-interface/pull/563/commits/d0aa393534509a3d89913eddb9141ec454ca25e7) which came from PR 557


src/clients/api/queries/getTransactions.ts
```diff
@@ -0,0 +1,66 @@
...
+  if ('result' in response && response.result === 'error') {
+    // @todo Add specific api error handling
+    throw new Error(response.message);
+  }
+  if (!payload) {
+    // @todo Add specific api error handling
+    throw new Error('Unexpected error retrieving transactions');
+  }
+  const transactions = payload.result.map(data => new Transaction(data));
+  return { ...payload, result: transactions };
```
therealemjy: Ok nice. Yeah I think that's a good approach.
I'd rename `result` to `transactions`, otherwise one might think that all queries should follow the same pattern and use a `result` prop for the output.

src/models/Transaction.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.vTokenAddress = data.vTokenAddress;
+  }
+
+  get createdAt() {
+    return new Date(this._createdAt);
+  }
+
+  get updatedAt() {
+    return new Date(this._updatedAt);
+  }
```
therealemjy, line 55: Why not have `createdAt` and `updatedAt` set as public props which are initialized with `Date` like you do here?

therealemjy (commented):
Looks good, I had two minor comments regarding the approach with formatting data.

I spotted two minor design things as well, one having been addressed only now by the design team unfortunately:
- the "Created at" column header should say "Creted"
- the content of cells in the transaction type column is slightly vertically misaligned:
Screen Shot 2022-05-31 at 5 21 51 PM

src/models/Transaction.ts
coreyar: As a pattern, I like to include all the original values on the model and return the manipulations of those values as getters.

therealemjy: Ok I see. Should the same be done for `amount` then?

coreyar: Yes, if we want to follow that pattern we should

therealemjy (approved):

PR #564, opened 2022-05-27
Setup voting page skeleton

coreyar:
This PR sets up the voting page to build on top of

### DIFF
--- a/src/theme/MuiThemeProvider/muiTheme.ts
+++ b/src/theme/MuiThemeProvider/muiTheme.ts
@@ -106,6 +106,7 @@ export default createTheme({
   typography: {
     fontFamily: FONTS.primary,
     color: PALETTE.text.primary,
+    lineHeight: 1.5,
     h1: {
       fontSize: '2.5rem',
       fontWeight: 600,
--- a/src/containers/Main/Vote/index.stories.tsx
+++ b/src/containers/Main/Vote/index.stories.tsx
@@ -3,7 +3,7 @@ import { withRouter, withProvider } from 'stories/decorators';
 import Vote from '.';
 
 export default {
-  title: 'Pages/Vote',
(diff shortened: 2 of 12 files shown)


src/theme/MuiThemeProvider/muiTheme.ts
```diff
@@ -121,6 +121,7 @@ export default createTheme({
     h4: {
       fontSize: '1.25rem',
       fontWeight: 600,
+      lineHeight: '1.875rem',
```
therealemjy: That corresponds to the line-height that should already be applied by default to all elements (1.5).

coreyar: According to the designs h4 has a lineHeight of 30

Screen Shot 2022-05-31 at 10 29 37 AM

therealemjy: Yeah, that's 1.5 times its font size (20)

coreyar: I am not sure that is correct. `rem` represents the fontsize of the root element so in this case 1.857 * 16 = 30.
If we were using the `em` unit then it would reference the element font size and `1.5em` would be correct

therealemjy: If you don't add any unit to the line height, it will be used as a multiplier of the font size assigned to the element it concerns.
That's why we should set the line height of all text elements to 1.5, so that for example in this case it would calculate the line height as being 1.5*1.25rem = 1.875rem.

However, I can't see anywhere in the theme that we have indeed set the line height of all text elements to 1.5, so I'm wondering if MUI does it by default or if we need to add that.

therealemjy (approved):

therealemjy (approved):

PR #568, opened 2022-05-30
Market pages - Final assembling

therealemjy:
- remove old Market page
- remove old Market Details page
- remove unused modules


src/pages/App.tsx
```diff
@@ -65,13 +64,7 @@ const App = () => (
...
                                 component={process.env.REACT_APP_RUN_V2 ? Market : MarketV1}
                               />
-                              <Route
-                                exact
-                                path={Path.MARKET_DETAILS}
-                                component={
-                                  process.env.REACT_APP_RUN_V2 ? MarketDetails : MarketDetailsV1
-                                }
-                              />
+                              <Route exact path={Path.MARKET_DETAILS} component={MarketDetails} />
```
coreyar: If this PR is to release this page and setup the version we can also remove the Market page and flag and bump the version

therealemjy: We can, I wasn't sure if you were done with the Market page so I just included the Market Details page

therealemjy: Done, I also found a lot of modules from the old app that could be removed so I did just that

coreyar (approved):

PR #570, opened 2022-05-30
Directly query markets on market table

coreyar:
We don't need to use the userMarkets on the Market table page.
This PR updates that page to use markets directly, it also formats some data after it is received from the query


src/pages/Market/MarketTable/index.tsx
```diff
@@ -164,14 +162,11 @@ export const MarketTableUi: React.FC<IMarketTableProps> = ({ assets, getRowHref
...
 
 const MarketTable = () => {
-  const { account } = useContext(AuthContext);
-  const {
-    data: { assets },
-  } = useGetUserMarketInfo({
-    accountAddress: account?.address || '',
+  const { data: { markets } = { markets: [], dailyVenus: undefined } } = useGetMarkets({
+    placeholderData: { markets: [], dailyVenus: undefined },
+    refetchInterval: 10 * 1000, // Refetch the data 10 seconds
```
therealemjy: A small typo:
```suggestion
    refetchInterval: 10 * 1000, // Refetch the data every 10 seconds
```

src/clients/api/queries/getMarkets.ts
```diff
@@ -29,11 +29,27 @@ const getMarkets = async (): Promise<IGetMarketsOutput> => {
...
+            activeMarket.totalBorrowsUsd,
+          ).times(100);
+          activeMarket.treasuryTotalSupplyUsdCents = new BigNumber(
+            activeMarket.totalSupplyUsd,
+          ).times(100);
+        }
+        return activeMarket;
+      })
       .filter(notUndefined);
   }
```
therealemjy: We should avoid mutating variables like that: if later on the code evolves and `response` is used after this, its content will be mutated which might be unexpected by developers.
We can also handle everything with one loop only:
```typescript
if (response?.data?.data) {
  dailyVenus = new BigNumber(response.data.data.dailyVenus);
  markets = Object.keys(VBEP_TOKENS).reduce<Market[]>((acc, vBepTokenId) => {
    const activeMarket = response.data?.data.markets.find(
      (market: Market) => market.underlyingSymbol.toLowerCase() === vBepTokenId.toLowerCase(),
    );

    if (activeMarket) {
      const formattedActiveMarket: Market = {
        ...activeMarket,
        id: activeMarket.underlyingSymbol.toLowerCase(),
        tokenPrice: new BigNumber(activeMarket.tokenPrice),
        liquidity: new BigNumber(activeMarket.liquidity),
        borrowVenusApy: new BigNumber(activeMarket.borrowVenusApy),
        borrowApy: new BigNumber(activeMarket.borrowApy),
        supplyVenusApy: new BigNumber(activeMarket.supplyVenusApy),
        supplyApy: new BigNumber(activeMarket.supplyApy),
        treasuryTotalBorrowsUsdCents: new BigNumber(activeMarket.totalBorrowsUsd).times(100),
        treasuryTotalSupplyUsdCents: new BigNumber(activeMarket.totalSupplyUsd).times(100),
      };

      return [...acc, formattedActiveMarket];
    }

    return acc;
  }, []);
}
```

therealemjy (approved):

therealemjy (approved):

PR #573, opened 2022-05-30
Snapshot textContent for more stable tests

coreyar:
Snapshotting html is brittle because if the markup changes in any way the tests will break but the component may still be valid. For those visual changes we can leverage viz reg snapshots.

In this case, we want to check values are showing up on the page correct. Testing `textContent` rather than the full markup will be a more stable test for this purpose.

This came to light because the generated css class changed from the snapshot in PR 564 causing failures


therealemjy (approved):
Very good idea 👍

PR #577, opened 2022-05-30
Add voting wallet UI

coreyar:
Adds UI for Voting Wallet section on the vote page


src/pages/Vote/index.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { ComponentMeta } from '@storybook/react';
+import Vote from '.';
+
+export default {
+  title: 'Pages/Vote',
+  component: Vote,
+} as ComponentMeta<typeof Vote>;
+
+export const Default = () => <Vote />;
```
therealemjy, line 12: We'll need to wrap this story with a `Router` tag:
Screen Shot 2022-05-31 at 9 00 14 AM

src/theme/MuiThemeProvider/muiTheme.ts
```diff
@@ -121,6 +121,7 @@ export default createTheme({
     h4: {
       fontSize: '1.25rem',
       fontWeight: 600,
+      lineHeight: '1.875rem',
```
therealemjy, line 127: This should be the default line-height of all text elements (1.5)

therealemjy:
I can't check the page in Storybooks for now because it needs to be wrapped with a `Router` tag. Once fixed I can have another look!

src/theme/MuiThemeProvider/muiTheme.ts
coreyar: According to the designs h4 has a larger lineHeight of 30px

therealemjy: It's not larger: 30 is 20 * 1.5, 1.5 being the line height we apply to all elements

src/pages/Vote/VotingWallet/index.tsx
```diff
@@ -1,20 +1,82 @@
...
+}
 
+export const VoteUi: React.FC<IVoteUiProps> = ({ votingWeight, xvsLockedWei }) => {
+  const { t, Trans } = useTranslation();
+  const styles = useStyles();
+  const readableXvsLocked = convertWeiToCoins({
+    valueWei: xvsLockedWei,
+    tokenId: 'xvs',
+    returnInReadableFormat: true,
+    addSymbol: false,
```
therealemjy, line 24: We'll also need to minimize decimals:
```suggestion
    addSymbol: false,
    minimizeDecimals: true,
```

src/components/v2/Button/index.tsx
```diff
@@ -50,3 +56,12 @@ export const PrimaryButton = (props: IButtonProps) => <Button variant="primary"
...
 export const TextButton = (props: IButtonProps) => <Button variant="text" {...props} />;
+export const LinkButton = ({
+  variant = 'primary',
+  fullWidth = false,
+  small = false,
+  ...props
+}: LinkProps & IBaseButtonProps) => {
+  const styles = useStyles({ fullWidth, variant, small });
+  return <Link {...props} css={styles.getButton({ disabled: false })} />;
+};
```
therealemjy, line 67: We should add this variant to the story

therealemjy (commented):
Spotted as few design issues:
- text inside `LinkButton` has the wrong font-weight (should be the same for other buttons)
- "Voting weight" block has a top padding that isn't present in the designs
- both the "Voting weight" and "Total locked" block have horizontal paddings that aren't present in the designs
- the padding between the "Total locked" label and the divider is 24px but should be 16px
- the padding between the "Total locked" block and the submit button is 16px but should be 24px

therealemjy (approved):

PR #587, opened 2022-06-01
Vote UI requests

coreyar:
This PR adds requests needed for the main vote page


src/__mocks__/models/proposal.ts
```diff
@@ -0,0 +1,7 @@
+import { Proposal } from 'models';
+import { IProposalApiResponse } from 'types/apiResponses/proposal';
+import proposalResponse from '../api/proposal.json';
+
+const proposals = proposalResponse.result.map(p => new Proposal(p as IProposalApiResponse));
```
therealemjy: So following our quick discussion yesterday, I'd propose to move to using types rather than models, this way each type isn't tied to specific requests

src/clients/api/queries/getCurrentVotes.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export type GetCurrentVotesOutput = BigNumber;
+
+const getCurrentVotes = async ({
+  xvsVaultContract,
+  accountAddress,
+}: IGetCurrentVotesInput): Promise<GetCurrentVotesOutput> => {
+  const resp = await xvsVaultContract.methods.getCurrentVotes(accountAddress).call();
+  return new BigNumber(resp).div(1e18);
```
therealemjy: You can use `convertWeiToCoins` to make this operation, this way the XVS decimals aren't hard-coded.

src/clients/api/queries/getPendingXvs.ts
```diff
@@ -0,0 +1,19 @@
...
+import { VaiVault } from 'types/contracts';
+
+export interface IGetPendingXvsInput {
+  vaiVaultContract: VaiVault;
+  accountAddress: string;
+}
+
+export type GetPendingXvsOutput = BigNumber;
+
+const getPendingXvs = async ({
```
therealemjy, line 11: This is more of a question than a suggestion: seeing we've now adopted the pattern of suffixing variables with their unit, shall we do the same with queries?

src/models/ProposalAction.ts
```diff
@@ -0,0 +1,23 @@
...
+  constructor(data: IProposalApiResponse['actions'][number]) {
+    this.data = data.data;
+    this.signature = data.signature;
+    this.target = data.target;
+    this.title = data.title;
+    this.value = data.value;
+  }
+}
+
+export default ProposalAction;
```
therealemjy: Same comment about moving to using types

src/__mocks__/models/proposal.ts
coreyar: Whether it is a a type or a model they will represent two things 1) the returned response 2) the shape of data in the app. 

I think models give us an advantage over returning and object. The model can return memoized values via getters, can easily be extended, and doesn't need an extra type defined in addition to the object. 

We don't have to take all the values from the response, we can take just the ones we need which is what I understood your point to be from the discussion

coreyar: Another advantage is shown here. Depending on what we are testing we may want the raw API Response or we may want the santized and formatted response that we will be using in the application. Following this pattern will give us clean mocks with clear purposes and definitions

coreyar: Alternatively if we prefer to return a plain object, we can write a function that would be in place of the class to format the data and could also be extended but wouldn't have the added benefit of memoized getters. This format function could be used to follow the mocking pattern I mentioned in the previous comment

src/components/Basic/Proposal.tsx
```diff
@@ -8,7 +8,7 @@ import moment from 'moment';
 import dashImg from 'assets/img/dash.png';
 import closeImg from 'assets/img/close.png';
 import { Row, Column } from 'components/Basic/Style';
-import { Proposal as ProposalObject } from 'types';
+import { IDeprecatedProposal as ProposalObject } from 'types';
```
coreyar: I have a file here with the deprecated types so they could be deleted when the component was deleted

therealemjy (approved):

PR #591, opened 2022-06-02
Tie together history page

coreyar:
This PR brings together the History page


src/components/v2/Pagination/usePagination.ts
```diff
@@ -54,7 +54,7 @@ export function usePagination({
     : activePageIndex + halfOfPagesCount;
 
   const handlePageChange = (pageIndex: number) => {
-    onChange(pageIndex);
+    onChange(pageIndex + 1);
```
therealemjy: I don't quite understand the +1 here. I can see pagination works perfectly fine on the History page so I'm sure this is needed, but I'm curious as to why

src/components/v2/Table/index.tsx
```diff
@@ -150,6 +151,7 @@ export const Table = ({
           </TableBody>
         </TableMUI>
       </TableContainer>
+      {isFetching && <Spinner />}
```
therealemjy, line 155: Nice touch, despite not having designs for it 👍

therealemjy (commented):
Really nice work man. The code is clean and the final result looks great 👍

I encountered a bug when I pressed on the "My transaction" checkbox:
Screen Shot 2022-06-03 at 11 05 00 AM

Other than that, everything looks good to me.

src/components/v2/Pagination/usePagination.ts
coreyar: Because the pagination starts at 1 and the logic in this hook is using a zero based index 
https://github.com/VenusProtocol/venus-protocol-interface/pull/591/files/c5a673816c92c94e26bb4f6bc2773dd24d727e7e#diff-365b7c2af29c5d548698a705c9813a94f9eb9c0fa5a761b13c5c1c2f583df72dR21

coreyar:
> Really nice work man. The code is clean and the final result looks great 👍
> 
> I encountered a bug when I pressed on the "My transaction" checkbox: Screen Shot 2022-06-03 at 11 05 00 AM
> 
> Other than that, everything looks good to me.

Interesting, I'm not sure why that happened. I can't reproduce it. What I think happened is using the filter brought up a transaction that is missing a `vTokenAddress` address. Are you able to reproduce and confirm?

src/components/v2/Pagination/usePagination.ts
therealemjy: Ahh right. I think this logic of defining the page number displayed to the user should be kept inside the `History` page component then, so that we always use a zero based indexation in our logic (especially since the `onChange` callback of `usePagination` names its param `newPageIndex`).

therealemjy (approved):

src/pages/History/index.tsx
```diff
@@ -16,32 +22,59 @@ export const HistoryUi: React.FC<IHistoryUiProps> = ({
...
       setShowOnlyMyTxns={setShowOnlyMyTxns}
+      walletConnected={walletConnected}
     />
-    <HistoryTable transactions={transactions} />
+    <HistoryTable transactions={transactions} isFetching={isFetching} />
+    {total && (
+      <Pagination
+        itemsCount={total}
+        onChange={(nextIndex: number) => {
+          setCurrentPage(nextIndex + 1);
```
therealemjy, line 44: As mentioned previously, I think the actual page number should reflect the index of the page in the array.
We're only changing the page number visually for users, so in my opinion this should be done on the UI inside `Pagination` directly (line 93 of this file: `src/components/v2/Pagination/index.tsx`)

coreyar: the backend pagination index starts at 1 which is why this is required. Following up in slack

therealemjy (approved):

therealemjy (approved):

PR #592, opened 2022-06-02
Vault Queries

therealemjy:
Adds the queries needed for the updated Vault page.


src/clients/api/queries/getXvsReward.ts
```diff
@@ -17,14 +16,7 @@ const getXvsReward = async ({
...
     .call();
-
-  const totalXvsEarned = new BigNumber(pendingVenus).dividedBy(1e18).dp(VBEP_TOKEN_DECIMALS, 1);
-
-  // Calculate and return total XVS reward
-  const xvsDecimals = getToken('xvs').decimals;
-  const xvsRewardWei = totalXvsEarned.multipliedBy(new BigNumber(10).pow(xvsDecimals));
-
-  return xvsRewardWei;
+  return new BigNumber(pendingVenus);
```
coreyar, line 19: I think this would be where we'd convert to  units / formats we want to use in the app, in this case wei

coreyar (approved):

PR #595, opened 2022-06-03
[VEN-228] Vault - List vesting vaults + VAI vault

therealemjy:
- optimize and decouple `useGetVaults`
- add queries to load VAI vault
- update queries related to fetching vault to add formatting layer and mocked data from contracts
- list vesting vaults and VAI vault on Vault page

The last vault that need to be fetched is the VRT vault.


src/pages/Vault/VaultItem/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+import Paper from '@mui/material/Paper';
+import Typography from '@mui/material/Typography';
+
+import { useTranslation } from 'translation';
+import useConvertToReadableCoinString from 'hooks/useConvertToReadableCoinString';
+import { convertWeiToCoins, formatToReadablePercentage } from 'utilities/common';
+import { TokenId } from 'types';
+import { getToken } from 'utilities';
+import { Icon } from '../../../components/v2/Icon';
+import { Button } from '../../../components/v2/Button';
```
coreyar: We can import using absolute paths

src/pages/Vault/index.tsx
```diff
@@ -1,8 +1,49 @@
...
+export const VaultUi: React.FC<IVaultUi> = ({ vaults }) => {
+  const styles = useStyles();
+
+  if (vaults.length === 0) {
+    return <Spinner />;
+  }
+
+  return (
+    <div css={styles.container}>
+      {vaults.map(vault => (
```
coreyar, line 28: We also need to add the VAI and  VRT vaults. I didn't see a separate ticket for those.

coreyar:
We should be able to achieve better responsiveness
Screen Shot 2022-06-07 at 4 02 41 PM

src/pages/Vault/index.tsx
coreyar: Looks like the VAI vault is in another PR so I assume the VRT will follow

src/pages/Vault/index.tsx
```diff
@@ -1,8 +1,49 @@
...
-const Vault: React.FC = () => <VaultUi />;
+export interface IVaultUi {
+  vaults: Vault[];
+}
 
-export default Vault;
+export const VaultUi: React.FC<IVaultUi> = ({ vaults }) => {
+  const styles = useStyles();
+
+  if (vaults.length === 0) {
```
coreyar: On mainnet this never return this XVS vault

src/__mocks__/models/vaults.ts
```diff
@@ -0,0 +1,24 @@
+import BigNumber from 'bignumber.js';
+
+import { Vault } from 'types';
+
+export const vaults: Vault[] = [
```
coreyar, line 5: I would recommend keeping a mock of raw api responses and then modifying those here

src/pages/Vault/index.tsx
therealemjy: Sorry I don't understand what you mean

therealemjy:
About the responsiveness issue: yes definitely. I've sent a message to Pavel about this last week but he's currently busy with the Isolated Lending feature. I'll ask him again

src/__mocks__/models/vaults.ts
therealemjy: It would be a nightmare to do. Check the `useGetVaults` hook: we get the data from multiple smart contracts and then manipulate it to create vaults.
It's a similar case to the `useGetUserMarketInfo` hook.

therealemjy:
I spoke with Natasha who proposed to use a wider endpoint to make the switch from displaying 2 columns to 1. This fixes the issue and I've pushed the change.

src/__mocks__/models/vaults.ts
coreyar: I don't think it is difficult to do and I think it is a nice organized setup for mocks that gives us the data we need for testing.

src/pages/Vault/index.tsx
coreyar: I expected the XVS vault to be returned when fetching the pools on mainnet. When I pulled down the PR and ran it on the mainnet no vaults are returned and it loads indefinitely.

therealemjy: Yeah it should be. I just tried and it worked for me, maybe you need to pull?

coreyar (approved):

coreyar (approved):

PR #596, opened 2022-06-03
Add Brave Wallet

coreyar:
This PR adds the brave wallet tot his list of wallets you can connect to the dApp


therealemjy:
Nice and easy then! Does Brave Wallet support testnet too?

therealemjy (approved):

PR #599, opened 2022-06-05
Add treasury total hook for Market page

coreyar:
Breaks out treasury totals into separate hook used on the market page


src/pages/Market/MarketTable/index.tsx
```diff
@@ -25,7 +25,7 @@ export const MarketTableUi: React.FC<IMarketTableProps> = ({ markets, getRowHref
 
   const columns = useMemo(
     () => [
-      { key: 'market', label: t('market.columns.market'), orderable: false, align: 'left' },
+      { key: 'asset', label: t('market.columns.asset'), orderable: false, align: 'left' },
```
therealemjy, line 28: The translation will need to be updated too. Watch out: the column header is "Currency"

src/__mocks__/models/markets.ts
```diff
@@ -85,8 +85,8 @@ export const markets: Market[] = [
     lastCalculatedBlockNumber: 19168526,
     borrowerCount: 232,
     supplierCount: 638,
-    treasuryTotalBorrowsUsdCents: new BigNumber(''),
-    treasuryTotalSupplyUsdCents: new BigNumber(''),
+    treasuryTotalBorrowsUsdCents: new BigNumber('0'),
+    treasuryTotalSupplyUsdCents: new BigNumber('0'),
```
therealemjy: So far, we've always considered fiat values to be in dollars:
```suggestion
    treasuryTotalBorrowsCents: new BigNumber('0'),
    treasuryTotalSupplyCents: new BigNumber('0'),
```

src/clients/api/queries/useGetTreasuryTotals.ts
```diff
@@ -0,0 +1,113 @@
...
+import { TREASURY_ADDRESS } from 'config';
+import { VBEP_TOKENS } from 'constants/tokens';
+import { Market } from 'types';
+import { indexBy } from 'utilities/common';
+
+export interface IData {
+  treasuryTotalSupplyUsdBalanceCents: BigNumber;
+  treasuryTotalBorrowUsdBalanceCents: BigNumber;
+  treasuryTotalUsdBalanceCents: BigNumber;
+  treasuryTotalAvailableLiquidityUsdBalanceCents: BigNumber;
```
therealemjy: Same as last comment:
```suggestion
  treasuryTotalSupplyBalanceCents: BigNumber;
  treasuryTotalBorrowBalanceCents: BigNumber;
  treasuryTotalBalanceCents: BigNumber;
  treasuryTotalAvailableLiquidityBalanceCents: BigNumber;
```

src/clients/api/queries/useGetTreasuryTotals.ts
```diff
@@ -0,0 +1,113 @@
...
+    placeholderData: {
+      markets: [],
+      dailyVenus: new BigNumber(0),
+    },
+  });
+
+  const vTokenAddresses: string[] = Object.values(VBEP_TOKENS).reduce(
+    (acc, item) => (item.address ? [...acc, item.address] : acc),
+    [],
+  );
```
therealemjy: Since this is calculated using a constant, it doesn't have to be recalculate on every execution and so can be moved outside the function

src/clients/api/queries/useGetTreasuryTotals.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  const { markets } = getMarketsData;
+  const {
+    treasuryTotalSupplyUsdBalanceCents,
+    treasuryTotalBorrowUsdBalanceCents,
+    treasuryTotalUsdBalanceCents,
+    treasuryTotalAvailableLiquidityUsdBalanceCents,
+  } = useMemo(() => {
+    const data = (markets as Market[]).reduce(
```
therealemjy: `markets` should already be typed

src/clients/api/queries/useGetTreasuryTotals.ts
```diff
@@ -0,0 +1,113 @@
...
+      (acc, curr) => {
+        const toDecimalAmount = (mantissa: string) =>
+          new BigNumber(mantissa).shiftedBy(-curr.underlyingDecimal);
+
+        let treasuryBalance = new BigNumber(0);
+        if (treasuryBalances && treasuryBalances[curr.address]) {
+          treasuryBalance = toDecimalAmount(treasuryBalances[curr.address].tokenBalance);
+        }
```
therealemjy: Given `toDecimalAmount` is only used once, you don't have to make it a function.
Same comment as previous one about adding unit to variable names:
```suggestion
        let treasuryBalanceTokens = new BigNumber(0);
        if (treasuryBalances && treasuryBalances[curr.address]) {
          const mantissa = treasuryBalances[curr.address].tokenBalance;
          treasuryBalanceTokens = new BigNumber(mantissa).shiftedBy(-curr.underlyingDecimal);
        }
```

src/pages/Market/MarketTable/index.tsx
coreyar: For consistency we are using asset across the site, although the design says currency

therealemjy: Ah right, that's fair

src/clients/api/queries/useGetTreasuryTotals.ts
coreyar: I get errors that it isn't which is why I included the type

therealemjy: Then there's an issue with the type of `getMarketsData`. 
Enforcing types like this can be dangerous because it potentially obfuscates what the real type of the variable is.

coreyar: updated

therealemjy (approved):

PR #609, opened 2022-06-07
Update column header on market page

therealemjy:
Change "Currency" to "Asset" to be consistent with other pages.

### DIFF
--- a/src/translation/translations/en.json
+++ b/src/translation/translations/en.json
@@ -188,7 +188,7 @@
     "columns": {
       "borrowApy": "Borrow APY",
       "liquidity": "Liquidity",
-      "market": "Asset",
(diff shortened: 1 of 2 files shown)


coreyar (approved):

src/translation/translations/en.json
```diff
@@ -188,7 +188,7 @@
     "columns": {
       "borrowApy": "Borrow APY",
       "liquidity": "Liquidity",
-      "market": "Currency",
+      "market": "Asset",
```
coreyar:
```suggestion
      "asset": "Asset",
```
Lets match the key and value when we can

therealemjy: Good idea

PR #612, opened 2022-06-07
Add UST / LUNA deprecation warning

coreyar:
Adds warning to disable UST/ LUNA as collateral to keep using the protocol after deprecation

### DIFF
--- a/src/components/v2/Layout/index.tsx
+++ b/src/components/v2/Layout/index.tsx
@@ -1,23 +1,22 @@
 /** @jsxImportSource @emotion/react */
 import React from 'react';
 import Box from '@mui/material/Box';
+import { useTranslation } from 'translation';
 import Sidebar from './Sidebar';
 import Header from './Header';
 import { PageContainer } from './PageContainer';
+import { Banner } from './Banner';
 import { useStyles } from './styles';
 
-export interface ILayoutProps {
-  children: React.ReactNode;
-}
-
-export const Layout: React.FC<ILayoutProps> = ({ children }) => {
+export const Layout: React.FC = ({ children }) => {
   const styles = useStyles();
-
+  const { t } = useTranslation();
   return (
     <div css={styles.layout}>
       <Sidebar />
 
       <Box display="flex" flexDirection="column" flex="1">
+        <Banner showBanner bannerText={t('layout.banner')} />
         <Header />
         <PageContainer>{children}</PageContainer>
       </Box>
--- a/src/translation/translations/en.json
+++ b/src/translation/translations/en.json
@@ -170,6 +170,7 @@
     }
   },
   "layout": {
+    "banner": "UST and LUNA are being deprecated, please ensure they are disabled as collateral. Failure to disable them will result in lost access to the protocol after deprecation.",
     "menuItems": {
       "convertVrt": "Convert VRT",
       "convertVrtTitle": "Convert VRT to XVS",
--- a/src/components/v2/Layout/Banner/index.tsx
+++ b/src/components/v2/Layout/Banner/index.tsx
@@ -1,0 +1,17 @@
+/** @jsxImportSource @emotion/react */
+import React from 'react';
+import { NoticeWarning } from '../../Notice';
+import { useStyles } from './styles';
+
+export interface IBannerProps {
+  showBanner: boolean;
+  bannerText: string;
+}
+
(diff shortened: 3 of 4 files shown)


src/translation/translations/en.json
```diff
@@ -54,14 +54,14 @@
...
     "repay": {
-      "fullRepaymentWarning": "The transaction value might be slightly higher than the current borrow amount to cover any newly-accumulated interests",
       "connectWalletMessage": "Please connect your wallet to repay",
       "currentlyBorrowing": "Currently borrowing",
       "enableToken": {
         "borrowInfo": "Borrow APY",
         "distributionInfo": "Distribution APY",
         "title": "To repay {{symbol}} to the Venus Protocol, you need to enable it first"
       },
+      "fullRepaymentWarning": "",
```
therealemjy: Not sure what this is doing here

src/components/v2/Layout/index.tsx
```diff
@@ -12,12 +14,14 @@ export interface ILayoutProps {
...
   const styles = useStyles();
-
+  const { t } = useTranslation();
+  const bannerText = t('layout.banner');
   return (
     <div css={styles.layout}>
       <Sidebar />
 
       <Box display="flex" flexDirection="column" flex="1">
+        {bannerText && <NoticeWarning css={styles.banner} description={bannerText} />}
```
therealemjy: `bannerText` will always be truthy since the translation key for `layout.banner` will be automatically generated every time someone generates the translations using the script command.

A solution to that can be to create a separate `Banner` component inside `Layout` that we comment out when we don't want to display the banner.

src/components/v2/Layout/styles.ts
```diff
@@ -13,5 +13,21 @@ export const useStyles = () => {
         flex-direction: column;
       }
     `,
+    banner: css`
+      margin: ${theme.spacing(4, 12)};
```
therealemjy: To be consistent with the rest of the page:
```suggestion
      margin: ${theme.spacing(4, 10, 0)};
```

Screen Shot 2022-06-07 at 17 28 01

With change:
Screen Shot 2022-06-07 at 17 30 20

src/components/v2/Layout/styles.ts
```diff
@@ -13,5 +13,21 @@ export const useStyles = () => {
...
+      ${theme.breakpoints.down('xl')} {
+        margin: ${theme.spacing(4, 10)};
```
therealemjy:
```suggestion
        margin: ${theme.spacing(4, 10, 0)};
```

src/components/v2/Layout/styles.ts
```diff
@@ -13,5 +13,21 @@ export const useStyles = () => {
...
+      }
+      ${theme.breakpoints.down('lg')} {
+        margin: ${theme.spacing(6)};
```
therealemjy:
```suggestion
        margin: ${theme.spacing(6, 6, 0)};
```

src/components/v2/Layout/styles.ts
```diff
@@ -13,5 +13,21 @@ export const useStyles = () => {
...
+      }
+      ${theme.breakpoints.down('md')} {
+        margin: ${theme.spacing(4)};
```
therealemjy:
```suggestion
        margin: ${theme.spacing(4, 4, 0)};
```

src/components/v2/Layout/index.tsx
coreyar: It isn't always truthy if you remove the translation but keep the key. That is what I was thinking. Running the translation script again will keep the empty value

therealemjy: That will create unexpected behaviors. Imagine we remove the translation key so the banner no longer appears, then a developer works on a new feature and runs `yarn extract-translation-keys`. The translation will be automatically added to the JSON file and the website will now have banner that says "TRANSLATION NEEDED".

src/translation/translations/en.json
coreyar: that was from testing it has been resolved

src/components/v2/Layout/index.tsx
coreyar: That being said I think a component is nicer

therealemjy (approved):

therealemjy (approved):

PR #621, opened 2022-06-08
Feature: Add VoteSummary component

baomastr:
https://app.clickup.com/t/2dvkpxw


src/components/v2/VoteSummary/index.tsx
```diff
@@ -0,0 +1,101 @@
+/** @jsxImportSource @emotion/react */
```
coreyar: Since this component will only be used on the Proposal page it can live there. I have pushed a branch I am working on for setting up the skeleton if you want to make the PR into that branch `vote-proposal-skeleton`

src/components/v2/VoteSummary/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+type VoteFrom = {
+  address: string;
+  voteWeightWei: BigNumber;
+  comment?: string;
+};
+
+interface IVoteSummaryProps {
+  className?: string;
+  votesFrom?: VoteFrom[];
```
coreyar: I understand we are doing this because of ActiveVotingProgress but we don't need to use that component here. We can use the `LabeledProgressBar`

ActiveVotingProgress can also use the `LabledProgressBar`

Then we can simplify this to take a vote type, number of votes and total votes

src/translation/translations/en.json
```diff
@@ -565,6 +565,13 @@
       "votedFor": "Voted for"
     }
   },
+  "voteSummary": {
+    "for": "For",
```
coreyar: The single word vote types can probably use a single translation, otherwise we will end up with many copies for the same word in the same context which should be translated the same

src/components/v2/VoteSummary/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+                {comment && (
+                  <Tooltip title={comment}>
+                    <Icon name="bubble" />
+                  </Tooltip>
+                )}
+              </EllipseText>
+              <Typography color="text.primary">
+                {formatCoinsToReadableValue({
+                  value: voteWeightWei,
+                  tokenId: 'xvs',
```
coreyar: nit: We have a constant that we could also use here if we wanted

We also memoize this function

src/components/v2/VoteSummary/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+        {votedForWei && t('voteSummary.for')}
+        {votedAgainstWei && t('voteSummary.against')}
+        {abstainedWei && t('voteSummary.abstain')}
+      </Button>
+
+      <LabeledInlineContent label={t('voteSummary.addresses', { length: votesFrom.length })}>
+        <Typography>{t('voteSummary.votes')}</Typography>
+      </LabeledInlineContent>
+
+      {votesFrom.length > 0 && (
```
coreyar: We can map over an empty list and it won't return anything

therealemjy (commented):
Looks good visually speaking. We'll need to integrate scrolling too:
Screen Shot 2022-06-09 at 13 56 37

https://www.figma.com/file/GYJwWWLbC4B8DrJOPcETz3/Venus-2.0?node-id=6345%3A44934

src/components/v2/VoteSummary/index.tsx
baomastr: I think, reusing of ActiveVotingProgress avoids us from code duplicating, but I agree, that it should be a shared component for that.

src/components/v2/VoteSummary/index.tsx
baomastr: moved component to pages/proposal

src/components/v2/VoteSummary/index.tsx
baomastr: but this makes sense:

> ActiveVotingProgress can also use the LabledProgressBar

src/components/v2/VoteSummary/index.tsx
coreyar: Looks like we created a new directory. This skeleton for this component is called Votes (We can change the name). It is setup in the layout of the page as well (`src/pages/Proposal/index.tsx`).

We should drop this component there so it is rendered on the page with mock data until the page is wired up

src/components/v2/VoteSummary/index.tsx
coreyar: We should cleanup this interface further. We don't need to pass all three options to every instance of this component. 
We need to pass the total for the type of vote and total for all votes and that is it.

If we do that we can swap out `ActiveVotingProgress` for `LabledProgressBar`

src/pages/Proposal/VoteSummary/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <ActiveVotingProgress
+        votedForWei={votedForWei}
+        votedAgainstWei={votedAgainstWei}
+        abstainedWei={abstainedWei}
+        votedTotalWei={votedTotalWei}
+      />
+      <Button css={styles.button} onClick={onClick} disabled={isDisabled}>
+        {votedForWei && t('vote.for')}
+        {votedAgainstWei && t('vote.against')}
+        {abstainedWei && t('vote.abstain')}
```
coreyar: We can pass this in as a prop called `title` so we don't need to render based on all three votes types. After we clean up the interface we won't have these props

src/components/v2/VoteSummary/index.tsx
baomastr: what do you think about using just `type, votedValueWei, votedTotalWei` for ActiveVotingProgress props for handling this cases?

coreyar: I have made this update already

baomastr: just pushed the updates

src/pages/Proposal/VoteSummary/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+  votesFrom = [],
+  className,
+  isDisabled,
+}: IVoteSummaryProps) => {
+  const styles = useStyles();
+  const { t } = useTranslation();
+
+  return (
+    <Paper css={styles.root} className={className}>
+      <ActiveVotingProgress
```
coreyar: This should be a `LabledProgressBar` If we use that component we don't need to have the `voteType` logic.
I made this change but it got force pushed over

src/pages/Proposal/VoteSummary/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+        votedForWei={voteType === 'for' ? votedValueWei : undefined}
+        votedAgainstWei={voteType === 'against' ? votedValueWei : undefined}
+        abstainedWei={voteType === 'abstain' ? votedValueWei : undefined}
+        votedTotalWei={votedTotalWei}
+      />
+      <Button css={styles.button} onClick={onClick} disabled={isDisabled}>
+        {t(`vote.${voteType}`)}
```
coreyar: I am not sure this will get picked up correctly to extract the translation keys. We can change the `voteType` prop to be title. We can drop the other logic around `voteType` if we use the `LabeledProgressBar` 

I had this update but it got force pushed over

src/pages/Proposal/VoteSummary/index.tsx
baomastr: It seems to me that handling texts and colors on the ActiveVotingProgress side should avoid us from duplicating this logic as we need it in several places

src/pages/Proposal/VoteSummary/index.tsx
baomastr: can make an extra function for executing the texts for this cases

src/pages/Proposal/VoteSummary/index.tsx
coreyar: the only duplication should be the assignment of the color which can be done easily using the `successColor` prop. I think the current logic of having a prop for every vote type here is confusing which is why I am requesting the change

coreyar (approved):

coreyar (approved):

PR #622, opened 2022-06-08
Correctly translate missing account error

coreyar:
- Correctly fetch the translation for if a user toggles collateral without a wallet
- add a fallback message in the event something goes wrong and we end up with an undefined translation


therealemjy (approved):
Nice, I like the safeguard

PR #628, opened 2022-06-08
[VEN-226] [VEN-227] [VEN-217] Wire up vote page

coreyar:
Adds final touches to Main Vote page


src/clients/api/queries/useGetVoteDelegates.ts
```diff
@@ -0,0 +1,25 @@
...
+
+type Options = QueryObserverOptions<
+  GetVoteDelegatesOutput,
+  Error,
+  GetVoteDelegatesOutput,
+  GetVoteDelegatesOutput,
+  [FunctionKey.GET_VOTE_DELEGATES, string]
+>;
+
+const useGetVoteDelegates = ({ accountAddress }: { accountAddress: string }, options?: Options) => {
```
therealemjy: This function seems to return the address the voting power has been delegated to.
I'd make it extra that's what this is and change its name to `useGetVoteDelegateAddress`

src/clients/api/queries/getVoteDelegates.ts
```diff
@@ -0,0 +1,21 @@
...
+  xvsVaultContract: XvsVault;
+  accountAddress: string;
+}
+
+export type GetVoteDelegatesOutput = string;
+
+/**
+ *
+ * @param address string (valid Ethereum address)
+ * @returns Delegated address, if no delegation returns null address
```
therealemjy: We could return undefined if the contract returns the null address, this way the front end doesn't need to important the null address from anywhere and we avoid risking that a dev who doesn't have this context simply displays the null address thinking it's valid.

src/constants/functionKey.ts
```diff
@@ -34,6 +34,8 @@ enum FunctionKey {
   GET_VOTE_RECEIPT = 'GET_VOTE_RECEIPT',
   GET_VAI_VAULT_USER_INFO = 'GET_VAI_VAULT_USER_INFO',
   GET_VAI_VAULT_PENDING_XVS = 'GET_VAI_VAULT_PENDING_XVS',
+  GET_VOTE_DELEGATES = 'GET_VOTE_DELEGATES',
```
therealemjy: Following my previous comment, I'd rename this `GET_VOTE_DELEGATE_ADDRESS`.

src/pages/Vote/Governance/index.tsx
```diff
@@ -1,19 +1,95 @@
...
   const styles = useStyles();
   return (
     <div css={styles.root}>
-      <Typography variant="h4">{t('vote.governanceProposals')}</Typography>
+      <div css={[styles.header, styles.bottomSpace]}>
+        <Typography variant="h4">{t('vote.governanceProposals')}</Typography>
+        <div css={styles.createProposal}>
+          <TextButton css={styles.marginless}>{t('vote.createProposal')}</TextButton>
+          <Tooltip title={t('vote.requiredVotingPower')} css={styles.infoIcon}>
+            <Icon size="16px" name="info" />
```
therealemjy: Ideally the icon size should be defined in the `styles` file through CSS

src/pages/Vote/Governance/index.tsx
```diff
@@ -1,19 +1,95 @@
...
+              proposalState={state}
+              endDate={endDate}
+              forVotesWei={forVotesWei}
+              againstVotesWei={againstVotesWei}
+              abstainedVotesWei={abstainedVotesWei}
+            />
+          ),
+        )}
+      </div>
+      {total && (
```
therealemjy, line 64: Do you think you could place a loader somewhere on the page when a new page is being loaded?
I described this UX on your PR for the History screen: I clicked on a new page button in the pagination component, the page scrolled up and for 1-2 seconds nothing happened. The old page data was still showing and I didn't have a visual indication something was happening

src/pages/Vote/VotingWallet/index.tsx
```diff
@@ -64,20 +133,63 @@ export const VoteUi: React.FC<IVoteUiProps> = ({ votingWeight, xvsLockedWei }) =
...
+  const { data: delegate } = useGetVoteDelegates(
+    { accountAddress: accountAddress || '' },
+    { enabled: !!accountAddress },
+  );
+
+  const { data: vaults } = useGetVestingVaults({ accountAddress });
+  const [xvsVault] = vaults.filter(v => v.stakedTokenId === XVS_TOKEN_ID);
+
+  const { mutateAsync: setVoteDelegation } = useSetVoteDelegate({
+    onSettled: () => {},
```
therealemjy: Can be removed

src/pages/Vote/VotingWallet/index.tsx
```diff
@@ -64,20 +133,63 @@ export const VoteUi: React.FC<IVoteUiProps> = ({ votingWeight, xvsLockedWei }) =
...
+  );
+  const { data: delegate } = useGetVoteDelegates(
+    { accountAddress: accountAddress || '' },
+    { enabled: !!accountAddress },
+  );
+
+  const { data: vaults } = useGetVestingVaults({ accountAddress });
+  const [xvsVault] = vaults.filter(v => v.stakedTokenId === XVS_TOKEN_ID);
+
+  const { mutateAsync: setVoteDelegation } = useSetVoteDelegate({
```
therealemjy: We'll want to track the loading state of the mutation (and display a loader on the submit button of the "Delegate/Redelegate voting" modal).
We'll also want to display a toast in case of an error, or a successful transaction modal in case of a success (on top of closing the "Delegate/Redelegate voting" modal). You can find the content for the successful transaction modal here: https://www.figma.com/file/GYJwWWLbC4B8DrJOPcETz3/Venus-2.0?node-id=5723%3A43139 (although I guess it should say "You successfully delegate**d** 8000 BNB)

src/translation/translations/en.json
```diff
@@ -532,11 +534,15 @@
     "governanceProposals": "Governance proposals",
     "pasteYourAddress": "Paste your address",
     "pleasePayAttention": "Please pay attention",
+    "redelegate": "Redelegate",
+    "redelegateVoting": "Redelegate voting",
+    "requiredVotingPower": "You need to have 300k voting power to create proposal",
```
therealemjy: I think "a" is missing in that sentence, right?
```suggestion
    "requiredVotingPower": "You need to have 300k voting power to create a proposal",
```

therealemjy (commented):
Small design issue: the "Voting weight" and "Total locked" blocks aren't vertically aligned.
Screen Shot 2022-06-10 at 09 45 40

src/pages/Vote/Governance/index.tsx
coreyar: I don't know because for things like size on the icon and color on text I think this is the cleanest option. I don't think we should have a wide prop surface for styling but simple styles that will change frequently and are a primary feature of the component make sense as a prop to me. Writing this as styles ends up being more verbose to apply a single line of css.

What do you think if the value was a number instead of a string with a unit? Otherwise, if we really don't want to use this prop we should remove it

src/pages/Vote/Governance/index.tsx
coreyar: yeah, I'll take care of that on the other PR

src/pages/Vote/VotingWallet/index.tsx
coreyar: ahh yeah this was supposed to remind me to take care of your following comment haha

src/translation/translations/en.json
coreyar: that's correct. I just copied from figma

src/pages/Vote/Governance/index.tsx
therealemjy: My main concern is that doing things like this moves us aways from separating the styles from the DOM. I think it makes sense when it's a variant ("primary", "secondary" etc..) because it represents a separate component in a way, but for granular settings like this one in my opinion it should be defined in the CSS.
I understand it makes things more verbose, but hopefully it won't be needed often because the default width/height set in the `Icon` component is the most commonly used one.

And by the way, in this case you can simply remove the `size` prop, because 16px is the default width and height.

src/translation/translations/en.json
therealemjy: Got you!

src/pages/Vote/VotingWallet/index.tsx
```diff
@@ -64,20 +139,86 @@ export const VoteUi: React.FC<IVoteUiProps> = ({ votingWeight, xvsLockedWei }) =
...
+  );
+
+  const { data: vaults } = useGetVestingVaults({ accountAddress });
+  const [xvsVault] = vaults.filter(v => v.stakedTokenId === XVS_TOKEN_ID);
+  const userStakedWei = xvsVault?.userStakedWei || new BigNumber(0);
+
+  const { openSuccessfulTransactionModal } = useSuccessfulTransactionModal();
+  const { mutateAsync: setVoteDelegation, isLoading: isVoteDelegationLoading } = useSetVoteDelegate(
+    {
+      onSettled: data => {
```
therealemjy: It's the `onSuccess` that needs to be used here, otherwise this callbacks will also gets executed on failure in which case `data` would correspond to the error.

We also still need to add an `onError` callback to display a toast on failure.

src/clients/api/queries/getVoteDelegateAddress.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+const getVoteDelegates = async ({
+  xvsVaultContract,
+  accountAddress,
+}: IGetVoteDelegatesInput): Promise<GetVoteDelegatesOutput> => {
+  const resp = await xvsVaultContract.methods.delegates(accountAddress).call();
+  return resp !== NULL_ADDRESS ? resp : undefined;
+};
+
+export default getVoteDelegates;
```
therealemjy:
```suggestion
export default getVoteDelegateAddress;
```

src/clients/api/queries/getVoteDelegateAddress.ts
```diff
@@ -0,0 +1,24 @@
+import { NULL_ADDRESS } from 'constants/address';
+import { XvsVault } from 'types/contracts';
+
+export interface IGetVoteDelegatesInput {
+  xvsVaultContract: XvsVault;
+  accountAddress: string;
+}
+
+export type GetVoteDelegatesOutput = string | undefined;
```
therealemjy:
```suggestion
export interface IGetVoteDelegateAddressInput {
  xvsVaultContract: XvsVault;
  accountAddress: string;
}

export type GetVoteDelegateAddressOutput = string | undefined;
```

src/clients/api/queries/getVoteDelegateAddress.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ *
+ * @param address string (valid Ethereum address)
+ * @returns Delegated address, if no delegation returns undefined
+ */
+const getVoteDelegates = async ({
```
therealemjy:
```suggestion
const getVoteDelegateAddress = async ({
```

src/pages/Vote/VotingWallet/index.tsx
coreyar: When we call `setVoteDelegation` it is done within a try catch and an toast is shown in the catch

therealemjy (approved):

PR #629, opened 2022-06-09
Allow non proposer to cancel if they meet requirements

coreyar:
Kirill followed up on my question in slack and informed that we actually have different cancel requirements than Compound.

The cancel proposal requirements are 
- Proposer
- Guardian
- Anyone if the proposer's voting power is less than govBravo.proposalThreshold()


therealemjy (approved):
Ok fair enough 👍

PR #630, opened 2022-06-09
Chore: apy to apr calculating

baomastr:
- use `venusLensContract.methods.getDailyXVS` for extracting the dailyXvs value


src/context/MarketContext.tsx
```diff
@@ -105,6 +105,7 @@ const MarketContextProvider = ({ children }: $TSFixMe) => {
             await lens.methods.vTokenBalancesAll(vTokenAddresses, account.address).call(),
           );
           xvsBalance = getXvsBalance(balances);
+          // console.log(xvsBalance);
```
coreyar: remove console

src/clients/api/queries/useGetUserMarketInfo.ts
```diff
@@ -268,6 +277,15 @@ const useGetUserMarketInfo = ({
...
     }));
 
+    let dailyXvsDistributionInterestsCents = new BigNumber(0);
+    const dailyXvsTokens = convertWeiToCoins({
+      valueWei: vTokenDailyXvsWei || new BigNumber(0),
+      tokenId: 'xvs',
+    });
+    const xvsAsset = assetList.find((a: Asset) => a.id === 'xvs');
+    const xvsPrice = xvsAsset.tokenPrice;
+    dailyXvsDistributionInterestsCents = dailyXvsTokens.multipliedBy(xvsPrice).times(100);
```
therealemjy: `dailyXvsDistributionInterestsCents` represents the interests for all the tokens, so we shouldn't attach it to every token (the value calculated will always be the same).
Instead I'd suggest calculating this value outside of `useGetUserMarketInfo` and instead in the `MyAccount` component directly since it's only used there.

src/pages/Dashboard/MyAccount/index.tsx
```diff
@@ -36,6 +38,7 @@ const MyAccount: React.FC<IMyAccountProps> = ({
     const yearlyEarningsCents = calculateYearlyEarningsForAssets({
```
therealemjy: To continue on my previous comment, that's where I'd import and call the hook to calculate `dailyXvsDistributionInterestsCents` and passing it to `calculateYearlyEarningsForAssets`.

src/clients/api/queries/useGetVTokenDailyXvs.ts
```diff
@@ -0,0 +1,29 @@
...
+  options?: Options,
+) => {
+  const venusLensContract = useVenusLensContract();
+  return useQuery(
+    [FunctionKey.GET_V_TOKEN_DAILY_XVS, account],
+    () => getVTokenDailyXvs({ account, venusLensContract }),
+    options,
+  );
+};
+export default vTokenDailyXvsWei;
```
therealemjy:
```suggestion
export default useGetVTokenDailyXvsWei;
```

The name of the file should also be changed to `useGetVTokenDailyXvsWei.ts`.

Please make sure to change the names of the types too.

src/clients/api/queries/useGetVTokenDailyXvs.ts
```diff
@@ -0,0 +1,29 @@
...
+>;
+
+const vTokenDailyXvsWei = (
+  { account }: Omit<IGetVTokenDailyXvsInput, 'venusLensContract'>,
+  options?: Options,
+) => {
+  const venusLensContract = useVenusLensContract();
+  return useQuery(
+    [FunctionKey.GET_V_TOKEN_DAILY_XVS, account],
+    () => getVTokenDailyXvs({ account, venusLensContract }),
```
therealemjy:
```suggestion
    () => getVTokenDailyXvsWei({ accountAddress, venusLensContract }),
```

src/clients/api/queries/getVTokenDailyXvs.ts
```diff
@@ -0,0 +1,22 @@
...
+  venusLensContract,
+  account,
+}: IGetVTokenDailyXvsInput): Promise<IGetVTokenDailyXvsOutput> => {
+  const response = await venusLensContract.methods
+    .getDailyXVS(account.toLowerCase(), getContractAddress('comptroller'))
+    .call();
+  return new BigNumber(response);
+};
+
+export default getVTokenDailyXvs;
```
therealemjy:
```suggestion
export default getVTokenDailyXvsWei;
```

The name of the files should also be changed to `getVTokenDailyXvs.ts`.

Please make sure to change the names of the types too.

src/clients/api/queries/getVTokenDailyXvs.ts
```diff
@@ -0,0 +1,22 @@
+import BigNumber from 'bignumber.js';
+import { VenusLens } from 'types/contracts';
+import { getContractAddress } from 'utilities';
+
+export interface IGetVTokenDailyXvsInput {
+  venusLensContract: VenusLens;
+  account: string;
```
therealemjy: To make it clear this is the address we want:
```suggestion
  accountAddress: string;
```

src/clients/api/queries/getVTokenDailyXvs.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export type IGetVTokenDailyXvsOutput = BigNumber;
+
+const getVTokenDailyXvs = async ({
+  venusLensContract,
+  account,
+}: IGetVTokenDailyXvsInput): Promise<IGetVTokenDailyXvsOutput> => {
+  const response = await venusLensContract.methods
+    .getDailyXVS(account.toLowerCase(), getContractAddress('comptroller'))
```
therealemjy: You can call `getContractAddress('comptroller')` from outside the function and assign it to a variable so that it isn't recalculated on every call, since it's a constant.

src/constants/functionKey.ts
```diff
@@ -18,6 +18,7 @@ enum FunctionKey {
   GET_V_TOKEN_BORROW_BALANCE = 'GET_V_TOKEN_BORROW_BALANCE',
   GET_V_TOKEN_BALANCE = 'GET_V_TOKEN_BALANCE',
   GET_V_TOKEN_BALANCES_ALL = 'GET_V_TOKEN_BALANCES_ALL',
+  GET_V_TOKEN_DAILY_XVS = 'GET_V_TOKEN_DAILY_XVS',
```
therealemjy:
```suggestion
  GET_V_TOKEN_DAILY_XVS_WEI = 'GET_V_TOKEN_DAILY_XVS_WEI',
```

src/pages/Dashboard/Modals/BorrowRepay/Repay/index.tsx
```diff
@@ -258,6 +267,7 @@ const Repay: React.FC<IRepayProps> = ({ asset, onClose, isXvsEnabled }) => {
             isXvsEnabled={isXvsEnabled}
             isRepayLoading={isRepayLoading}
             limitTokens={limitTokens.toFixed()}
+            dailyXvsDistributionInterestsCents={dailyXvsDistributionInterestsCents}
```
therealemjy: We're drilling this prop a lot (and some others too, but that's an issue for another day). We can avoid doing that by creating a hook that calls `useGetVTokenDailyXvsWei` and returns `dailyXvsDistributionInterestsCents` so that it can be imported and called everywhere we need it (Borrow/Repay form, Supply/Withdraw form, My Account section).

therealemjy (commented):
The logic is good, but the PR does a lot of prop drilling that we can avoid by creating a hook to calculate `dailyXvsDistributionInterestsCents`, memoize it and return it to be used everywhere it is needed.

src/clients/api/queries/useGetUserMarketInfo.ts
baomastr: it's used in `AccountData` also, but I agree with you. made a separate hook for this value

src/hooks/useDailyXvs.ts
```diff
@@ -0,0 +1,36 @@
...
+import { Asset } from 'types';
+import { XVS_TOKEN_ID } from 'constants/xvs';
+
+export const useDailyXvs = ({
+  accountAddress,
+  assets,
+}: {
+  accountAddress?: string;
+  assets: Asset[];
+}) => {
```
therealemjy: Passing the account address is a good call, however I think `assets` should be grabbed from the hook directly using `useGetAssets`. This way this hook can be used from anywhere without needing to receive anything else than the account address.
What do you think?

src/hooks/useDailyXvs.ts
```diff
@@ -0,0 +1,36 @@
...
+    const dailyXvsTokens = convertWeiToCoins({
+      valueWei: vTokenDailyXvsWei || new BigNumber(0),
+      tokenId: XVS_TOKEN_ID,
+    });
+    const xvsAsset = assets.find((a: Asset) => a.id === 'xvs');
+    const xvsPrice = xvsAsset?.tokenPrice || new BigNumber(0);
+    return {
+      dailyXvsDistributionInterestsCents: dailyXvsTokens.multipliedBy(xvsPrice).times(100),
+    };
+  }, [JSON.stringify(vTokenDailyXvsWei), accountAddress]);
```
therealemjy: `assets` needs to be in the list of dependencies:
```suggestion
  }, [JSON.stringify(vTokenDailyXvsWei), accountAddress, JSON.stringify(assets)]);
```

src/hooks/useDailyXvs.ts
```diff
@@ -0,0 +1,36 @@
...
+    { enabled: !!accountAddress },
+  );
+
+  const { dailyXvsDistributionInterestsCents } = useMemo(() => {
+    const dailyXvsTokens = convertWeiToCoins({
+      valueWei: vTokenDailyXvsWei || new BigNumber(0),
+      tokenId: XVS_TOKEN_ID,
+    });
+    const xvsAsset = assets.find((a: Asset) => a.id === 'xvs');
+    const xvsPrice = xvsAsset?.tokenPrice || new BigNumber(0);
```
therealemjy: nit: let's follow the pattern of adding the unit to the variable name:
```suggestion
    const xvsPriceDollars = xvsAsset?.tokenPrice || new BigNumber(0);
```

src/hooks/useDailyXvs.ts
```diff
@@ -0,0 +1,36 @@
...
+  assets: Asset[];
+}) => {
+  const { data: vTokenDailyXvsWei, isLoading } = useGetVTokenDailyXvs(
+    { accountAddress: accountAddress || '' },
+    { enabled: !!accountAddress },
+  );
+
+  const { dailyXvsDistributionInterestsCents } = useMemo(() => {
+    const dailyXvsTokens = convertWeiToCoins({
+      valueWei: vTokenDailyXvsWei || new BigNumber(0),
```
therealemjy: While `vTokenDailyXvsWei` is `undefined`, I think we should return `dailyXvsDistributionInterestsCents` as `undefined` too, this way we're consistent with our logic.

src/clients/api/queries/useGetVTokenDailyXvs.ts
```diff
@@ -0,0 +1,29 @@
...
+  options?: Options,
+) => {
+  const venusLensContract = useVenusLensContract();
+  return useQuery(
+    [FunctionKey.GET_V_TOKEN_DAILY_XVS_WEI, accountAddress],
+    () => getVTokenDailyXvsWei({ accountAddress, venusLensContract }),
+    options,
+  );
+};
+export default useGetVTokenDailyXvsWei;
```
therealemjy: Following my previous comment, I'd rename this hook `useGetDailyXvsWei` and the file `useGetDailyXvsWei.ts`.

src/clients/api/queries/getVTokenDailyXvsWei.ts
```diff
@@ -0,0 +1,23 @@
...
+  venusLensContract,
+  accountAddress,
+}: IGetVTokenDailyXvsWeiInput): Promise<IGetVTokenDailyXvsWeiOutput> => {
+  const response = await venusLensContract.methods
+    .getDailyXVS(accountAddress, comptrollerAddress)
+    .call();
+  return new BigNumber(response);
+};
+
+export default getVTokenDailyXvsWei;
```
therealemjy: I think we should rename this function `getDailyXvsWei`.
The name `getVTokenDailyXvsWei` might be confusing and make one think we are returning the daily XVS associated to a specific token, but in this case we return the daily XVS obtained from all the tokens.

You'll also need to write some tests to cover this function.

src/pages/Dashboard/Modals/SupplyWithdraw/SupplyWithdrawForm.tsx
```diff
@@ -73,6 +76,11 @@ export const SupplyWithdrawContent: React.FC<ISupplyWithdrawFormUiProps> = ({
     ? calculateNewBalance(asset.supplyBalance, amount)
     : undefined;
 
+  const { dailyXvsDistributionInterestsCents } = useVTokenDailyXvs({
```
therealemjy: Can you add a TODO comment about needing to handle the loading state please?

src/pages/Dashboard/Modals/BorrowRepay/AccountData/index.tsx
```diff
@@ -32,15 +33,17 @@ const AccountData: React.FC<IAccountDataProps> = ({
...
 
   // TODO: handle loading state (see https://app.clickup.com/t/2d4rcee)
   const {
     data: { assets, userTotalBorrowBalanceCents, userTotalBorrowLimitCents },
   } = useGetUserMarketInfo({
-    accountAddress: account?.address,
+    accountAddress,
   });
 
+  const { dailyXvsDistributionInterestsCents } = useVTokenDailyXvs({ assets, accountAddress });
```
therealemjy: Can you add a TODO comment about needing to handle the loading state please?

src/hooks/useDailyXvs.ts
baomastr: nice point

baomastr: do you mean `useGetAssetsInAccount` here?

baomastr: removed props for hook, used `getMarketsData` for `xvsPriceDollars`

therealemjy (commented):
The code looks great, however it's changed the empty state of the "My account" section when no account is connected to the dApp:
Previously:
Screen Shot 2022-06-14 at 16 40 56

Now:
Screen Shot 2022-06-14 at 16 41 15

therealemjy (approved):

PR #631, opened 2022-06-10
Transaction fetching improvements

coreyar:
[Update transactions to use isLoading](https://github.com/VenusProtocol/venus-protocol-interface/commit/793a081a5afcc5b2cc6b3a44e89efb6f6ea1681c) IsFetching applies on every query while isLoading is only true when the cache is empty
 
[Remove Transaction model](https://github.com/VenusProtocol/venus-protocol-interface/commit/533f5040fe2b1bb3e99e3b71d5551d810121715d) and format data with function to be in compliance with our data layer
 
[Zero index pagination](https://github.com/VenusProtocol/venus-protocol-interface/commit/4fd6476171e6a08d26890a060241b9ad46657de0) some of this change is actually in the first commit because I patched wrong 😬 
 
[Fix history table rendering issues](https://github.com/VenusProtocol/venus-protocol-interface/commit/ebf1b3f9ff745bc5f0c45f458dff747e17a3fa8a) Cleaner solution for when the vTokenAddress is missing and fixes an issue where a 0 was rendered in place of pagination


src/clients/api/queries/getTransactions/formatTransaction.ts
```diff
@@ -0,0 +1,20 @@
...
+const formatTransaction = ({
+  amount,
+  createdAt,
+  updatedAt,
+  category,
+  event,
+  ...rest
+}: ITransactionResponse) => ({
+  ...rest,
+  amount: new BigNumber(amount),
```
therealemjy: I'd suggest adding the unit:
```suggestion
  amountWei: new BigNumber(amount),
```

src/pages/History/HistoryTable/index.tsx
```diff
@@ -46,133 +45,135 @@ export const HistoryTableUi: React.FC<IHistoryTableProps> = ({ transactions, isF
...
-              {txn.event}
-            </Typography>
-          </div>
-          <div css={[styles.cards, styles.cardTitle]}>
-            <div css={styles.typeCol}>
-              <Icon name={getTokenIdFromVAddress(txn.vTokenAddress) as TokenId} css={styles.icon} />
+  const rows: TableProps['data'] = transactions.map(txn => {
+    const tokenId = txn.vTokenAddress
+      ? (getTokenIdFromVAddress(txn.vTokenAddress) as TokenId)
+      : XVS_TOKEN_ID;
```
therealemjy: We can memoize `rows` to optimize things a bit

src/pages/History/index.tsx
```diff
@@ -36,27 +35,27 @@ export const HistoryUi: React.FC<IHistoryUiProps> = ({
...
 
 const History: React.FC = () => {
   const { account } = useContext(AuthContext);
   const accountAddress = account?.address;
-  const [currentPage, setCurrentPage] = useState(1);
+  const [currentPage, setCurrentPage] = useState(0);
   const [eventType, setEventType] = useState<TransactionEvent | typeof ALL_VALUE>(ALL_VALUE);
   const [showOnlyMyTxns, setShowOnlyMyTxns] = useState(false);
-  const { data: { transactions, total, limit } = { transactions: [] }, isFetching } =
+  const { data: { transactions, total, limit } = { transactions: [] }, isLoading } =
```
therealemjy: So I get why you renamed `isFetching` to `isLoading` on the `Table` component, however I think in this case passing it `isFetching` is the right approach.
When the page changes, displaying the loading indicator is great because it shows users something is happening. Otherwise, when I press on a button in the navigation nothing really happens and the current data doesn't change until the new data has been loaded: as a user I could think this is a bug.

therealemjy (commented):
Looks good, just a few small comments

therealemjy (approved):

PR #638, opened 2022-06-13
[VEN-289] Filter XVS from borrow market

coreyar:
This PR removes XVS from the borrow market  because  XVS borrowing is not allowed and the transaction will fail.

It refactors the filtering of the supply market as well so that they are symmetrical

### DIFF
--- a/src/pages/Dashboard/index.tsx
+++ b/src/pages/Dashboard/index.tsx
@@ -3,6 +3,7 @@ import React, { useMemo } from 'react';
 import BigNumber from 'bignumber.js';
 import { AuthContext } from 'context/AuthContext';
 import { useGetUserMarketInfo } from 'clients/api';
+import { XVS_TOKEN_ID } from 'constants/xvs';
 import { Asset } from 'types';
 
 import MyAccount from './MyAccount';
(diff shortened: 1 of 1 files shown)


src/pages/Dashboard/index.tsx
```diff
@@ -31,16 +32,16 @@ const DashboardUi: React.FC<IDashboardUiProps> = ({
...
 
-          if (curr.borrowBalance.isZero()) {
-            acc.borrowMarketAssets.push(curr);
-          } else {
+          if (curr.borrowBalance.isGreaterThan(0)) {
             acc.borrowingAssets.push(curr);
+          } else if (curr.id !== XVS_TOKEN_ID) {
+            acc.borrowMarketAssets.push(curr);
           }
           return acc;
```
therealemjy, line 50: Although technically users can't borrow XVS and so this logic is enough, I think we should make it a bit more obvious that we don't want to display XVS on the borrow market at all (and so neither on the "Borrowing" table):
```suggestion
          if (curr.id === XVS_TOKEN_ID) {
            return acc;
          }

          if (curr.borrowBalance.isGreaterThan(0)) {
            acc.borrowingAssets.push(curr);
          } else {
            acc.borrowMarketAssets.push(curr);
          }
          return acc;
```

therealemjy (approved):

src/pages/Dashboard/index.tsx
coreyar: I left it in the borrowing table in case users had outstanding balances. This would remove it once the outstanding balances have been paid

therealemjy: We should update the `Borrow` modal so the "Borrow" tab doesn't display then

PR #641, opened 2022-06-13
[VEN-187] Markdown components

coreyar:
Add basic markdown components for viewing and editing markdown


therealemjy:
Giving people the possibility to write down any markdown exposes us to the issue Kirill highlighted a few weeks back, where they can break our design:
Screen Shot 2022-06-14 at 16 32 31

Perhaps we could use a rich text editor instead? Something like this library does:
https://jpuri.github.io/react-draft-wysiwyg

coreyar:
> Giving people the possibility to write down any markdown exposes us to the issue Kirill highlighted a few weeks back, where they can break our design: Screen Shot 2022-06-14 at 16 32 31
> 
> Perhaps we could use a rich text editor instead? Something like this library does: https://jpuri.github.io/react-draft-wysiwyg

This is mitigated by the accepted tags in the markdown renderer but I agree a Rich text editor is even safe

Looks like the WYSIWYG also supports images. I'll dig to see how we can prevent users from using options not listed in the toolbar of either

coreyar:
> Giving people the possibility to write down any markdown exposes us to the issue Kirill highlighted a few weeks back, where they can break our design: Screen Shot 2022-06-14 at 16 32 31
> 
> Perhaps we could use a rich text editor instead? Something like this library does: https://jpuri.github.io/react-draft-wysiwyg

Looks like the WYSIWYG also supports images

therealemjy (commented):
It looks good to me code wise, I'd just change the file structure a bit.
For me, Editor and Viewer should have their own folders because they are separate components. We can keep them under the `Markdown` folder. This means `styles.scss` can be moved inside the editor's folder, while `styles.ts` would be renamed `sharedStyles.ts`

therealemjy (approved):

PR #646, opened 2022-06-13
Create proposal UI

coreyar:
Adds UI for  creating a proposal.

The designs are still being updated but the required elements and logic are in place. Once the designs are updated they will likely just need to be tweaked


coreyar:
This PR is waiting on the markdown components

src/pages/Vote/CreateProposalModal/proposalSchema.ts
```diff
@@ -0,0 +1,35 @@
...
+  actions: yup
+    .array()
+    .of(
+      yup.object({
+        address: yup
+          .string()
+          .required()
+          .test('isPositive', ErrorCode.ACTION_ADDRESS_NOT_VALID, value =>
+            web3.utils.isAddress(value as string),
+          ),
```
therealemjy:
```suggestion
          .test('isValidAddress', ErrorCode.ACTION_ADDRESS_NOT_VALID, value =>
            web3.utils.isAddress(value as string),
          ),
```

therealemjy (commented):
A few observations:
- any text field input shows as invalid with a red border after entering the first character. If we enter more than that, it shows as valid
Screen Shot 2022-06-16 at 14 22 06

- on step 2, the spacing between the form and the submit button is too small:
Screen Shot 2022-06-16 at 14 23 40

- The icon used for the accordion is different from the one used in the designs:
dApp:
Screen Shot 2022-06-16 at 14 29 05

Design:
Screen Shot 2022-06-16 at 14 25 23

- the spacing between the submit button and the edge of the modal should be smaller:
Screen Shot 2022-06-16 at 14 26 41

- on step 3, when I start entering an address it shows as invalid with a red border until I've entered a full address. I thing this is a negative user experience and we should instead validate the input on submit (you could change the form so it doesn't validate on change, then when pressing the submit button to move to the next step it would asynchronously validate the form and shows incorrect fields)

coreyar:
> A few observations:
> 
> ...


Thanks for the review, this PR is still in progress I should  have moved it back to draft after the designs were updated

therealemjy:
Ok sure! Let me know when you want me to review

therealemjy (commented):
Code looks good.
I spotted a few minor issues with the display on step 3:
- the cross icon on accordions isn't aligned with the edge of the accordion like in the designs (you could use a negative margin to keep the hit box of the button the same)
- the cross icon is bigger than in the designs. I saw you're using the default width and height from the `Icon` component, 16px, while the designs use a width and height of 24px while displaying a small icon. My guess is they used a different icon for this one specifically. Perhaps a solution here is to set a smaller size for the icon so that it matches the actual size in the designs
- closed accordions have a bottom margin of 28px against 12px in the designs
- closed accordions have their title displayed in white, while they're displayed in grey in the designs
- there's an additional 12px margin at the bottom of the last action compared to the designs
- the submit button says "Next step" but should say "Confirm"

In the final "Confirmation" step, the actions section overflows with a long enough signature: 
Screen Shot 2022-06-21 at 17 48 33

therealemjy (approved):

PR #647, opened 2022-06-13
Add errors to translation strings and improve wallet mismatch error

coreyar:
When testing connecting wrong wallets it came to my attention that the error message could be more useful.

In the process I added the errors to the translation strings.

We can refactor this to use our VError VEN-292

### DIFF
--- a/src/translation/translations/en.json
+++ b/src/translation/translations/en.json
@@ -574,6 +574,11 @@
     "binanceChainWallet": "Binance Chain Wallet",
     "braveWallet": "Brave Wallet",
     "coinbaseWallet": "Coinbase Wallet",
+    "errors": {
+      "authorizeAccess": "You need to authorize access to your account",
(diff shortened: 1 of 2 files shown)


src/translation/translations/en.json
```diff
@@ -574,6 +574,11 @@
     "binanceChainWallet": "Binance Chain Wallet",
     "braveWallet": "Brave Wallet",
     "coinbaseWallet": "Coinbase Wallet",
+    "errors": {
+      "authorizeAccess": "You need to authorize access to your account",
+      "noProvider": "If you are connecting an external wallet, make sure you have a provider such as Metamask or compatible wallet such as Brave",
```
therealemjy: nit:
```suggestion
      "noProvider": "If you are connecting an external wallet, make sure you have a provider such as MetaMask or compatible wallet such as Brave",
```

therealemjy (approved):

PR #650, opened 2022-06-14
[VEN-228] Claim vault rewards mutations

therealemjy:
Add mutations to claim rewards on various vault types.


src/clients/api/mutations/useClaimVaiVaultReward.ts
```diff
@@ -0,0 +1,41 @@
...
+import { useVaiVaultContract } from 'clients/contracts/hooks';
+
+type Options = MutationObserverOptions<
+  ClaimVaiVaultRewardOutput,
+  Error,
+  Omit<IClaimVaiVaultRewardInput, 'vaiVaultContract'>
+>;
+
+const useClaimVaiVaultReward = (options?: Options) => {
+  const vaiVaultContract = useVaiVaultContract();
```
coreyar, line 19: Should we be using the ProxyContract here? 

We are checking for failure events but this contract doesn't send failure events. The Proxy contract does

therealemjy: I don't think so, the current implementation doesn't use the proxy.
I can remove the checks in that case

coreyar: In that case we don't have to check for the Failure event. The VAiVaultContract doesn't send the `fail` or `failOpaque` events. It doesn't hurt anything though

therealemjy: Sounds good, I've removed the check

src/errors/transactionErrors.ts
```diff
@@ -47,6 +55,12 @@ export const checkForTokenTransactionError = (receipt: TransactionReceipt) =>
...
-    VAIControllerErrorReporterError,
-    VAIControllerErrorReporterFailureInfo,
+    VaiControllerErrorReporterError,
+    VaiControllerErrorReporterFailureInfo,
   );
+
+export const checkForVaiVaultTransactionError = (receipt: TransactionReceipt) =>
+  checkForTransactionError(receipt, VaiVaultErrorReporterError, VaiVaultErrorReporterInfo);
+
+export const checkForXvsVaultTransactionError = (receipt: TransactionReceipt) =>
```
coreyar: This is actually for checking the vaultProxyContract because the Vault Contract doesn't send these events

coreyar (approved):

PR #653, opened 2022-06-14
Fix sorting on Table

therealemjy:
- converts values to number or use them as string based on their content, then use that sorting rows


src/pages/History/HistoryTable/index.tsx
```diff
@@ -51,6 +51,8 @@ export const HistoryTableUi: React.FC<IHistoryTableProps> = ({ transactions, isF
         const tokenId =
           (txn.vTokenAddress && getTokenIdFromVAddress(txn.vTokenAddress)) || XVS_TOKEN_ID;
 
+        console.log(txn.event);
```
coreyar: remove console log

coreyar (approved):

coreyar (approved):

PR #658, opened 2022-06-14
Add stepper component

coreyar:
This adds a stepper component for the Proposal detail screen. 

I did not use the MaterialMui because I spent too much time trying to work with it only to find that my changes weren't going to play nicely with its responsiveness and the only way to make their component responsive was to listen to window  change events.

Thankfully, I was able to rewrite from scratch it in short time


src/pages/Proposal/ProposalSummary/Stepper/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+                )}
+                <Typography variant="small1" css={styles.labelText({ completed })}>
+                  {step.getLabel()}
+                  {step.getTimestamp() ? (
+                    <Typography variant="tiny" css={styles.dateTablet}>
+                      {t('voteProposalUi.statusCard.dateOnly', { date: step.getTimestamp() })}
+                      <br />
+                      {t('voteProposalUi.statusCard.timeOnly', { date: step.getTimestamp() })}
+                    </Typography>
+                  ) : null}
```
therealemjy: nit: you can return false and React won't render it in the DOM:
```suggestion
                  {!!step.getTimestamp() && (
                    <Typography variant="tiny" css={styles.dateTablet}>
                      {t('voteProposalUi.statusCard.dateOnly', { date: step.getTimestamp() })}
                      <br />
                      {t('voteProposalUi.statusCard.timeOnly', { date: step.getTimestamp() })}
                    </Typography>
                  )}
```

src/pages/Proposal/ProposalSummary/Stepper/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+                </Typography>
+              </div>
+              <Typography variant="tiny" css={styles.dateDefault}>
+                {step.getTimestamp()
```
therealemjy: nit: same thing here about returning `false`

src/pages/Proposal/ProposalSummary/Stepper/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+                  ? t('voteProposalUi.statusCard.dateAndTime', { date: step.getTimestamp() })
+                  : null}
+              </Typography>
+            </div>
+            {idx + 1 !== steps.length ? <div css={styles.connector} /> : null}
```
therealemjy: nit: same thing here:
```suggestion
            {idx + 1 !== steps.length && <div css={styles.connector} />}
```

src/pages/Proposal/ProposalSummary/Stepper/styles.ts
```diff
@@ -0,0 +1,145 @@
...
+    `,
+    numberIconContainer: css`
+      background-color: transparent;
+      border: 1px solid ${theme.palette.text.secondary};
+    `,
+    connector: css`
+      border-left: 1px solid ${theme.palette.text.secondary};
+      height: ${theme.spacing(4)};
+      width: 0;
+      margin-left: calc(${theme.spacing(3)} - 0.5px);
```
therealemjy, line 124: Out of curiosity, `-0.5px` is to account for the border right?

src/pages/Proposal/ProposalSummary/Stepper/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  Defeated: { active: 3, error: true },
+  Succeeded: { active: 3, error: undefined },
+  Canceled: { active: 3, error: true },
+  Expired: { active: 3, error: true },
+
+  Queued: { active: 4, error: undefined },
+
+  Executed: { active: 5, error: undefined },
+};
```
therealemjy: It looks like `error` is never used, so you could use a function that returns a step index based on a proposal state instead:
```suggestion
const getStepIndex = (proposalState: ProposalState) => {
  switch (proposalState) {
    case 'Pending':
      return 1;
    case 'Active':
      return 2;
    case 'Defeated':
    case 'Succeeded':
    case 'Canceled':
    case 'Expired':
      return 3;
    case 'Queued':
      return 3;
    default:
      return 5;
  }
};
```

src/pages/Proposal/ProposalSummary/Stepper/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+        completedIcon: () => (
+          <span css={[styles.iconContainer, styles.markIconContainer]}>
+            <Icon name="mark" css={styles.markIcon} />
+          </span>
+        ),
+      },
+    ],
+    [JSON.stringify(proposal)],
+  );
+  const activeStep = StepStates[proposal.state];
```
therealemjy: and so following my previous comment:
```typescript
const activeStepIndex = getStepIndex(proposal.state);
```

src/pages/Proposal/ProposalSummary/Stepper/styles.ts
coreyar: correct

src/pages/Proposal/ProposalSummary/Stepper/index.tsx
coreyar: ahh yeah that was left over from trying to use material ui. I'll update this

therealemjy (approved):

PR #661, opened 2022-06-15
Feature: Description component for proposal

baomastr:
https://app.clickup.com/t/2ggz5qu

- Upgraded react-markdown version
- Added React-Markdown for proposal description


package.json
```diff
@@ -57,7 +57,7 @@
     "react-countdown": "^2.3.2",
     "react-dom": "npm:@hot-loader/react-dom@^17.0.1",
     "react-i18next": "^11.16.5",
-    "react-markdown": "^4.3.1",
+    "react-markdown": "^7.1.2",
```
coreyar, line 60: We are not going to continue using this library. There is a PR to replace it and a thread in slack about this

src/pages/Proposal/Description/index.tsx
```diff
@@ -1,15 +1,26 @@
 /** @jsxImportSource @emotion/react */
 import React from 'react';
+import ReactMarkdown from 'react-markdown';
```
coreyar, line 3: We should wrap this as a component. I've done that in this PR - https://github.com/VenusProtocol/venus-protocol-interface/pull/641

src/pages/Proposal/Description/index.tsx
```diff
@@ -1,15 +1,26 @@
...
-export const Description: React.FC<IDescriptionSummary> = ({ className }) => {
+export const Description: React.FC<IDescriptionSummary> = ({ className, descriptionMarkdown }) => {
   const styles = useStyles();
-  return <Paper css={styles.root} className={className} />;
+  return (
+    <Paper css={styles.root} className={className}>
+      <Typography variant="h4" color="textSecondary">
+        Description
+      </Typography>
+      <ReactMarkdown>{descriptionMarkdown}</ReactMarkdown>
```
coreyar, line 21: We will have to versions of descriptions we need to account for.
v1 IS plain string when trying to parse the json we’ll fallback and assume it is v1
v2 would be as follows
```{
  version: 'v2',
  title: '',
  description: '',
  forDescription: '',
  againstDescription: '',
  abstainDescription: ''
}```

coreyar, line 21: In addition to parsing the description, we want to list the actions in the description

baomastr:
rebased in https://github.com/VenusProtocol/venus-protocol-interface/pull/664

PR #662, opened 2022-06-16
[VEN-228] Fetch VRT vault

therealemjy:
- add necessary query functions to fetch VRT vault
- update `useGetVaults` hook to add VRT vault fetching


src/clients/api/queries/useGetVaults/useGetVrtVault.ts
```diff
@@ -0,0 +1,98 @@
...
+import { DAYS_PER_YEAR } from 'constants/daysPerYear';
+import { BLOCKS_PER_DAY } from 'constants/bsc';
+import { getContractAddress } from 'utilities';
+
+export interface UseGetVrtVaultOutput {
+  isLoading: boolean;
+  data: Vault | undefined;
+}
+
+const VRT_TOKEN_ID = 'vrt';
```
coreyar: This is going to be duplicated throughout the app. I think we  should create a VRT contant file like we have one for XVS. Currently it lives in `pages/ConvertVrt/constants`,

therealemjy: Actually, I just updated it to use the `TOKENS` constant directly.
By creating a separate constant for a specific token, we're creating a loop:
```typescript
const TOKENS = {};

const getToken = (tokenId) => TOKENS[tokenId];

const VRT_TOKEN_ID = getToken('vrt').id;
```

When in fact we can just write:
```typescript
const VRT_TOKEN_ID = TOKENS.vrt.id;
```

coreyar: great!

coreyar (approved):

PR #664, opened 2022-06-16
Feat: proposal details

baomastr:
ex https://github.com/VenusProtocol/venus-protocol-interface/pull/661

### DIFF
--- a/src/components/Vote/VoteOverview/ProposalDetail.tsx
+++ b/src/components/Vote/VoteOverview/ProposalDetail.tsx
@@ -3,7 +3,6 @@ import ReactMarkdown from 'react-markdown';
 import styled from 'styled-components';
 import { Card } from 'components/Basic/Card';
 import { Label } from 'components/Basic/Label';
-import { uid } from 'react-uid';
 import { ProposalInfo } from '../types';
 
 const ProposalDetailWrapper = styled.div`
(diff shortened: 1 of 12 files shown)


therealemjy:
Now that the new markdown component has been added to the codebase, you can use it here :)

therealemjy (commented):
A few things:
- the font-family used in the viewer is incorrect. On my machine I get the Apply system font instead of proxima nova
- the designs show the content can only reach a maximum width of 800 pixels
- the paddings set for the container are incorrect: the spacing between the content and the top and bottom edges of the container should be 32px instead of 24

src/pages/Proposal/Description/index.tsx
```diff
@@ -1,15 +1,28 @@
...
-export const Description: React.FC<IDescriptionSummary> = ({ className }) => {
+export const Description: React.FC<IDescriptionSummary> = ({ className, description }) => {
   const styles = useStyles();
-  return <Paper css={styles.root} className={className} />;
+  const { t } = useTranslation();
+
+  return (
+    <Paper css={styles.root} className={className}>
+      <Typography variant="h4" color="textSecondary">
+        {t('history.description')}
```
therealemjy: That translation key seems incorrect to me, where does "history" come from?

src/components/Vote/VoteOverview/ProposalDetail.tsx
```diff
@@ -49,17 +49,21 @@ function ProposalDetail({ proposalInfo }: ProposalDetailProps) {
           <Label size="14">
             {/**/}
```
therealemjy: This needs to be removed

src/components/Vote/VoteOverview/ProposalDetail.tsx
```diff
@@ -49,17 +49,21 @@ function ProposalDetail({ proposalInfo }: ProposalDetailProps) {
...
             {(proposalInfo.actions || []).map((s: $TSFixMe) => (
-              <ReactMarkdown className="proposal-detail" source={s.title} key={uid(s)} />
+              <ReactMarkdown className="proposal-detail" key={uid(s)}>
```
therealemjy: We need to get rid of `uid` there, otherwise we'll have the same issue we had with the vaults a while back and a different key will be generated every time `proposalInfo.actions` changes.

PR #668, opened 2022-06-16
[VEN-190] Proposal summary UI

coreyar:
## Jira ticket(s)

VEN-190

## Changes
There are a couple minor outstanding design questions about the stepper responsiveness.
The mock data in this PR is better at representing the different proposal states but it isn't perfect. That will be covered the wire up PR


src/components/v2/BscLink/index.tsx
```diff
@@ -27,12 +28,8 @@ export const BscLink: React.FC<IBscLinkProps> = ({ hash, className, urlType }) =
         variant="small1"
         css={styles.text}
       >
-        <Trans
-          i18nKey="bscLink.content"
-          components={{
-            Icon: <Icon name="open" css={styles.icon} />,
-          }}
-        />
+        {text || t('bscLink.content')}
```
therealemjy, line 31: If `text` is an empty string, it will still render:
```suggestion
        {!!text || t('bscLink.content')}
```

src/pages/Proposal/ProposalSummary/Stepper/index.stories.tsx
```diff
@@ -14,59 +14,112 @@ const Template: Story<IStepperProps> = (args: IStepperProps) => <Stepper {...arg
...
+  cancelDate: undefined,
+  state: 'Active',
+  startDate: undefined,
+  executedDate: undefined,
+  endDate: new Date(1654000009000),
+  queuedDate: undefined,
 };
+// .args = {
+//   proposal: { ...proposals[0], state: 'Defeated', executedDate: undefined, queuedDate: undefined },
+// };
```
therealemjy: Do you need this for your next PR? Otherwise I'd say let's remove it (and the other pieces of code commented out)

src/pages/Proposal/ProposalSummary/index.tsx
```diff
@@ -1,15 +1,135 @@
...
+        <div css={styles.content}>
+          <div>
+            <Typography variant="h3" css={styles.title}>
+              {title}
+            </Typography>
+            {transactionHash && (
+              <BscLink text={createdTxHash} urlType="tx" hash={transactionHash} />
+            )}
+          </div>
+          <div>{updateProposalButton}</div>
```
therealemjy, line 101: When `updateProposalButton` is `null`, an empty `div` will still be rendered in the DOM:
```suggestion
          {updateProposalButton && <div>{updateProposalButton}</div>}
```

You might need it to handle aligning the children, in this case feel free to ignore this comment.

src/pages/Proposal/ProposalSummary/index.tsx
```diff
@@ -1,15 +1,135 @@
...
+      transactionHash = queuedTxHash;
+      break;
+    case 'Defeated':
+      transactionHash = endTxHash;
+      break;
+    case 'Executed':
+      transactionHash = executedTxHash;
+      break;
+    default:
+      updateProposalButton = null;
```
therealemjy: You don't need this default statement since that's already the value you assigned to `updateProposalButton`.

src/pages/Proposal/ProposalSummary/index.tsx
```diff
@@ -1,15 +1,135 @@
...
+    cancelTxHash,
+    queuedDate,
+    queuedTxHash,
+    executedDate,
+    executedTxHash,
+    endTxHash,
+    endDate,
+  } = proposal;
+
+  let updateProposalButton = null;
```
therealemjy: As a rule of thumb, I try to avoid using `null` as much as possible because of all its weird behaviors (like `typeof null === true` returned `true` 😅).
You can use undefined:
```suggestion
  let updateProposalButton: React.ReactNode | undefined;
```

therealemjy (commented):
Some small design issues:
- the "Proposal history" title has the wrong color and font weight
- on tablet and mobile, there's an additional spacing between the stepper and the left border of its container (you could use a negative margin to make the 1px delimiter stick to the edge of the container like on the designs)
Screen Shot 2022-06-17 at 08 49 05
- there's a short resolution span where elements get a bit cranked up:
Screen Shot 2022-06-17 at 08 53 36
I saw you asked the design team about a similar situation (might even be this exact one) on Slack so my assumption is you're aware of it, but I'm just putting commenting it here so we have a record on the concerned PR too

src/pages/Proposal/ProposalSummary/index.tsx
coreyar: You cannot render undefined which is why I chose null here.

Edit: You cannot return undefined from a component but you can render undefined as a nested element in the component so I can update this. React 18 is removing this all together

src/pages/Proposal/ProposalSummary/index.tsx
coreyar: I included the default case for completeness out of habit following the eslint rule `default-case`. The motivation being explicitly defining it removes doubt about whether it was forgotten or not. Happy to remove it

src/pages/Proposal/ProposalSummary/Stepper/index.stories.tsx
coreyar: Removed

src/components/v2/BscLink/index.tsx
coreyar: That is not correct. A logical operator of  `'' || 'bar'` will return `'bar'`
A logical operator of `!!'foo' || 'bar'` will return `true`

src/pages/Proposal/ProposalSummary/index.tsx
coreyar: Actually, turns out we also have this eslint rule because I get errors when I remove it so I'm going to leave it in.

src/components/v2/BscLink/index.tsx
therealemjy: Yes, you're completely right. I confused myself by thinking about an AND operator. My bad!

therealemjy (approved):

PR #670, opened 2022-06-17
[VEN-247] Vault - Stake modal wire up

therealemjy:
## Jira ticket(s)

VEN-247

## Changes

- add mutation functions to handle staking in all types of vault: XVS (non vesting), VRT and VAI
- add optional `poolIndex` field to `Vault` type
- add `useStakeWeiInVault` mutation to handle staking in any type of vault
- add `useIsMounted` hook to check whether a component is mounted
- update `TransactionForm` to handle asynchronous submission with success and error handling
- wire up "Stake" button of `VaultItem`
- create `ActionModal` component to be used for all vault actions (withdraw, stake, claim reward)
- create `StakeModal` component to handle staking in any type of vault

## Minor changes

- rename `isOpened` prop to `isOpen` on `Modal`
- move `truncateAddress` util reference to `utilities/index.tsx`


src/hooks/useStakeWeiInVault.ts
```diff
@@ -0,0 +1,75 @@
...
+        fromAccountAddress: accountAddress,
+        amountWei,
+      });
+    }
+
+    // This cose should never be reached, but just in case we throw a generic
+    // internal error
+    throw new VError({
+      type: 'unexpected',
+      code: t('errors.somethingWentWrong'),
```
coreyar: I am not sure how we currently do this but I wouldn't think something labeled code would be translatable. I would expect it to be static so that it could be checked with equality statements.

baomastr (approved):

coreyar (approved):

src/hooks/useStakeWeiInVault.ts
therealemjy: You're right. I copied this from somewhere else but that's wrong, good catch.
I've fixed it here as well as where I copied it from.

PR #679, opened 2022-06-21
Bugfix: IOS menu

baomastr:
## Jira ticket(s)

VEN-273

## Changes

- menu icon fix
- menu position fix
- Venus logo fix


src/components/v2/Layout/Sidebar/index.tsx
```diff
@@ -24,6 +24,7 @@ export const SidebarUi: React.FC = () => {
   const open = Boolean(anchorEl);
   const { t } = useTranslation();
   const styles = useStyles();
+  const { pathname } = useLocation();
```
therealemjy: Do you need the key to change when the location changes?
If you just need a unique key, you can use `useUID` from `react-uid`

therealemjy (commented):
Looks good.
I found a weird edge case where the logo will disappear. On iOS, if while on the dashboard you open the hamburger menu, then press on "Dashboard" the logo disappears.

src/components/v2/Layout/Sidebar/index.tsx
baomastr: yeah. we lose the logo every time after changing the route. so the solution with pathname works nice here.

baomastr: case with root route is fixed with useUID. thanks for catch

src/components/v2/Layout/Sidebar/index.tsx
```diff
@@ -24,6 +25,8 @@ export const SidebarUi: React.FC = () => {
...
+  const logoKey = pathname || useUID();
```
therealemjy: You can't conditionally render a hook, it needs to render all the time.
In this case however it will never render, because `pathname` is never falsy.

src/components/v2/Layout/Sidebar/index.tsx
therealemjy: I'm still seeing the same issue unfortunately, I've commented on your update

src/components/v2/Layout/Sidebar/index.tsx
baomastr: done the fix in other way.

therealemjy (approved):
Nice 👍

PR #680, opened 2022-06-21
[VEN-251] Vault - Claim reward

therealemjy:
## Jira ticket(s)

VEN-251

## Changes

- create `useClaimVaultReward` hook
- wire up claim reward button of vaults

### DIFF
--- a/src/hooks/useStakeWeiInVault.spec.tsx
+++ b/src/hooks/useStakeWeiInVault.spec.tsx
@@ -4,7 +4,7 @@ import { waitFor, fireEvent } from '@testing-library/react';
 
 import { TokenId } from 'types';
 import { TOKENS } from 'constants/tokens';
-import { stakeWeiInXvsVault, stakeWeiInVaiVault } from 'clients/api';
+import { stakeWeiInXvsVault, stakeWeiInVaiVault, stakeWeiInVrtVault } from 'clients/api';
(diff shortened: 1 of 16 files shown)


src/hooks/useStakeWeiInVault.spec.tsx
```diff
@@ -92,5 +92,39 @@ describe('hooks/useStakeWeiInVault', () => {
...
   });
 
-  // TODO: add tests for stakeWeiInVaiVault and stakeWeiInVrtVault
+  it('calls stakeWeiInVrtVault with correct parameters when calling stake without a poolIndex and stakedTokenId is equal to "vrt"', async () => {
+    const TestComponent: React.FC = () => {
+      const { stake } = useStakeWeiInVault({
+        stakedTokenId: TOKENS.vrt.id as TokenId,
+      });
+
+      return (
```
coreyar, line 101: What is the advantage of rendering the button and clicking it over rendering a null component and calling the hook in the render method of the test?

coreyar (approved):

src/hooks/useStakeWeiInVault.spec.tsx
therealemjy: None I'd say, for me both approaches are valid. I personally prefer this one because it feels closer to the actual implementation, but down the line I think the result is the same in both scenarios.

PR #681, opened 2022-06-21
[VEN-226] Add request to get proposal detail

coreyar:
## Jira ticket(s)

[VEN-226]

## Changes
Adds request to fetch individual proposal


src/clients/api/queries/useGetProposal.ts
```diff
@@ -0,0 +1,19 @@
...
+type Options = QueryObserverOptions<
+  GetProposalOutput,
+  Error,
+  GetProposalOutput,
+  GetProposalOutput,
+  [FunctionKey.GET_PROPOSAL, IGetProposalInput]
+>;
+
+const useGetProposal = (params: IGetProposalInput, options?: Options) =>
+  useQuery([FunctionKey.GET_PROPOSAL, params], () => getProposal(params), {
```
therealemjy, line 15: I've been spreading the params in function keys, but I actually think your approach is better becomes it makes it easy to know what `params` should look like. Nice 👍

therealemjy (approved):

PR #683, opened 2022-06-21
[VEN-193] Fetch votes by type

coreyar:
## Jira ticket(s)

[VEN-226]

## Changes

Adds fetching votes for proposal detail


src/__mocks__/models/vote.ts
```diff
@@ -0,0 +1,7 @@
+import formatToVoter from 'clients/api/queries/getVoters/formatToVoter';
+import { IGetVotersApiResponse } from 'clients/api/queries/getVoters';
+import voteResponse from '../api/vote.json';
+
+const votes = formatToVoter(voteResponse.data as IGetVotersApiResponse);
```
therealemjy, line 5: This approach is a bit different from what we've been doing so far, since we were mapping through each item and applying the formatter on each of them rather than passing the whole response to it.
I'm cool with either approach, but we should be consistent

therealemjy (approved):

src/__mocks__/models/vote.ts
coreyar: To be honest this endpoint is a bit unique. It paginates but also returns a vote summary

therealemjy (approved):

PR #684, opened 2022-06-21
[VEN-188] Add proposal requests

coreyar:
## Jira ticket(s)

VEN-188

## Changes

Wires up requests for creating, canceling, executing and queuing a proposal


src/pages/Vote/CreateProposalModal/formatProposalPayload.ts
```diff
@@ -0,0 +1,84 @@
...
+      .map(v => v.trim());
+  }
+  return value;
+};
+
+const formatProposalPayload = (data: FormValues) => {
+  const payload = {
+    targets: [] as string[],
+    signatures: [] as string[],
+    callDatas: [] as string[],
```
therealemjy: I'd type `payload` directly instead

src/pages/Vote/CreateProposalModal/formatProposalPayload.ts
```diff
@@ -0,0 +1,84 @@
...
+      const processedCallData = action.callData.reduce((acc, curr) => {
+        if (curr !== undefined) {
+          acc.push(formatIfArray(curr));
+        }
+        return acc;
+      }, [] as (string | number | string[])[]);
+      const callDataTypes = parseFunctionSignature(action.signature)?.inputs.map(
+        input => input.type,
+      );
+      console.log(callDataTypes, processedCallData);
```
therealemjy: Needs to be removed

src/pages/Vote/Governance/index.tsx
```diff
@@ -107,7 +110,10 @@ const Governance: React.FC = () => {
       total={total}
       limit={limit}
       setCurrentPage={setCurrentPage}
-      createProposal={createProposal}
+      createProposal={payload =>
+        createProposal({ ...payload, accountAddress: account?.address || '' })
```
therealemjy: Say a developer refactors this code and calls `createProposal` while the account address is undefined, would the contract return an error or would the proposal still be created with an empty address?
I expect the latter, in which case we don't need to make any modification. Otherwise we'd need to add a safeguard

coreyar: This shouldn't be available because the button is hidden until a user connects a wallet with enough XVS. I can double check the call without an address though

therealemjy: Yeah I know right now that's fine, hence why I mentioned that if someone was to refactor this code and call the method without checking the user address exists, then it could create issues.

But if the contract returns an error anyway if we pass an empty address (which I highly suspect it does), then we're good. I'm almost certain that's the case since addresses are typed as such in Solidity and so an empty string shouldn't be accepted

therealemjy (approved):

PR #689, opened 2022-06-23
[VEN-248] Add mutation functions to withdraw from vaults

therealemjy:
## Jira ticket(s)

VEN-248

## Changes

- add mutation functions to withdraw from vaults
- remove "wei" from staking mutations' names


src/pages/Vault/modals/StakeModal/index.spec.tsx
```diff
@@ -8,12 +8,12 @@ import MAX_UINT256 from 'constants/maxUint256';
...
 import { getBalanceOf, getAllowance } from 'clients/api';
 import renderComponent from 'testUtils/renderComponent';
-import useStakeWeiInVault from 'hooks/useStakeWeiInVault';
+import useStakeInVault from 'hooks/useStakeInVault';
 import en from 'translation/translations/en.json';
 import StakeModal, { IStakeModalProps } from '.';
 
 jest.mock('clients/api');
-jest.mock('hooks/useStakeWeiInVault');
+jest.mock('hooks/useStakeInVault');
```
coreyar, line 16: These don't work as I expect them. when declared like this. If we set them in in a `beforeAll` hook we can declare default return values in the mocked file. 

Then we can manually override the mock for specific tests.

coreyar (approved):

src/pages/Vault/modals/StakeModal/index.spec.tsx
therealemjy: `jest.mock` calls get hoisted at the top of the file, so putting these lines in a `beforeAll` block would produce the same result as right now.
We also do currently declare default return values in the mocked file (`undefined`/void in this instance), and we override these defaults for specific tests (in this particular test it isn't necessary however).

PR #695, opened 2022-06-27
[VEN-345] Add voter leaderboard page

coreyar:
## Jira ticket(s)

VEN-345

## Changes

Add Voter Leaderboard page


therealemjy:
It looks like this PR needs to be rebased with `develop`

src/pages/Proposal/ProposalSummary/Stepper/index.stories.tsx
```diff
@@ -44,6 +44,9 @@ Defeated.args = {
   endDate: new Date(1654000009000),
   queuedDate: undefined,
 };
+// .args = {
+//   proposal: { ...proposals[0], state: 'Defeated', executedDate: undefined, queuedDate: undefined },
+// };
```
therealemjy: Let's remove that

src/pages/Proposal/ProposalSummary/Stepper/index.stories.tsx
```diff
@@ -55,6 +58,9 @@ Succeeded.args = {
   endDate: new Date(1654000009000),
   queuedDate: undefined,
 };
+// .args = {
+//   proposal: { ...proposals[0], state: 'Succeeded', executedDate: undefined, queuedDate: undefined },
+// };
```
therealemjy: Let's remove that

src/pages/Proposal/ProposalSummary/Stepper/index.stories.tsx
```diff
@@ -66,6 +72,15 @@ Canceled.args = {
...
 };
+// .args = {
+//   proposal: {
+//     ...proposals[0],
+//     state: 'Canceled',
+//     cancelDate: new Date(1654005000700),
+//     executedDate: undefined,
+//     queuedDate: undefined,
+//   },
+// };
```
therealemjy: Let's remove that

src/pages/Proposal/ProposalSummary/Stepper/index.stories.tsx
```diff
@@ -88,6 +106,9 @@ Queued.args = {
   endDate: new Date(1654000009000),
   queuedDate: undefined,
 };
+// .args = {
+//   proposal: { ...proposals[0], state: 'Queued', executedDate: undefined },
+// };
```
therealemjy: Let's remove that

src/pages/Proposal/ProposalSummary/Stepper/index.stories.tsx
```diff
@@ -77,6 +92,9 @@ Expired.args = {
   endDate: new Date(1654000009000),
   queuedDate: undefined,
 };
+// .args = {
+//   proposal: { ...proposals[0], state: 'Expired', executedDate: undefined, queuedDate: undefined },
+// };
```
therealemjy: Let's remove that

src/pages/Proposal/ProposalSummary/Stepper/index.stories.tsx
```diff
@@ -99,3 +120,6 @@ Executed.args = {
   endDate: new Date(1654000009000),
   queuedDate: undefined,
 };
+// .args = {
+//   proposal: { ...proposals[0], state: 'Executed' },
+// };
```
therealemjy: Let's remove that

src/pages/VoterLeaderboard/LeaderboardTable/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+    ],
+    [],
+  );
+
+  const cardColumns = useMemo(() => {
+    const newColumns = [...columns];
+    newColumns[2].align = 'center';
+    newColumns[3].align = 'left';
+    return newColumns;
+  }, [columns]);
```
therealemjy: These aren't based on any state or props, so they can be declared outside the component

src/pages/VoterLeaderboard/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { useGetVoterAccounts } from 'clients/api';
+import { VoterAccount } from 'types';
+import LeaderboardTable from './LeaderboardTable';
+import { useStyles } from './styles';
+
+interface IVoterLeaderboardProps {
+  voterAccounts: VoterAccount[];
+  offset: number;
+  total: number | undefined;
+  limit: number | undefined;
```
therealemjy, line 13: nit: I think making these optional would fit the spec better, since we're basically saying a `total` and a `limit` don't have to be passed rather than saying they have to be passed but can be `undefined`.

therealemjy (commented):
Small alignment issue on the headers:
Screen Shot 2022-06-28 at 11 12 40

src/pages/VoterLeaderboard/LeaderboardTable/index.tsx
coreyar: They rely on the translation context though. We can still define them outside the context and access the translation function from the instance but I think doing that should be an exemption when we can't set the translation in the render. What do you think?

src/pages/VoterLeaderboard/index.tsx
coreyar: They do need to be passed though. They are required for this component in order for pagination to work correctly. What I am trying to convey here is that they are required by could be undefined until the query loads.

src/pages/VoterLeaderboard/LeaderboardTable/index.tsx
therealemjy: Ah you're right yeah, let's keep these there then

src/pages/VoterLeaderboard/index.tsx
therealemjy: Ok I see. Works for me then!

therealemjy (approved):

PR #696, opened 2022-06-28
[VEN-346] Add voter detail page

coreyar:
## Jira ticket(s)

[VEN-346]

## Changes

Add Voter Detail page


src/types/index.ts
```diff
@@ -281,3 +281,69 @@ export interface Vault {
...
+  updatedAt: Date;
+  support: VoteSupport;
+};
+
+export type VoteDetailTransaction = VoteDetailTransactionTransfer | VoteDetailTransactionVote;
+
+export interface IVoterDetail {
+  balanceWei: BigNumber;
+  delegateCount: number;
+  delegates: string;
```
therealemjy: I think we should make this prop name more explicit: `delegateeAddress`

src/translation/translations/en.json
```diff
@@ -607,7 +607,18 @@
     "operation": "Operation",
     "proposalHistory": "Proposal History",
     "proposalState": {
-      "active": "Active"
+      "active": "Active",
+      "canceled": "Canceled",
+      "defeated": "Defeated",
+      "executed": "Executed",
+      "queued": "Queued",
+      "succeeded": "Suceeded",
```
therealemjy:
```suggestion
      "succeeded": "Succeeded",
```

src/translation/translations/en.json
```diff
@@ -607,7 +607,18 @@
...
+      "activeTimestamp": "Active: <Span>{{date, dd MMM yyyy HH:mm}}</Span>",
+      "canceledTimestamp": "Canceled: <Span>{{date, dd MMM yyyy HH:mm}}</Span>",
+      "defeatedTimestamp": "Defeated: <Span>{{date, dd MMM yyyy HH:mm}}</Span>",
+      "executedTimestamp": "Executed: <Span>{{date, dd MMM yyyy HH:mm}}</Span>",
+      "queuedTimestamp": "Queued: <Span>{{date, dd MMM yyyy HH:mm}}</Span>",
+      "succeededTimestamp": "Suceeded: <Span>{{date, dd MMM yyyy HH:mm}}</Span>"
```
therealemjy:
```suggestion
      "succeededTimestamp": "Succeeded: <Span>{{date, dd MMM yyyy HH:mm}}</Span>"
```

src/pages/VoterDetail/History/VoterProposal/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+              }}
+              values={{
+                date: cancelDate,
+              }}
+            />
+          ),
+        ];
+      case 'Succeeded':
+        return [
+          <ActiveChip text={t('voteProposalUi.proposalState.passed')} />,
```
therealemjy: Judging from the translation file, you've changed this translation:
```suggestion
          <ActiveChip text={t('voteProposalUi.proposalState.succeeded')} />,
```

I also suggest running `extract-translation-keys` (I ran it on the PR and the order of translations changes).

src/clients/api/queries/getVoterDetail/formatVoterDetailResponse.ts
```diff
@@ -0,0 +1,63 @@
...
+          blockTimestamp: new Date(blockTimestamp * 1000),
+          createdAt: new Date(createdAt),
+          from,
+          to,
+          transactionHash,
+          transactionIndex,
+          type,
+          updatedAt: new Date(updatedAt),
+          votesWei: new BigNumber(votes),
+        } as VoteDetailTransactionTransfer;
```
therealemjy: You don't need to duplicate the whole object:
```suggestion
       const voteBase = {
          blockNumber,
          blockTimestamp: new Date(blockTimestamp * 1000),
          createdAt: new Date(createdAt),
          from,
          to,
          transactionHash,
          transactionIndex,
          updatedAt: new Date(updatedAt),
          votesWei: new BigNumber(votes),
        };

        if (type === 'vote' && 'support' in rest) {
          const transactionVote: VoteDetailTransactionVote = {
            ...voteBase,
            type: 'vote',
            support: getSupportName(rest.support),
          };

          return transactionVote;
        }

        const transactionTransfer: VoteDetailTransactionTransfer = {
          ...voteBase,
          type: 'transfer',
          amountWei: new BigNumber(rest.amount),
        };

        return transactionTransfer;
```

src/pages/VoterDetail/Holding/styles.ts
```diff
@@ -0,0 +1,25 @@
...
+  return {
+    root: css`
+      flex-direction: column;
+    `,
+    title: css`
+      margin-bottom: ${theme.spacing(8)};
+    `,
+    value: css`
+      margin-top: ${theme.spacing(1)};
+      line-height: 1.25rem;
```
therealemjy: This needs to be removed

src/pages/VoterDetail/History/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+              queuedDate={queuedDate}
+              cancelDate={cancelDate}
+              executedDate={executedDate}
+              endDate={endDate}
+            />
+          ),
+        )
+      ) : (
+        <Spinner />
+      )}
```
therealemjy: I think we should always display the list of `voterHistory` isn't empty and display the loader when `isFetching` is truthy, because currently the page blinks every time it loads new content.

This would also make the UX consistent with the other pages (e.g.: History).

src/pages/VoterDetail/History/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <Pagination
+          css={styles.pagination}
+          itemsCount={total}
+          onChange={(nextIndex: number) => {
+            setCurrentPage(nextIndex);
+            window.scrollTo(0, 0);
+          }}
+          itemsPerPageCount={limit}
+        />
+      ) : null}
```
therealemjy:
```suggestion
      {total > 0 && (
        <Pagination
          css={styles.pagination}
          itemsCount={total}
          onChange={(nextIndex: number) => {
            setCurrentPage(nextIndex);
            window.scrollTo(0, 0);
          }}
          itemsPerPageCount={limit}
        />
      )}
```

src/pages/VoterDetail/Holding/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+    if (!votesWei) {
+      return PLACEHOLDER_KEY;
+    }
+    return convertWeiToTokens({
+      valueWei: votesWei,
+      tokenId: 'xvs',
+      addSymbol: false,
+      returnInReadableFormat: true,
+    });
+  }, [votesWei]);
```
therealemjy: You can use `useConvertWeiToReadableTokenString` for both of these. I'd also suggest grabbing the id of XVS from the `TOKENS` constant (`import { TOKENS } from 'constants/tokens';`)

src/pages/VoterDetail/Holding/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+              {delegateCount?.toString() || PLACEHOLDER_KEY}
+            </Typography>
+          </>
+        }
+        value={100}
+        step={1}
+        min={0}
+        max={100}
+        ariaLabel={t('voterDetail.holdingProgressBar')}
+      />
```
therealemjy: I've added a few comments regarding the final UI of this DOM not quite corresponding to the designs; I'd suggest using the `ProgressBar` component directly here, so that you can set the correct background color for the progress bar (blue instead of green) and use `Typography` with the variant `h4` to get the right font size.

src/pages/VoterDetail/Transactions/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const styles = useStyles();
+  const { t } = useTranslation();
+  const columns = useMemo(
+    () => [
+      { key: 'action', label: t('voterDetail.actions'), orderable: false, align: 'left' },
+      { key: 'age', label: t('voterDetail.age'), orderable: false, align: 'left' },
+      { key: 'amount', label: t('voterDetail.amount'), orderable: false, align: 'right' },
+    ],
+    [],
+  );
```
therealemjy: This can be declared outside the component

src/pages/VoterDetail/Transactions/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  const rows: TableProps['data'] = voterTransactions.map(voterTxs => {
```
therealemjy: Let's memoize this variable

src/pages/VoterDetail/Transactions/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+            {action}
+          </Typography>
+        ),
+        value: uid(action),
+        align: 'left',
+      },
+      {
+        key: 'age',
+        render: () => t('voterDetail.readableAge', { age: voterTxs.blockTimestamp }),
+        value: uid(voterTxs.blockTimestamp),
```
therealemjy: Same thing here, it feels like uid isn't necessary

src/pages/VoterDetail/Transactions/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+      {
+        key: 'amount',
+        render: () =>
+          convertWeiToTokens({
+            valueWei: amountWei,
+            tokenId: 'xvs',
+            minimizeDecimals: true,
+            returnInReadableFormat: true,
+          }),
+        value: uid(amountWei.toFixed()),
```
therealemjy: Same thing here, it feels like `uid` isn't necessary

src/pages/VoterDetail/Transactions/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    return [
+      {
+        key: 'action',
+        render: () => (
+          <Typography css={styles.action} variant="small2" color="textPrimary">
+            {action}
+          </Typography>
+        ),
+        value: uid(action),
```
therealemjy: Why do you need `uid` here? `value` doesn't need to be unique it seems

src/pages/VoterDetail/History/VoterProposal/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+            abstainedWei={abstainedVotesWei}
+            votedTotalWei={votedTotalWei}
+          />
+        </Grid>
+      </Grid>
+    </Paper>
+  );
+};
+
+export default GovernanceProposal;
```
therealemjy: This component is a duplicate of `GovernanceProposal`. Since the UI of these two components is similar with a few modifications we should create a reusable component to share between the Vote page and the Voter Details page.

therealemjy (commented):
- To be consistent with the codebase, I think `detail` should be pluralized everywhere it is used here
- The address of the voter needs to be added to the header like in the designs:
Screen Shot 2022-06-28 at 12 05 16
- The "Transaction" section has some responsive issues on tablet:
Screen Shot 2022-06-28 at 11 29 50
- The progress bar in the "Transaction" section should be blue according to the designs
- The arrow icon and the text content of each transaction line aren't vertically aligned:
Screen Shot 2022-06-28 at 11 53 24
- The font size used for the voting weight is smaller than in the designs:
Screen Shot 2022-06-28 at 11 49 30
- The user icon and the number on its right, on top of the progress bar, aren't vertically aligned:
Screen Shot 2022-06-28 at 11 49 17
- The date in the footer of each voting action is sometimes on two lines, sometimes on just one. According to the designs, it should always be on one line only:
Screen Shot 2022-06-28 at 11 55 45
Screen Shot 2022-06-28 at 11 55 46
- The spacing between the content and the edge of each voting history item is too big (32px instead of 24px):
Screen Shot 2022-06-28 at 11 59 07

src/clients/api/queries/getVoterDetail/formatVoterDetailResponse.ts
coreyar: Nice, I tried other ways that didn't work but this does!

src/pages/VoterDetail/History/index.tsx
coreyar: This is actually handled internally by Pagination, the check exists because `itemsCount` can't be undefined

src/pages/VoterDetail/Holding/index.tsx
coreyar: In the end this was removed

src/pages/VoterDetail/Transactions/index.tsx
coreyar: I didn't declare this outside the component because of the translations

src/pages/VoterDetail/Transactions/index.tsx
coreyar: This is the one where the uid is necessary because it is getting used as a key and multiple transactions can be in the same block timestamp

coreyar: There is a better way to solve that though

therealemjy: Even using `uid` I should still end up with duplicates if two items are given the exact same value.

Perhaps you could append another prop to it?

therealemjy:
Nice, looks very good 👍

What do you think about adding an hover effect on the "copy" icon next to the address in the header and displaying a toast when clicking on it to indicate it's been copied to the clipboard?
Similarly to how we do with the same icon in the "Your wallet" modal:
Screen Shot 2022-06-29 at 10 25 37

Apart from this it looks good to merge for me.

coreyar:
> Nice, looks very good 👍
> 
> What do you think about adding an hover effect on the "copy" icon next to the address in the header and displaying a toast when clicking on it to indicate it's been copied to the clipboard? Similarly to how we do with the same icon in the "Your wallet" modal: Screen Shot 2022-06-29 at 10 25 37
> 
> Apart from this it looks good to merge for me.

yes, that was the idea, just waiting on the other PR to merge

therealemjy:
> > Nice, looks very good 👍
> > What do you think about adding an hover effect on the "copy" icon next to the address in the header and displaying a toast when clicking on it to indicate it's been copied to the clipboard? Similarly to how we do with the same icon in the "Your wallet" modal: Screen Shot 2022-06-29 at 10 25 37
> > Apart from this it looks good to merge for me.
> 
> yes, that was the idea, just waiting on the other PR to merge

Which other PR?

I had another look and spotted an issue with pagination. While I'm on page 1, if I click on the page 2 button the page does refresh but the pagination still displays page 1 as being the active one. If I click on page 2 again, the content stays the same but the page 2 is set as active. If I now click on page 3, it refreshed the content and displays page 1 as being active.

I also found two little design issues
- user count: the icon and the text aren't vertically aligned and the text should also be grey:
Screen Shot 2022-06-30 at 08 17 51
- header address: only the icon should be clickable, not the whole block, like in the "Your wallet" modal. The address should also be aligned with the edge of the page content:
Screen Shot 2022-06-30 at 08 20 14

therealemjy:
Nice 👍

therealemjy (approved):

PR #704, opened 2022-06-28
Remove check if form is dirty

coreyar:
## Changes

This change fixes an issue with the form in the vote modal where it is not initially valid. 

This form should be initially valid and clean. Other forms should be initially invalid and the validation schema should run and set it valid if the criteria are met. If we need this check there may be another issue going on. I check other forms on the market page and this change rendered those forms initially disabled


therealemjy (approved):
Looks good, we'll just need to merge https://github.com/VenusProtocol/venus-protocol-interface/pull/704 before we merge this in

PR #710, opened 2022-06-30
[VEN-320] Remove comma number

coreyar:
## Jira ticket(s)

VEN-320 - Remove comma number formatting function
VEN-371 - Pin to react 17

## Changes

Removes comma-number
Upgrades to react 17
Adjusts types to work with upgraded versions


package.json
```diff
@@ -51,17 +49,17 @@
...
     "react-i18next": "^11.16.5",
     "react-query": "^3.34.16",
     "react-router": "^5.2.1",
     "react-router-dom": "^5.0.1",
     "react-scripts": "4.0.3",
-    "react-toastify": "^9.0.1",
+    "react-toastify": "9.0.1",
     "react-uid": "^2.3.1",
-    "recharts": "^2.1.9",
+    "recharts": "2.1.9",
```
therealemjy, line 62: Why do we need to fix the versions of these packages?

therealemjy (approved):

package.json
coreyar: Recharts was throwing an error relating to a resizing property missing from window. react-toastify had a esm compilation issue if I recall correctly

PR #716, opened 2022-07-01
[VEN-252] Vault - Final assembling

therealemjy:
…ing vault withdraw while disconnected

## Jira ticket(s)

VEN-252

## Changes

- use new Vault
- remove old Vault page components
- display error toast when clicking on withdraw button of VRT vesting vault while not being connected with any account


src/pages/Vault/VaultItem/index.tsx
```diff
@@ -286,17 +287,21 @@ const VaultItem: React.FC<VaultItemProps> = ({
 
   const onWithdraw = async () => {
     if (stakedTokenId !== TOKENS.vrt.id || typeof poolIndex === 'number') {
-      // Handle withdrawing from any vault except the VRT vault
+      // Handle withdrawing from any vault except the VRT vesting vault
```
coreyar, line 290: This Modal has a comment about checking that the account has to exist. We can remove that comment and check for the account before opening the modal like we do on line 296

therealemjy: There's two different cases here: in what case we open a modal, which then uses `ConnectWallet` to display a prompt to the user asking them to connect their wallet if they haven't done so yeah. On the other case, clicking on "Withdraw" directly sends transaction because users can only withdraw the totality of staked tokens (this only applies to the VRT vesting vault).

In the first case I've followed the same UX we follow for modal, which consists in displaying a prompt inside the modal if no wallet is connected. In the second case, the only other option I have is to display a toast error since we don't display any modal.

coreyar (approved):

PR #722, opened 2022-07-05
Fix display issues on Vote screen

therealemjy:
## Jira ticket(s)

N/A

## Changes

- always display "Create proposal" button but disable it if user can't create proposals (UX described in Figma doc)
- fix page loader size

### DIFF
--- a/src/clients/api/queries/useGetProposals.ts
+++ b/src/clients/api/queries/useGetProposals.ts
@@ -14,9 +14,11 @@ type Options = QueryObserverOptions<
 >;
 
 const useGetProposals = (params: IGetProposalsInput = {}, options?: Options) =>
-  // This endpoint is paginated so we keep the previous responses by default to create a more seamless paginating experience
+  // This endpoint is paginated so we keep the previous responses by default to
+  // create a more seamless paginating experience
   useQuery([FunctionKey.GET_PROPOSALS, params], () => getProposals(params), {
     keepPreviousData: true,
+    placeholderData: { limit: 0, total: 0, offset: 0, proposals: [] },
     refetchInterval: params.page === 0 ? 15000 : undefined,
     ...options,
   });
--- a/src/clients/api/queries/useGetTransactions.ts
+++ b/src/clients/api/queries/useGetTransactions.ts
@@ -15,9 +15,9 @@ type Options = QueryObserverOptions<
 
 const useGetTransactions = (params: IGetTransactionsInput, options?: Options) =>
   useQuery([FunctionKey.GET_TRANSACTIONS, params], () => getTransactions(params), {
-    ...options,
     keepPreviousData: true,
     placeholderData: { limit: 0, page: 0, total: 0, transactions: [] },
+    ...options,
   });
 
 export default useGetTransactions;
--- a/src/clients/api/queries/useGetVoters.ts
+++ b/src/clients/api/queries/useGetVoters.ts
@@ -10,11 +10,11 @@ type Options = QueryObserverOptions<
   [FunctionKey.GET_VOTERS, IGetVotersInput]
 >;
 
-const useGetProposals = (params: IGetVotersInput, options?: Options) =>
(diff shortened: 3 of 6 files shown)


src/clients/api/queries/useGetTransactions.ts
```diff
@@ -15,9 +15,9 @@ type Options = QueryObserverOptions<
 
 const useGetTransactions = (params: IGetTransactionsInput, options?: Options) =>
   useQuery([FunctionKey.GET_TRANSACTIONS, params], () => getTransactions(params), {
-    ...options,
```
coreyar, line 18: I did it this way because the previous options should not be overriden. For example, because this endpoint is paginated we should always `keepPreviousData`. It would not be correct to override it

src/clients/api/queries/useGetProposals.ts
```diff
@@ -14,9 +14,11 @@ type Options = QueryObserverOptions<
 >;
 
 const useGetProposals = (params: IGetProposalsInput = {}, options?: Options) =>
-  // This endpoint is paginated so we keep the previous responses by default to create a more seamless paginating experience
+  // This endpoint is paginated so we keep the previous responses by default to
+  // create a more seamless paginating experience
   useQuery([FunctionKey.GET_PROPOSALS, params], () => getProposals(params), {
     keepPreviousData: true,
+    placeholderData: { limit: 0, total: 0, offset: 0, proposals: [] },
```
coreyar, line 21: Similar to the previous comment, I think options is better set above these values because they shouldn't be overriden

src/clients/api/queries/useGetVoters.ts
```diff
@@ -10,11 +10,11 @@ type Options = QueryObserverOptions<
   [FunctionKey.GET_VOTERS, IGetVotersInput]
 >;
 
-const useGetProposals = (params: IGetVotersInput, options?: Options) =>
+const useGetVoters = (params: IGetVotersInput, options?: Options) =>
   // This endpoint is paginated so we keep the previous responses by default to create a more seamless paginating experience
   useQuery([FunctionKey.GET_VOTERS, params], () => getVoters(params), {
     keepPreviousData: true,
```
coreyar, line 16: same here where keepPreviousData shouldn't be overridden

src/pages/Vote/Governance/index.tsx
```diff
@@ -85,7 +91,8 @@ export const GovernanceUi: React.FC<IGovernanceUiProps> = ({
           ),
         )}
       </div>
-      {total && (
+
+      {total && total > 0 && (
```
coreyar, line 95: This is handled by the Pagination component on line 56

src/clients/api/queries/useGetTransactions.ts
therealemjy: This is inconsistent with other queries that also use pagination and `keepPreviousData`.
Being able to override this option is the right approach in my opinion, since there could be instances where we only one to load one page at a time and not keep the previous data.
Also, allowing options to be passed but not overriding the default ones will create issues for devs who won't understand why the options they pass to the queries aren't taken in consideration.

src/pages/Vote/Governance/index.tsx
therealemjy: We need the additional check because of the placeholder data I added. I added them to be consistent with how we use pagination on other pages

src/clients/api/queries/useGetTransactions.ts
coreyar: Yeah I commented on the others as well. The comment was meant to explain why it was setup that way to make it easier to understand, since the guarding the implementation is important. Even loading one page it would make sense to keep the previous data

coreyar (approved):

src/clients/api/queries/useGetTransactions.ts
therealemjy: Not in the case I described where you only load one page and display that one page, plus the rest of my comment (consistency, giving an easy to understand experience to devs)

PR #727, opened 2022-07-05
[VEN-377] Fix issues on Proposal Details page

therealemjy:
## Jira ticket(s)

VEN-377

## Changes

- replace `EllipsisText` component with `EllipsisAddress`, which uses a CSS breakpoint to determine when to display truncated address
- fix alignment and responsive issues
- fix voting buttons which were always disabled
- invalidate queries after voting
- fix countdown and timestamp in footer of vote proposals


src/clients/api/mutations/vote/useCastVote.ts
```diff
@@ -18,7 +19,28 @@ const useCastVote = (
...
+        queryClient.invalidateQueries([
+          FunctionKey.GET_VOTERS,
+          {
+            id: proposalId,
+            filter: voteType,
+          },
+        ]);
+
+        // Invalidate query to fetch proposal list
+        queryClient.invalidateQueries([FunctionKey.GET_PROPOSALS]);
```
coreyar, line 37: This is not necessary because this query polls while the proposal is active.

src/clients/api/mutations/vote/useCastVote.ts
```diff
@@ -18,7 +19,28 @@ const useCastVote = (
...
         ...params,
       }),
-    options,
+    {
+      ...options,
+      onSuccess: (...onSuccessParams) => {
+        const { proposalId, voteType } = onSuccessParams[1];
+
+        // Invalidate query to fetch voters
+        queryClient.invalidateQueries([
```
coreyar, line 28: We likely want to invalidate this endpoint the same way so that is isn't only updated when the user votes but updated with ever block so all votes are updated.

src/clients/api/mutations/vote/useCastVoteWithReason.ts
```diff
@@ -22,7 +23,28 @@ const useCastVoteWithReason = (
         fromAccountAddress,
         ...params,
       }),
-    options,
+    {
+      ...options,
+      onSuccess: (...onSuccessParams) => {
+        const { proposalId, voteType } = onSuccessParams[1];
+
+        // Invalidate query to fetch voters
```
coreyar, line 31: Same comment about invalidating these queries on an interval if the proposal is active

src/pages/Proposal/ProposalSummary/Stepper/index.tsx
```diff
@@ -159,8 +159,7 @@ const Stepper: React.FC<IStepperProps> = ({
               </div>
               {step.getTimestamp() && (
                 <Typography variant="tiny" css={styles.dateDefault}>
-                  {t('voteProposalUi.statusCard.dateOnly', { date: step.getTimestamp() })}
-
+                  {t('voteProposalUi.statusCard.dateOnly', { date: step.getTimestamp() })}{' '}
```
coreyar: If we have to hardcode a space I think we should use a nonbreaking space character. `&nbsp;`

src/clients/api/mutations/vote/useCastVote.ts
therealemjy: Although we automatically refresh the query on an interval, it is unlikely to be synced exactly with when a new block is mined (since the polling interval is just set as soon as the query mounts), meaning users could have to wait up to a full interval (15 seconds currently) before seeing the UI update

src/clients/api/mutations/vote/useCastVote.ts
therealemjy: We should update the query in both cases: when a user casts a vote so their UI updates in real time and on an interval to keep the front end updated.
That second case will be covered by this ticket: https://jira.toolsfdg.net/browse/VEN-298

coreyar (approved):

PR #728, opened 2022-07-06
Add vote tests

coreyar:
## Changes

Add tests for the vote page.
Tests
- interacting with links on the page
- Interacting with the delegation modal

Adds a new feature where the modal still opens but prompts the user to connect a wallet if it isn't connected.

### DIFF
--- a/src/pages/Vote/DelegateModal/index.tsx
+++ b/src/pages/Vote/DelegateModal/index.tsx
@@ -92,7 +92,9 @@ const DelegateModal: React.FC<IDelegateModalProps> = ({
               {currentUserAccountAddress ? (
                 <FormikSubmitButton
                   fullWidth
-                  enabledLabel={previouslyDelegated ? t('vote.redelegate') : t('vote.delgateVotes')}
+                  enabledLabel={
+                    previouslyDelegated ? t('vote.redelegate') : t('vote.delegateVotes')
+                  }
                   css={styles.submitButton}
                   loading={isVoteDelegationLoading}
                 />
--- a/src/pages/Vote/DelegateModal/index.tsx
+++ b/src/pages/Vote/DelegateModal/index.tsx
@@ -9,6 +9,7 @@ import {
   NoticeInfo,
   FormikSubmitButton,
   FormikTextField,
+  PrimaryButton,
   TextButton,
 } from 'components';
 import Path from 'constants/path';
@@ -23,6 +24,7 @@ interface IDelegateModalProps {
   setVoteDelegation: (address: string) => void;
   previouslyDelegated: boolean;
   isVoteDelegationLoading: boolean;
+  openAuthModal: () => void;
 }
 
 const DelegateModal: React.FC<IDelegateModalProps> = ({
@@ -32,6 +34,7 @@ const DelegateModal: React.FC<IDelegateModalProps> = ({
   setVoteDelegation,
   previouslyDelegated,
   isVoteDelegationLoading,
+  openAuthModal,
 }) => {
   const { t } = useTranslation();
   const styles = useStyles();
@@ -75,6 +78,7 @@ const DelegateModal: React.FC<IDelegateModalProps> = ({
                 <TextButton
                   css={styles.inline}
                   onClick={() => setFieldValue('address', currentUserAccountAddress)}
+                  disabled={!currentUserAccountAddress}
                 >
                   {t('vote.pasteYourAddress')}
                 </TextButton>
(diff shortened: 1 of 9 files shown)


src/pages/Vote/testIds.ts
```diff
@@ -0,0 +1,10 @@
+const TEST_IDS = {
+  votingWallet: {
+    votingWeightValue: 'vote-voting-wallet-voting-weight-value',
+    totalLockedValue: 'vote-voting-wallet-total-locked-value',
+  },
+  governance: {
+    governanceProposal: (id: string) => `vote-governance-governance-proposal-${id}`,
+  },
+};
+export default TEST_IDS;
```
therealemjy: We now have a file containing all the test IDs used in the app: `constants/tedIds`, these should be added to it

src/pages/Vote/index.spec.tsx
```diff
@@ -1,19 +1,222 @@
...
+  it('successfully delegates to me', async () => {
+    const { openSuccessfulTransactionModal } = useSuccessfulTransactionModal();
+    const { getByText } = renderComponent(<Vote />, {
+      authContextValue: {
+        account: {
+          address: fakeAccountAddress,
+        },
+      },
+    });
+    const createProposalButton = getByText('Delegate your voting power');
```
therealemjy: Let 's get this from the translation file

src/pages/Vote/index.spec.tsx
```diff
@@ -1,19 +1,222 @@
...
+    });
+
+    waitFor(() =>
+      expect(getByTestId(TEST_IDS.votingWallet.votingWeightValue)).toHaveTextContent('50'),
+    );
+    waitFor(() =>
+      expect(getByTestId(TEST_IDS.votingWallet.totalLockedValue)).toHaveTextContent('10'),
+    );
+
+    const delegateVotingPowerTextButton = getByText('Delegate your voting power');
```
therealemjy: Let's get this from the translation file

src/pages/Vote/index.spec.tsx
```diff
@@ -1,19 +1,222 @@
...
+      fireEvent.click(delgateVoteText);
+    });
+    waitFor(() => getByText(en.vote.delegateAddress));
+
+    expect(getByText(en.vote.delgateVotes));
+  });
+
+  it('can navigate to vault when clicking deposit tokens', async () => {
+    const { getByText } = renderComponent(<Vote />);
+    const deposityYourTokensText = getByText('Deposit your tokens');
```
therealemjy: Let's get this from the translation file

src/pages/Vote/index.spec.tsx
```diff
@@ -1,19 +1,222 @@
...
+    act(() => {
+      fireEvent.click(createProposalButton);
+    });
+
+    waitFor(() => getByText(en.vote.pages.proposalInformation));
+  });
+
+  it('opens delegate modal when clicking text with connect wallet button when unauthenticated', async () => {
+    const { getByText, getAllByText } = renderComponent(<Vote />);
+    const delgateVoteText = getByText('Delegate your voting power');
```
therealemjy: Let's get this from the translation file

src/pages/Vote/DelegateModal/index.tsx
```diff
@@ -83,13 +87,20 @@ const DelegateModal: React.FC<IDelegateModalProps> = ({
...
+                  fullWidth
+                  enabledLabel={previouslyDelegated ? t('vote.redelegate') : t('vote.delgateVotes')}
+                  css={styles.submitButton}
+                  loading={isVoteDelegationLoading}
+                />
+              ) : (
+                <PrimaryButton onClick={openAuthModal} css={styles.submitButton} fullWidth>
+                  {t('connectWallet.connectButton')}
+                </PrimaryButton>
+              )}
```
therealemjy: Nice

src/pages/Vote/DelegateModal/index.tsx
```diff
@@ -83,13 +87,20 @@ const DelegateModal: React.FC<IDelegateModalProps> = ({
...
-              <FormikSubmitButton
-                fullWidth
-                enabledLabel={previouslyDelegated ? t('vote.redelegate') : t('vote.delgateVotes')}
-                css={styles.submitButton}
-                loading={isVoteDelegationLoading}
-              />
+              {currentUserAccountAddress ? (
+                <FormikSubmitButton
+                  fullWidth
+                  enabledLabel={previouslyDelegated ? t('vote.redelegate') : t('vote.delgateVotes')}
```
therealemjy:
```suggestion
                  enabledLabel={previouslyDelegated ? t('vote.redelegate') : t('vote.delegateVotes')}
```

therealemjy (commented):
Great, that makes the code more robust.

It would be good to add tests to the Proposal Details page too, especially to check the different call to actions: cancelling a proposal, queuing it, executing it, voting for, voting against etc...

src/pages/Vote/index.spec.tsx
coreyar: The text in the translation file doesn't work because it includes markup

src/pages/Vote/testIds.ts
coreyar: This would be more correct considering the usage and naming is localized to the page per the pattern of keeping variables not used globally with the component

Other examples of this includes, hooks, constants and utility functions

coreyar:
> Great, that makes the code more robust.
> 
> 
> 
> It would be good to add tests to the Proposal Details page too, especially to check the different call to actions: cancelling a proposal, queuing it, executing it, voting for, voting against etc...

Absolutely, that is out of scope of this PR but I have started that work in another local branch

src/pages/Vote/testIds.ts
therealemjy: This is used globally, since multiple tests from different pages and component will potentially use this file.

Furthermore we need to make sure we're consistent with that. If we change the approach here, then we have to change it for all the components and so we should split the `testIds` file.

src/pages/Vote/index.spec.tsx
therealemjy: I'd suggest using a test ID then

therealemjy (approved):

PR #733, opened 2022-07-06
[VEN-383] Remove VaiContext

therealemjy:
## Jira ticket(s)

VEN-383

## Changes

- remove `VaiContext`
- remove `useVaiUser` hook in favor of using query hooks


src/clients/api/queries/getMintableVai/index.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+        accountAddress: fakeAddress,
+      });
+
+      throw new Error('getMintableVai should have thrown an error but did not');
+    } catch (error) {
+      expect(error).toMatchInlineSnapshot('[Error: Fake error message]');
+    }
+  });
+
+  test('returns the user info in the correct format', async () => {
```
coreyar: this be updated to describe the test better

src/clients/api/queries/getMintableVai/types.ts
```diff
@@ -0,0 +1,12 @@
+import BigNumber from 'bignumber.js';
+
+import { VaiUnitroller } from 'types/contracts';
+
+export interface GetMintableVaiInput {
+  vaiControllerContract: VaiUnitroller;
+  accountAddress: string;
+}
+
+export type GetMintableVaiOutput = {
```
coreyar: nit: observing that previously we would just return the value without wrapping it in an object

src/clients/api/queries/getMintableVai/useGetMintableVai.ts
```diff
@@ -0,0 +1,30 @@
...
+type Options = QueryObserverOptions<
+  GetMintableVaiOutput,
+  Error,
+  GetMintableVaiOutput,
+  GetMintableVaiOutput,
+  [FunctionKey.GET_MINTABLE_VAI, Omit<GetMintableVaiInput, 'vaiControllerContract'>]
+>;
+
+const useGetMintableVai = (
+  params: Omit<GetMintableVaiInput, 'vaiControllerContract'>,
```
coreyar, line 18:
```suggestion
  params: GetMintableVaiInput,
```

src/clients/api/queries/getMintableVai/types.ts
coreyar: Also should be an interface

src/clients/api/queries/getMintableVai/types.ts
```diff
@@ -0,0 +1,12 @@
+import BigNumber from 'bignumber.js';
+
+import { VaiUnitroller } from 'types/contracts';
+
+export interface GetMintableVaiInput {
+  vaiControllerContract: VaiUnitroller;
+  accountAddress: string;
+}
```
coreyar, line 8:
```suggestion
export interface GetMintableVaiInput {
  accountAddress: string;
}
```

src/clients/api/queries/getMintableVai/index.ts
```diff
@@ -0,0 +1,15 @@
+import { GetMintableVaiInput, GetMintableVaiOutput } from './types';
+import formatToOutput from './formatToOutput';
+
+export * from './types';
+
+const getMintableVai = async ({
+  vaiControllerContract,
+  accountAddress,
+}: GetMintableVaiInput): Promise<GetMintableVaiOutput> => {
```
coreyar, line 9:
```suggestion
interface ContractInput {
   vaiControllerContract: VaiUnitroller
}

const getMintableVai = async ({
  vaiControllerContract,
  accountAddress,
}: GetMintableVaiInput & ContractInput): Promise<GetMintableVaiOutput> => {
```

src/clients/api/queries/useGetMintedVai.ts
```diff
@@ -12,7 +12,7 @@ type Options = QueryObserverOptions<
   [FunctionKey.GET_MINTED_VAI, string]
 >;
 
-const useGetMintedVai = (accountAddress: string, options?: Options) => {
+const useGetMintedVai = ({ accountAddress }: { accountAddress: string }, options?: Options) => {
```
coreyar:
```suggestion
const useGetMintedVai = ({ accountAddress }: GetMintableVaiInput, options?: Options) => {
```

coreyar (approved):
Looks good. Main comments are some thoughts on how to better structure query types and tests need to pass

src/clients/api/queries/getMintableVai/index.ts
therealemjy: The contract from the input should not be returned, but I think you meant to divide the input in two. I wouldn't do that, as then `GetMintableVaiInput` would be incorrect; it should contain all the input parameters of the function, including the contract itself.

src/clients/api/queries/getMintableVai/types.ts
therealemjy: The function needs the contract to be passed to it

src/clients/api/queries/getMintableVai/types.ts
therealemjy: We did and there's actually two issues with that:
- we can't know what unit the resulting amount is in (unless we'd rename the function, which we've done in a few instances but not most)
- if this query was to change and more parameters would be added, it would be simpler to extend the object returned than to transform the return type from a BigNumber to an object

Let me know your thoughts; I think it could be nice to implement in the codebase.

therealemjy: It does need to be an interface yeah

coreyar: > We did and there's actually two issues with that:
> 
> * we can't know what unit the resulting amount is in (unless we'd rename the function, which we've done in a few instances but not most)
> * if this query was to change and more parameters would be added, it would be simpler to extend the object returned than to transform the return type from a BigNumber to an object
> 
> Let me know your thoughts; I think it could be nice to implement in the codebase.

works for me. Sounds nicer

src/clients/api/queries/getMintableVai/index.ts
coreyar: yeah I miswrote it. Updated

src/clients/api/queries/getMintableVai/types.ts
therealemjy: Ok great, I'll create a ticket to update the other queries and mutations

src/clients/api/queries/getMintableVai/index.ts
therealemjy: Got you. I still wouldn't divide the input as it can lead to misunderstandings for future developers: if one imports `GetMintableVaiInput`, you'd expect it contains all the parameters required by `getMintableVai`.

We can use `Omit` to remove the contract out of the type when needed (which is the current solution we've been following so far).

PR #734, opened 2022-07-06
Add Tests to Proposal page

coreyar:
## Changes
Tests
- casting votes
- interacting with proposal


src/pages/Proposal/ProposalSummary/index.tsx
```diff
@@ -91,13 +100,16 @@ export const ProposalSummaryUi: React.FC<
...
   let transactionHash = startTxHash;
+
   switch (state) {
     case 'Active':
       updateProposalButton = (
         <SecondaryButton
           onClick={handleCancelProposal}
           css={styles.updateProposalButton}
           loading={isCancelProposalLoading}
+          data-testid={TEST_IDS.proposalSummary.cancelButton}
```
therealemjy, line 111: In this instance we can use the content directly to target the button since it doesn't include any string interpolation

src/pages/Proposal/ProposalSummary/index.tsx
```diff
@@ -113,6 +125,7 @@ export const ProposalSummaryUi: React.FC<
           onClick={handleQueueProposal}
           css={styles.updateProposalButton}
           loading={isQueueProposalLoading}
+          data-testid={TEST_IDS.proposalSummary.queueButton}
```
therealemjy, line 128: In this instance we can use the content directly to target the button since it doesn't include any string interpolation

src/pages/Proposal/ProposalSummary/index.tsx
```diff
@@ -125,6 +138,7 @@ export const ProposalSummaryUi: React.FC<
           onClick={handleExecuteProposal}
           css={styles.updateProposalButton}
           loading={isExecuteProposalLoading}
+          data-testid={TEST_IDS.proposalSummary.executeButton}
```
therealemjy, line 141: In this instance we can use the content directly to target the button since it doesn't include any string interpolation

src/pages/Proposal/VoteModal/index.tsx
```diff
@@ -85,10 +87,16 @@ const VoteModal: React.FC<IVoteModal> = ({
...
+              id="reason"
               placeholder={t('vote.addComment')}
               css={styles.comment}
             />
-            <FormikSubmitButton enabledLabel={title} fullWidth loading={isVoteLoading} />
+            <FormikSubmitButton
+              enabledLabel={title}
+              fullWidth
+              loading={isVoteLoading}
+              data-testid={TEST_IDS.voteModal.submitButton}
```
therealemjy, line 98: In this instance we can use the content directly to target the button since it doesn't include any string interpolation

coreyar: This is a valid solution. It is better than using the translation string because the text appears multiple times on the page. Using a testId allows the selection of the specific component we want to use

src/pages/Proposal/ProposalSummary/index.tsx
coreyar: This is a valid solution. It is better than using the translation string because the text appears multiple times on the page. Using a testId allows the selection of the specific component we want to use

src/pages/Proposal/ProposalSummary/index.tsx
coreyar: This is a valid solution. It is better than using the translation string because the text appears multiple times on the page. Using a testId allows the selection of the specific component we want to use

src/pages/Proposal/ProposalSummary/index.tsx
coreyar: This is a valid solution. It is better than using the translation string because the text appears multiple times on the page. Using a testId allows the selection of the specific component we want to use

src/pages/Proposal/VoteModal/index.tsx
coreyar: This is a valid solution. It is better than using the translation string because the text appears multiple times on the page. Using a testId allows the selection of the specific component we want to use

therealemjy (approved):

PR #736, opened 2022-07-07
[VEN-298] Remove RefreshContext

therealemjy:
## Jira ticket(s)

VEN-298

## Changes

- rename parameters from various queries and functions to be consistent
- remove `RefreshContext` and its associated hooks
- add `refetchInterval` prop to all relevant queries to effectively replace the refresh context


coreyar (approved):
Looks good

PR #738, opened 2022-07-07
[VEN-393] Use 8 decimals to display token values + improve responsiveness of LabeledInlineContent

therealemjy:
## Jira ticket(s)

VEN-393

## Changes

- always use 8 decimals at minimum to display token values
- display columns of `LabeledInlineContent` as rows on mobile to gain horizontal space
- update how supply balance is displayed on Supply form so it takes less horizontal space (token symbol is now displayed in the header rather than next to each token value)


coreyar (approved):
👍

PR #746, opened 2022-07-11
Create shared formatToProposal util

therealemjy:
## Changes

- create `formatToProposal` to prevent duplicated code


src/utilities/formatToProposal.ts
```diff
@@ -51,22 +87,13 @@ const formatToProposal = ({
     descriptionObj = { version: 'v1' as const, title: plainTitle, description: descriptionText };
   }
 
-  if (!endDate) {
```
coreyar, line 54: Are we guaranteed to have an endDate? I took this from the original implementation that was calculate the end date based on the endBlock if it was missing

src/utilities/formatToProposal.ts
```diff
@@ -87,6 +114,16 @@ const formatToProposal = ({
     startTxHash: startTxHash ?? undefined,
     totalVotesWei: abstainedVotesWei.plus(againstVotesWei).plus(forVotesWei),
   };
+
+  if (actions) {
```
coreyar: Actions are required to submit a proposal so there should always be at least 1

src/utilities/formatToProposal.ts
therealemjy: Nope, that's a mistake on my end! Good catch

src/pages/Proposal/index.tsx
```diff
@@ -105,7 +105,7 @@ export const ProposalUi: React.FC<ProposalUiProps> = ({
         />
       </div>
 
-      <Description description={proposal.description} actions={proposal.actions} />
+      <Description description={proposal.description} actions={proposal.actions || []} />
```
coreyar: An alternative philosophy is to format all data in the `format` functions so it is usable in the frontend. For example, arrays returned as undefined would be set as an empty array when formatting.

src/utilities/formatToProposal.ts
therealemjy: They are yes, but the `/voters/history/` endpoint does not return any `actions` field which is why this check is here

coreyar: Interesting, that seems like a bug on the endpoint

therealemjy: My thought as well. In the meantime, I've followed your advice of assigning an empty array of no action was present in the input

coreyar (approved):

PR #748, opened 2022-07-12
[VEN-382] Format action signature on frontend

coreyar:
## Jira ticket(s)

VEN-382

## Changes

- Renames create proposal form field so action shape resembles the shape returned from the backend
- Removes reliance on returned formatted title and formats the action signature on the frontend. This removes the need to keep contract addresses synced on the frontend and backend.

### DIFF
--- a/src/components/ReadableActionSignature/formatSignature.ts
+++ b/src/components/ReadableActionSignature/formatSignature.ts
@@ -1,0 +1,39 @@
+import { ethers } from 'ethers';
+import type { Result } from '@ethersproject/abi';
+import { FormValues } from 'pages/Vote/CreateProposalModal/proposalSchema';
+import { IProposalAction } from 'types';
+
+const formatSignature = (action: FormValues['actions'][number] | IProposalAction) => {
+  try {
+    const fragment = ethers.utils.FunctionFragment.from(action.signature || '');
+    let args: Result = [];
+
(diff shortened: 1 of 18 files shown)


src/components/ReadableActionSignature/formatSignature.ts
```diff
@@ -0,0 +1,29 @@
+import { ethers } from 'ethers';
+import type { Result } from '@ethersproject/abi';
+
+const formatSignature = (
+  action:
+    | { signature: string; data: (string | undefined)[]; target: string }
+    | { data: string; signature: string; target: string },
```
therealemjy: I'd suggest putting this in a `type` file in this component folder so it can be imported from this util and the component itself rather than being duplicated

therealemjy (commented):
I spotted two issues:
- old proposals can't load anymore and the page crashes (link: http://localhost:3000/vote/proposal/47)
Screen Shot 2022-07-13 at 12 18 56
 - addresses no longer display inside double quotes:
 Screen Shot 2022-07-13 at 12 21 31

coreyar:
> | { data: string; signature: string; target: string },

Nice catch. I added a test that represents the bad data saved from the previous frontend.
Updated the types as well

therealemjy (approved):

PR #749, opened 2022-07-13
[VEN-391] Display time until proposal can be executed

therealemjy:
## Jira ticket(s)

VEN-391

## Changes

- add `getProposalEta` query function
- display time until proposal can be executed on Proposal Details page
- hide "Execute" button while proposal's ETA hasn't been passed yet


src/translation/translations/en.json
```diff
@@ -657,6 +657,7 @@
       "succeededTimestamp": "Succeeded: <Span>{{date, dd MMM yyyy HH:mm}}</Span>"
     },
     "queue": "Queue",
+    "readyToQueueAt": "Ready to queue at",
```
coreyar: This should be `Ready to execute at`

src/pages/Proposal/ProposalSummary/index.tsx
```diff
@@ -157,6 +161,26 @@ export const ProposalSummaryUi: React.FC<
...
+    if (state === 'Active' && endDate) {
+      return {
+        date: endDate,
+        // DO NOT REMOVE COMMENT: needed by i18next to extract translation key
+        // t('voteProposalUi.activeUntilDate')
+        i18nKey: 'voteProposalUi.activeUntilDate',
+      };
+    }
+
+    if (state === 'Succeeded' && proposalEta && isAfter(proposalEta, new Date())) {
```
coreyar:
```suggestion
    if (state === 'Queued') {
```

src/pages/Proposal/ProposalSummary/index.tsx
```diff
@@ -157,6 +161,26 @@ export const ProposalSummaryUi: React.FC<
...
+      return {
+        date: proposalEta,
+        // DO NOT REMOVE COMMENT: needed by i18next to extract translation key
+        // t('voteProposalUi.readyToQueueAt')
+        i18nKey: 'voteProposalUi.readyToQueueAt',
```
coreyar:
```suggestion
        i18nKey: 'voteProposalUi.readyToExecuteAt',
```

src/pages/Proposal/ProposalSummary/index.tsx
```diff
@@ -157,6 +161,26 @@ export const ProposalSummaryUi: React.FC<
...
+        // t('voteProposalUi.activeUntilDate')
+        i18nKey: 'voteProposalUi.activeUntilDate',
+      };
+    }
+
+    if (state === 'Succeeded' && proposalEta && isAfter(proposalEta, new Date())) {
+      return {
+        date: proposalEta,
+        // DO NOT REMOVE COMMENT: needed by i18next to extract translation key
+        // t('voteProposalUi.readyToQueueAt')
```
coreyar: nit: an option that wouldn't require the comment is to assign the `components` prop to an object so that it can be reused for both conditions in a returned `Trans` component here

src/pages/Proposal/ProposalSummary/index.tsx
therealemjy: You mean:
```
if (state === 'Queued' && proposalEta && isAfter(proposalEta, new Date())) {
```

Right? Since we want to display the waiting period

coreyar: the waiting period is the length of the time the proposal will be in the queued state. So the proposalEta is when the queued state will end

therealemjy: Yeah, so we need to check that the ETA isn't passed to determine whether we display the countdown, otherwise it is indeed ready to be executed and so the countdown would be over

src/pages/Proposal/ProposalSummary/index.tsx
therealemjy: Do you mean doing something like this?
Screen Shot 2022-07-14 at 15 24 43
Because in this case we'd be duplicating the component itself, when only the i18nkey changes

coreyar: I was thinking about extracting the props into variables at the top of the function to reduce the duplication, because I understand this was done to reduce duplication. Either way looks good to me, I was just thinking how we can avoid having comments that affect code

therealemjy: I see. Yeah I'd rather keep the comment, because this way we clearly indicate only the translation key changes

coreyar (approved):

PR #778, opened 2022-07-28
bump package version to 1.1.0

therealemjy:


coreyar (approved):

coreyar:
This PR looks like it needs to be rebased. By its title it looks like it should only change the version in the package.json

coreyar (approved):

PR #781, opened 2022-07-29
[VEN-358] Force LUNA and UST disabling as collateral

therealemjy:
## Jira ticket(s)

VEN-358

## Changes

- update design of modal to indicate users need to disable LUNA/UST as collateral
- only allow users to use the voting features and repay their UST/LUNA loans or withdraw their UST/LUNA if they have either of these tokens enabled as collateral


src/context/DisableLunaUstWarning/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const DisableLunaUstWarningProvider: React.FC = ({ children }) => {
+  const { account } = useContext(AuthContext);
+  const accountAddress = account?.address || '';
+  const {
+    data: { assets },
+  } = useGetUserMarketInfo({
+    accountAddress,
+  });
+
+  const [shouldShowLunaUstWarningModal, setShouldShowModal] = useState(false);
```
coreyar: nit: Should implies a question or that an action can happen in the future 
```suggestion
  const [showLunaUstWarningModal, setShowModal] = useState(false);
```

therealemjy: I disagree. `showLunaUstWarningModal` sounds like a function name to me, I'd expect it to be callable like this:
```
showLunaUstWarningModal();
```

We can change `shouldShowLunaUstWarningModal` if you think it's unclear what it represents, but my goal with this variable is to emulate the same logic as with `isLoading` for example: it implies a state and a boolean.

coreyar (approved):

src/context/DisableLunaUstWarning/index.tsx
coreyar: should is not clear. You could also think `shouldShowLunaUstWarningModal` is callable and that you would receive a variable back telling you whether the modal should be shown or not. I only removed the word `should` as the main comment because it is not a precise word for describing state but conveys additional connotations that aren't applicable here. 
Other names could be something like `[isOpenLunaUstWarningModal, setIsOpenModal]`. We use similar names to this in other places

therealemjy: I see. How about `isLunaUstWarningModalOpen`/`isLunaUstWarningModalShown` then?

coreyar (approved):

PR #784, opened 2022-07-29
hotfix/ active menu

coreyar:
## Changes

A bug was introduced when we changed the dashboard to the root where the active menu item for the dashboard is stuck on active

This PR also removes some unused styles

### DIFF
--- a/src/components/Layout/Sidebar/Link/index.tsx
+++ b/src/components/Layout/Sidebar/Link/index.tsx
@@ -1,6 +1,6 @@
 /** @jsxImportSource @emotion/react */
 import React from 'react';
-import { NavLink } from 'react-router-dom';
+import { NavLink, useLocation } from 'react-router-dom';
 
 import { MenuItem } from '../../types';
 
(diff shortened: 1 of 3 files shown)


src/components/Layout/Sidebar/Link/index.tsx
```diff
@@ -11,11 +11,16 @@ export interface LinkProps {
...
   if (href[0] === '/') {
     const activeClassName = isMobile ? 'active-mobile-menu-item' : 'active-menu-item';
-
     return (
-      <NavLink to={href} onClick={onClick} activeClassName={activeClassName}>
+      <NavLink
+        to={href}
+        onClick={onClick}
+        activeClassName={activeClassName}
+        isActive={() => href === pathname}
```
therealemjy, line 22: I think this would prevent sub-pages from marking the parent page as active in the sidebar menu.

You might have missed it, but I already opened a PR to fix this issue: https://github.com/VenusProtocol/venus-protocol-interface/pull/782
I took a different approach and set a conditional `exact` prop on the NavLink so the root route is only marked as active if the current URL is the root and the root only.

PR #796, opened 2022-08-04
[VEN-476] Fetch v2 proposals

therealemjy:
## Jira ticket(s)

VEN-476

## Changes

- pass `version` query parameter set to `v2` when fetching proposals
- update `Proposal` type to v2


src/utilities/formatToProposal.ts
```diff
@@ -81,7 +81,14 @@ const formatToProposal = ({
     endDate = now;
   }
 
-  let descriptionObj = { version: 'v1' as const, title: '', description: '' };
+  let descriptionObj: Proposal['description'] = {
```
coreyar, line 84: This shouldn't change. v1 descriptions are not changed by the v2 endpoint. v1 descriptions are markdown where as v2 are JSON

src/utilities/formatToProposal.ts
```diff
@@ -69,19 +65,17 @@ const formatToProposal = ({
   queuedTxHash,
   startTxHash,
   actions,
-  blockNumber,
 }: FormatToProposalInput): Proposal => {
-  let endDate = endTimestamp ? createDateFromSecondsTimestamp(endTimestamp) : undefined;
-
-  if (!endDate && blockNumber) {
```
coreyar: This handle the case were endDate is undefined

therealemjy: Yeah, but `endTimestamp` is never `undefined` it seems

coreyar (approved):

PR #799, opened 2022-08-09
[VEN-492] Create RiskLevel component

therealemjy:
## Jira ticket(s)

VEN-492

## Changes

- create RiskLevel component
Screen Shot 2022-08-09 at 10 53 58

### DIFF
--- a/src/components/RiskLevel/types.ts
+++ b/src/components/RiskLevel/types.ts
@@ -1,0 +1,1 @@
+export type RiskLevelVariant = 'low' | 'medium' | 'high';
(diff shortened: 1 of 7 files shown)


src/components/RiskLevel/types.ts
```diff
@@ -0,0 +1 @@
+export type RiskLevelVariant = 'low' | 'medium' | 'high';
```
coreyar: This comment is to provide context on the RiskRating constants and make sure we are aligned.
The contract enum has more variations.
```
enum RiskRating {
        VERY_HIGH_RISK,
        HIGH_RISK,
        MEDIUM_RISK,
        LOW_RISK,
        MINIMAL_RISK
    }
  ```
  It will also come a cross as an integer 0 - 5. This could be mapped to these values in the format function of the query. We could also format them in the subgraph.

coreyar (approved):

src/components/RiskLevel/types.ts
therealemjy: That's very helpful, thank you!

I think using the risk level integer might be the simplest, since this way the component doesn't need to import the enum from the contract types to define how to display the content.

PR #808, opened 2022-08-15
[VEN-494] Update dashboard UI

therealemjy:
## Jira ticket(s)

VEN-494

## Changes

- add small variant to `TextField` component
- add magnifier icon
- add `ButtonGroup` component
- update font sizes used in `Select` component
- add light variant to `Toggle` component
- add hooks to handle breakpoints in CSS only
- remove unused `Pool` type
- update UI of Dashboard to fit new designs (features will be added in another PR). Note that the supply market and borrow market tables haven't been updated yet; this will come in a PR of its own


src/pages/Dashboard/index.tsx
```diff
@@ -1,37 +1,160 @@
...
-  </>
-);
+  const showXlDownCss = useShowXlDownCss();
+  const hideXlDownCss = useHideXlDownCss();
+
+  // TODO: add all options
+  const mobileSelectOptions = [
+    {
+      value: 'riskLevel',
+      label: 'Risk level',
```
coreyar, line 47: Looks like this should be a translation

coreyar (approved):

src/hooks/responsive.ts
```diff
@@ -11,3 +13,35 @@ export const useIsSmDown = () => useBreakpointDown('sm');
...
+  const theme = useTheme();
+
+  return css`
+    ${theme.breakpoints.down(breakpoint)} {
+      display: none;
+    }
+  `;
+};
+
+export const useHideXsDownCss = () => useBreakpointHideDown('xs');
```
coreyar, line 27: I don't see where some of these are used. I also don't think we want to prefer using  hooks for responsiveness

src/pages/Dashboard/index.tsx
therealemjy: Yeah, there's a TODO next to the select to the variable to remind to add all the options. This will come with a next PR once I wire the page to the contracts

src/hooks/responsive.ts
therealemjy: For now only a few are used, but they are util hooks that can be used in any component.

They declare styles the way we currently do (using hooks and Emotions' `css` function), so as a result only CSS is output. The reason I created them is so the DOM can be a bit more explicit when it comes to defining what elements show on what screen dimensions and to avoid having to duplicate CSS. If you look in the codebase, there's currently quite a few places declaring the same CSS, something like:
```js
css`
  display: block;
 
  ${theme.breakpoints.down('md')} {
    display: none;
  }
`
```
Using these hooks, we don't have to write down this logic for each component and the DOM becomes easier to read in my opinion, since you'll have something like this:

```jsx
<ElementAlwaysVisible />

<ElementVisibleOnDesktopOnly css={[styles.customStyling, hideXlDownCss]} />

<ElementVisibleOnMobileOnly css={[styles.customStyling, showXsDownCss]} />
```

coreyar: Yeah, this is a downside of css in js. It isn't as easy to compose classes. Is this the best way to reduce the duplicated code though, because the tradeoff is adding event listeners to the dom. 

One alternative would be to assign the `css` block in your example to a variable and importing and interpolating that variable.

coreyar: I conflated these hooks with the ones that do use listeners. Looks like these don't. I like this for sharing styles

coreyar (approved):

PR #828, opened 2022-09-07
[VEN-553] Create MarketBreakdown component

therealemjy:
## Jira ticket(s)

VEN-553

## Changes

- create MarketBreakdown component
- rename local type `RiskLevel` to `MarketRiskLevel` and move to global type file
- add `supplyBalance`, `borrowBalance` and `percentOfLimit` columns to `MarketTable` component
- make minor styling update on `CellGroup` component


src/pages/Account/MarketBreakdown/useExtractData.ts
```diff
@@ -0,0 +1,93 @@
+/** @jsxImportSource @emotion/react */
```
coreyar: This pragma isn't necessary because there are no components using the css prop in this file

src/pages/Account/MarketBreakdown/useExtractData.ts
```diff
@@ -0,0 +1,93 @@
...
+  dailyXvsDistributionInterestsCents,
+}: {
+  assets: Asset[];
+  isXvsEnabled: boolean;
+  dailyXvsDistributionInterestsCents?: BigNumber;
+}) =>
+  useMemo(() => {
+    const { totalBorrowCents, totalSupplyCents, borrowLimitCents } = assets.reduce(
+      (acc, asset) => ({
+        ...acc,
```
coreyar: It looks like this isn't necessary because all properties are getting overwritten

coreyar (approved):

src/pages/Account/MarketBreakdown/useExtractData.ts
therealemjy: Indeed, good catch

src/pages/Account/MarketBreakdown/useExtractData.ts
therealemjy: Yep, nice 👍

PR #835, opened 2022-09-14
[VEN-448] Fix hypothetical earnings calculation

therealemjy:
## Jira ticket(s)

VEN-448

## Changes

- add keys to market tables on dashboard to fix a bug where React would not consider both tables as being different elements, resulting in display issues including with the sorting feature
- add distribution APRs to formatted assets fetched from the API
- use XVS distribution APRs in `calculateYearlyEarningsForAsset` util instead of the total daily XVS distribution


src/clients/api/queries/getMarkets/index.ts
```diff
@@ -24,32 +24,36 @@ const getMarkets = async (): Promise<GetMarketsOutput> => {
...
-          treasuryTotalSupplyCents: new BigNumber(activeMarket.totalSupplyUsd).times(100),
-        };
-        return [...acc, formattedActiveMarket];
-      }
-      return acc;
-    }, []);
+
+    markets = response.data?.data.markets
+      // Filter out unlisted tokens
+      .filter(market =>
```
coreyar: This can be combined into `reduce` loop

src/utilities/calculateYearlyEarnings.ts
```diff
@@ -14,23 +18,38 @@ export const calculateYearlyEarningsForAsset = ({ asset }: { asset: Asset }) =>
...
   // the borrow APY is expressed with a negative percentage)
-  const borrowYearlyInterestsCents = assetBorrowBalanceCents.multipliedBy(
+  const borrowYearlyEarningsCents = assetBorrowBalanceCents.multipliedBy(
     asset.borrowApy.dividedBy(100),
   );
 
-  return supplyYearlyEarningsCents.plus(borrowYearlyInterestsCents);
+  const yearlyEarningsCents = supplyYearlyEarningsCents.plus(borrowYearlyEarningsCents);
+
+  if (!includeXvs || !asset.xvsSupplyApr.isFinite() || !asset.xvsBorrowApr.isFinite()) {
```
coreyar, line 29: I think this is more readable with postive checks, doing the necessary calculations in the if statement and returning `yearlyEarningsCents` if the if statement is not entered

therealemjy: I'd rather not. Doing what I'm doing here means we can use a return statement and avoid nesting.

src/clients/api/queries/getMarkets/index.ts
therealemjy: Good point

coreyar (approved):

PR #837, opened 2022-09-15
[VEN-607] Add VAI page

therealemjy:
## Jira ticket(s)

VEN-607

## Changes

- add VAI page to Mint and Repay VAI
- disable hover effect on `TextField` component when the component is disabled


src/pages/Vai/MintVai/getReadableFeeVai.ts
```diff
@@ -0,0 +1,21 @@
...
+import { VAI_ID } from '../constants';
+
+const getReadableFeeVai = ({
+  valueWei,
+  mintFeePercentage,
+}: {
+  valueWei: BigNumber;
+  mintFeePercentage: number;
+}) => {
+  const feeWei = new BigNumber(valueWei || 0).multipliedBy(mintFeePercentage).dividedBy(100);
```
coreyar: The 0 could be set as a default value in the destruction of the argument. It would make this line a bit more readable

src/pages/Vai/MintVai/index.spec.tsx
```diff
@@ -0,0 +1,129 @@
...
+      amount: {
+        tokenId: TOKENS.vai.id as TokenId,
+        valueWei: fakeGetMintableVaiOutput.mintableVaiWei,
+      },
+      content: expect.any(String),
+      title: expect.any(String),
+    });
+  });
+
+  // @TODO: add tests to cover failing scenarios
```
coreyar: Do we have a ticket for this?

src/pages/Vai/MintVai/getReadableFeeVai.ts
coreyar: Actually if the type is always BigNumber, this should never default to 0 because BigNumber would always be truthy even if it isn't valid

coreyar (approved):

src/pages/Vai/MintVai/getReadableFeeVai.ts
therealemjy: Good point yes, I took this code from what we currently have in production without checking it since it's test covered. I'll update this

src/pages/Vai/MintVai/index.spec.tsx
therealemjy: No, I just created one and linked it

coreyar (approved):

PR #838, opened 2022-09-23
Update breadcrumb implementation

therealemjy:
## Changes

- update breadcrumb implementation to be based on a central component responsible for rendering breadcrumbs for all pages, rather than letting pages set their own breadcrumbs individually


src/pages/Asset/index.tsx
```diff
@@ -407,18 +369,14 @@ export type AssetProps = RouteComponentProps<{ vTokenId: VTokenId; marketId: str
...
 const Asset: React.FC<AssetProps> = ({
   match: {
-    params: { vTokenId, marketId },
+    params: { vTokenId },
   },
 }) => {
-  const { account } = useContext(AuthContext);
   const vToken = getVBepToken(vTokenId);
 
-  // TODO: fetch actual values (see VEN-546)
```
stubbi, line 416: is that TODO not necessary anymore?

therealemjy: It isn't. Check the line after this comment, you'll see a variable set with a fixed value which has now ben removed and that was the one that needed to be fetched originally.

coreyar (approved):

PR #840, opened 2022-09-28
[VEN-608] Add warning about isolated assets

therealemjy:
## Jira ticket(s)

VEN-608

## Changes

- create `IsolatedAssetWarning` component
- add `IsolatedAssetWarning` instance to the supply and borrow modals
- unify styling used to render scrollbars within the dApp
- rename `arrowDown` icon to `arrowUp`

### DIFF
--- a/src/components/IsolatedAssetWarning/AssetTable/index.tsx
+++ b/src/components/IsolatedAssetWarning/AssetTable/index.tsx
@@ -11,6 +11,7 @@ import { Token } from '../../Token';
 import { useStyles as useParentStyles } from '../styles';
 import { WarningType } from '../types';
 import { useStyles } from './styles';
+import TEST_IDS from './testIds';
 
 export interface AssetTableProps {
   type: WarningType;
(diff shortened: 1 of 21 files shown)


src/components/IsolatedAssetWarning/AssetTable/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+    <div css={styles.container}>
+      <div css={[styles.row, styles.headerRow]}>
+        {headerColumnLabels.map(headerColumnLabel => (
+          <div css={styles.cell} key={`isolated-asset-warning-header-column-${headerColumnLabel}`}>
+            <Typography variant="small2">{headerColumnLabel}</Typography>
+          </div>
+        ))}
+      </div>
+
+      <div css={styles.dataRowList} data-testid={TEST_IDS.assetTable}>
```
stubbi, line 51: the `TEST_IDS` is intended here?

src/components/IsolatedAssetWarning/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export const IsolatedAssetWarning: React.FC<IsolatedAssetWarningProps> = ({
+  assetId,
+  type,
+  className,
+}) => {
+  // TODO: fetch actual value (see VEN-546)
+
+  const pool = poolData[0];
```
stubbi, line 85: any risk `poolData[0]` is not set and might break here?

src/components/IsolatedAssetWarning/AssetTable/index.tsx
therealemjy: It is yes. Test IDs are used to select specific elements when running tests. You can check this file to see this in action: `src/components/IsolatedAssetWarning/index.spec.tsx`

src/components/IsolatedAssetWarning/index.tsx
therealemjy: Not in this case, since `poolData` is currently loaded from a JSON file which contains samples of pools.

src/components/IsolatedAssetWarning/AssetTable/index.tsx
stubbi: alright!

src/components/IsolatedAssetWarning/index.tsx
stubbi: perfect 👍

stubbi (approved):

PR #843, opened 2022-10-04
[VEN-656] Add fallback message when website cannot load from IPFS

therealemjy:
## Jira ticket(s)

VEN-656

## Changes

- add fallback message when website cannot load from IPFS

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -98,7 +98,9 @@
     <noscript>
       Venus is a decentralized finance (DeFi) algorithmic money market protocol on BNB Chain.
     </noscript>
-    <div id="root"></div>
+    <div id="root">
+      Site could not be loaded from IPFS. Please wait a few minutes then refresh the page.
+    </div>
     <!--
       This HTML file is a template.
       If you open it directly in the browser, you will see an empty page.


public/index.html
```diff
@@ -98,7 +98,9 @@
     <noscript>
       Venus is a decentralized finance (DeFi) algorithmic money market protocol on BNB Chain.
     </noscript>
-    <div id="root"></div>
+    <div id="root">
+      Site could not be loaded from IPFS. Please wait a few minutes then refresh the page.
```
coreyar, line 102: What do you think about looping Pavel in and creating a design so that it looks cool?

coreyar: This also isn't accurate which is how we introduced the discussion around redirection. The site will fail to load with the switch to hash routing so we need to direct the user to a route with a hash by either sending them to the home page or optimistically adding a hash to the route they tried to access

If the route they are trying to access has changed and doesn't exist anymore, this would still work because we will have successfully loaded the app and react-router can render a 404 page

therealemjy: This only covers the case where IPFS is failing to load the website: it isn't related to hash routing.

Once we switch to hash routing, if a user tries to access a page that no longer exists (say https://app.venus.io/markets), React Router will automatically redirect them to the home page thanks to the `<Redirect />` tag we already have in the code.

coreyar: Due to the change to hash routing and the way our CDN resolves files the example you gave will result in an error that the ipfs was unable to resolve the page. The react bundle will not load. 

This is why I propose adding a hash to the route so that the urls is resolved, the bundle loads and react router redirects. It is also why the current message doesn't reflect the solution / problem.

therealemjy: We don't have the possibility to add a hash from to the route from the client as far as I can, because the file that's loaded when navigating to a page without a hash isn't a file that comes from this project. If you got to this link for example: https://testnet.venus.io/markets, the file loaded isn't the index file of our project but a file that's loaded from somewhere else.

coreyar: That is correct. I overlooked that. We need to think more about how we can handle this because this still leaves us without a nice solution for broken links.

For the sake of this PR the only outstanding comment is if we can design a nice looking fallback instead of a single line of text

therealemjy: It got me confused too. I think we should just close this PR, as the fallback would actually never work in the case where IPFS doesn't load the page. I would only be displayed if React failed to load and render any DOM, which wouldn't be related to IPFS with the current setup.

PR #846, opened 2022-10-14
[VEN-691] Implement URL pagination

therealemjy:
## Jira ticket(s)

VEN-691

## Changes

- link pagination to URL on relevant pages

### DIFF
--- a/src/hooks/useUrlPagination.ts
+++ b/src/hooks/useUrlPagination.ts
@@ -1,0 +1,45 @@
+import { useEffect, useMemo } from 'react';
+import { RouteComponentProps } from 'react-router-dom';
+
+export type UseUrlPaginationParams = Pick<RouteComponentProps, 'location' | 'history'>;
+
+export const PAGE_PARAM_NAME = 'page';
+
+const useUrlPagination = ({ location, history }: UseUrlPaginationParams) => {
+  const { search } = location;
+
+  useEffect(() => {
+    // Add page param to URl if none has been set
+    const searchParams = new URLSearchParams(search);
+
+    if (!searchParams.get(PAGE_PARAM_NAME)) {
+      // Note: although the pagination starts from 0, we make it start from 1 in the
+      // URL to make it more user-friendly. This is something we need to account for
+      // when updating the page search param
+      searchParams.append(PAGE_PARAM_NAME, '1');
+      history.replace(`${location.pathname}?${searchParams.toString()}`);
+    }
+
+    // Scroll to the top of the page on search change
+    window.scrollTo(0, 0);
+  }, [search]);
+
+  const currentPage = useMemo(() => {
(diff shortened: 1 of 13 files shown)


src/hooks/useUrlPagination.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const useUrlPagination = ({ location, history }: UseUrlPaginationInput): UseUrlPaginationOutput => {
+  const { search } = location;
+
+  useEffect(() => {
+    // Add page param to URl if none has been set
+    const searchParams = new URLSearchParams(search);
+
+    if (!searchParams.get(PAGE_PARAM_NAME)) {
+      // Note: although the pagination starts from 0, we make it start from 1 in the
```
coreyar, line 21: This shouldn't be necessary. We discussed handling this on the backend and versioned the endpoint to paginate with an initial page value of 1 on the transactions endpoint. 

We can do the same for proposals

coreyar: Actually proposals paginates with a limit and offset not a page so a page number isn't applicable.

src/hooks/useUrlPagination.ts
```diff
@@ -0,0 +1,50 @@
...
+      // URL to make it more user-friendly. This is something we need to account for
+      // when updating the page search param
+      searchParams.append(PAGE_PARAM_NAME, '1');
```
coreyar: We should use set here because append will fail if the key already exists (is an empty string). This block assumes that the value is `undefined` on searchParams

src/pages/Vote/index.stories.tsx
```diff
@@ -1,14 +1,23 @@
 import { ComponentMeta } from '@storybook/react';
 import React from 'react';
+import { State } from 'react-powerplug';
```
coreyar, line 3: nit: This is done two ways in the project currently. Our own hook and this dependency.

src/hooks/useUrlPagination.ts
therealemjy: As of today, this isn't implemented so we need the front end to take care of it (like it currently does). We'll need to update this logic once the backend is updated.

As per the pagination, the implementation in this PR does work with a page number (you can spin up the app and check). That part I actually didn't change: that's how the current implementation is.

src/hooks/useUrlPagination.ts
therealemjy: Good catch

src/pages/Vote/index.stories.tsx
therealemjy: Yeah, we had this discussion a little while back. This package has a few additional things that can be beneficial.
I'd say it's a very minor thing and for now wouldn't spend time on it.

coreyar (approved):

PR #847, opened 2022-10-19
[VEN-697] Add skeleton for Swap page

therealemjy:
## Jira ticket(s)

VEN-697

## Changes

- add skeleton for Swap page
- update sidebar so items can display a "NEW" badge

### DIFF
--- a/src/components/Layout/Sidebar/index.tsx
+++ b/src/components/Layout/Sidebar/index.tsx
@@ -52,6 +52,8 @@ export const SidebarUi: React.FC = () => {
                 disableRipple
               >
                 <Link href={menuItem.href}>
+                  <div className="left-border" />
+
                   <ListItemIcon css={styles.listItemIcon}>
                     <Icon name={menuItem.icon} />
                   </ListItemIcon>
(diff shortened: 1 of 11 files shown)


src/components/Layout/Sidebar/index.tsx
```diff
@@ -60,7 +62,13 @@ export const SidebarUi: React.FC = () => {
                     {t(menuItem.i18nKey)}
                   </Typography>
 
-                  <div className="left-border" />
+                  {menuItem.isNew && (
+                    <div css={styles.listItemNewBadge}>
```
coreyar, line 66: Would it make sense to use our chip component here?

src/components/Layout/Sidebar/index.tsx
```diff
@@ -124,6 +132,14 @@ export const SidebarUi: React.FC = () => {
                     >
                       {t(i18nKey)}
                     </Typography>
+
+                    {isNew && (
+                      <div css={styles.listItemNewBadge}>
```
coreyar, line 137: Would it make sense to use our chip component here?

coreyar (approved):

src/components/Layout/Sidebar/index.tsx
therealemjy: Unfortunately it doesn't have the same dimensions and styling :/

src/components/Layout/Sidebar/index.tsx
therealemjy: Same as above

PR #849, opened 2022-10-20
[VEN-697] Add Swap page UI

therealemjy:
## Jira ticket(s)

VEN-697

## Changes

- add UI for the Swap page. Note: I've left a ton of TODOs without links to JIRA tickets since I'm actively working on them. I'll split the work on the UI into multiple PRs to avoid having to review a massive chunk of code in one go
- update `SelectTokenTextField` component


src/pages/Swap/index.tsx
```diff
@@ -1,8 +1,236 @@
...
+      fromTokenAmountTokens: currentFormValues.toTokenAmountTokens,
+      toToken: currentFormValues.fromToken,
+      toTokenAmountTokens: currentFormValues.fromTokenAmountTokens,
+      direction:
+        currentFormValues.direction === 'exactAmountIn' ? 'exactAmountOut' : 'exactAmountIn',
+    }));
+
+  return (
+    <Paper css={styles.container}>
+      <SelectTokenTextField
```
coreyar, line 98: Is there a reason we are not using our formik here?

src/pages/Swap/useGetSwapInfo/index.ts
```diff
@@ -0,0 +1,82 @@
+import BigNumber from 'bignumber.js';
+import { useMemo } from 'react';
+import { Token } from 'types';
+import { convertTokensToWei } from 'utilities';
+
+import { Swap } from '../types';
+
+// TODO: move to global constants
```
coreyar: We can resolve this TODO in this PR

coreyar: I see there are other todos in this PR that relate to this so it could also wait

src/pages/Swap/types.ts
```diff
@@ -0,0 +1,25 @@
+import BigNumber from 'bignumber.js';
+import { Token } from 'types';
+
+interface SwapBase {
+  fromToken: Token;
+  toToken: Token;
+  exchangeRate: BigNumber;
+  direction: 'exactAmountIn' | 'exactAmountOut';
```
coreyar: Since this strings are using in multiple types here and on the swap page they could be constants

src/pages/Swap/index.tsx
```diff
@@ -1,8 +1,236 @@
...
 
-const Swap: React.FC = () => <SwapUi />;
+import { useStyles } from './styles';
+import { Swap } from './types';
+import useGetSwapInfo from './useGetSwapInfo';
 
-export default Swap;
+const tokenIds = Object.keys(TOKENS) as TokenId[];
+
+const SLIPPAGE_TOLERANCE_PERCENTAGE = 0.5;
```
coreyar: Global constant from hook

src/pages/Swap/useGetSwapInfo/index.ts
therealemjy: Definitely 👍

src/pages/Swap/types.ts
therealemjy: Yeah that's a good point

src/pages/Swap/index.tsx
therealemjy: Yeah I'll move it to a global file

src/pages/Swap/index.tsx
therealemjy: Ideally we'd use it, but unfortunately that would force us to store form values in two different ways because `Formik` only allows to access form values from within the `<Formik>` tag and in this case its validation schema and submit function are defined based on the form values.
For that reason I figured it would make the logic complicated to understand if we were to use Formik.

coreyar: You can handle both of those cases with formik
Validation schema - We have an example of a dynamic validation schema on the proposals form
Submit function - you can wrap the submit functions in another function with logic about how they are called

Do we need error messages on this form? I don't see any logic for that. Formik would handle that as well

therealemjy: Validation schema: we do, but the value it uses to be defined dynamically does not come form the form itself, hence why it's doable.
Submit function: same issue here. Since we want to display the best trade possible to the user while they're completing the form, then using Formik means the logic for fetching and displaying the trade needs to be contained within the `Formik` tag. However, when submitting the form we then don't have access to this data since the `onSubmit` handler is attached to the `Formik` tag. The solution for that would be to store the trade data again in a state variable that sits outside of the `Formik` tag.

Using Formik here feels like a hack and will make the codebase harder to understand.

As per error messages, no we shouldn't need any. We'll probably only make the border of the `fromToken` input red when the user enters a value that's higher than their current balance (similarly to what we do in other forms).

coreyar (approved):

PR #850, opened 2022-10-28
[VEN-698] Generate testnet swap

therealemjy:
## Jira ticket(s)

VEN-698

## Changes

This is very much a WIP, but because the swap feature involves a lot of changes I need to stop this PR here or it will become unmanageable.
In the current state, the front end is able to find a swap similar to what PancakeSwap would propose on testnet. I've left as many TODOs as possible so I won't forget anything, since there's still a lot of work to be done.

Because we now handle tokens that aren't supported by Venus and since the list will get quite big (PancakeSwap's extended token list, which is the one I'll load in the app, has 533 tokens), we can no longer use our current way of defining tokens within the app since we rely on generated token IDs to fetch information about a given token.
The next step for me then is to update how we define tokens within the dApp to come up with a more flexible solution (https://jira.toolsfdg.net/browse/VEN-712).


src/pages/Swap/index.tsx
```diff
@@ -34,33 +38,41 @@ interface FormValues {
...
-      }));
-    }
+  // TODO: reinitialize form values if swap becomes invalid
+  // if (!swapInfo) {
+  //   setFormValues(currentFormValues => ({
+  //     ...currentFormValues,
+  //     fromTokenAmountTokens: '',
+  //     toTokenAmountTokens: '',
+  //   }));
+  // }
```
therealemjy, line 73: Keeping this here as an example

coreyar (approved):
Looks good.

PR #851, opened 2022-11-01
[VEN-712] Update how tokens are defined

therealemjy:
## Jira ticket(s)

VEN-712

## Changes
We currently assume the list of tokens is known and always the same regardless of what feature it is used in. That's why so far we relied on using token IDs (generated from their symbol) to grab token details.
Now that we will support isolated lending and swapping through PancakeSwap, we can't always assume the list of tokens is always the same (nor that any given token will have the same address for all features - for example PancakeSwap uses different token addresses than us on testnet).

The goal of this PR is to update how we define tokens and import their information so we rely much less on token IDs and thus can be much more modular in our way of writing features. The end goal would be to move away from using token IDs altogether, but that will involve further changes that would make this PR even bigger.

Although the amount of changes is enormous due to the fact this is an update to a core logic of the app, the actual updates are as follows:
- update `Token` type. The `id` field is now a string to better reflect the fact token IDs aren't known in advance anymore
- update how tokens are stored and declared within the app. We had token icons in multiple places and all the tokens were stored in one big file. They're now split into folders and files for each token list and each environment
- deprecate `getToken` and `getVToken` (now renamed `unsafelyGetToken` and `unsafelyGetVToken`), in favor of importing tokens directly and providing all the token information necessary for utilities such as `convertWeiToTokens` instead of relying on a token ID. As described earlier, these are still relied upon in some places in the app, but the end goal would be to get rid of them completely (see: https://jira.toolsfdg.net/browse/VEN-723)

Other changes:
- split `Icon` component into `Icon` and `TokenIcon`. `TokenIcon` can now be used to render a token icon and optionally its symbol next to it
- update `Asset` type. Token information is now contained in a `token` field of the type `Token`
- `iconName` prop of the `LabeledInlineContent` component has now been renamed `iconSrc` and accepts an `IconName` (string) or a `Token`, rendering an `Icon` or `Token` instance based on prop type
- update `convertTokensToWei`, `convertWeiToTokens` and `formatTokensToReadableValue` utilities to now take a `token` parameter (of the type `Token`) rather than a token ID


src/constants/tokens/pancakeSwap/mainnet.ts
```diff
@@ -0,0 +1,5 @@
+// import { Token, BscChainId } from 'types';
+
+export const MAINNET_PANCAKE_SWAP_TOKENS = {
+  // TODO: add tokens
```
therealemjy, line 4: This will be handled in the next PR

src/constants/tokens/pancakeSwap/testnet.ts
```diff
@@ -1,36 +1,29 @@
...
     id: 'wbnb',
     symbol: 'WBNB',
     decimals: 18,
     address: '0xae13d989daC2f0dEbFf460aC112a837C89BAa7cd',
     asset:
-      'https://pancakeswap.finance/images/tokens/0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c.png',
-  },
+      'https://pancakeswap.finance/images/tokens/0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c.svg',
+  } as Token,
+  // TODO: add other tokens (BNB for example)
```
therealemjy, line 28: This will be handled in the next PR

src/components/TokenIcon/index.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+          <h4 style={{ marginRight: '8px' }}>{token.symbol}</h4>
+
+          <TokenIcon token={token} />
+        </div>
+      ))}
+    </div>
+  </>
+);
+
+export const WithShowSymbol = () => <TokenIcon token={TOKENS.usdc} showSymbol />;
```
coreyar: nit: Functions prefixed with `with` make me think they are story modifiers. Maybe `ShowSymbol` or TokenIconWithSymbol

coreyar (approved):

src/components/TokenIcon/index.stories.tsx
therealemjy: Works for me 👍

PR #853, opened 2022-11-02
[VEN-698] Swap feature (missing approval step)

therealemjy:
## Jira ticket(s)

VEN-698

## Changes

As the title indicates, this is the implementation of the swap feature. The dApp is now capable of finding the best trade possible for a given swap and to execute. The final step is to give approval to the PancakeSwap router contract so tokens can be exchanged on behalf of the user (you can already test the feature by swapping BNB to any other token, since BNB doesn't require any approval being the native currency of BNC). Here's the full list of changes:
- create `swapToken` mutation function that can execute any given swap
- update `getBalanceOf` query function; it now accepts a `Token` instance as parameter instead of just a token ID to make it more modular. In this case, since PancakeSwap uses different tokens than us in testnet it means we can fetch the user balance for each of them (where before we could only fetch the balances of our custom testnet tokens)
- update `getTokenContract` function: it now accepts a `Token` instance as parameter instead of just a token ID for modularity
- add `PancakeRouter` contract (currently pointing to PancakeSwap's router until ours is fixed)
- add `isNative` field to `Token` type to differentiate the native currency (BNB) from other tokens
- add form validation
- add all mainnet PancakeSwap tokens, based on PancakeSwap's extended token list (https://github.com/pancakeswap/token-list/blob/main/src/tokens/pancakeswap-extended.json)


src/clients/api/queries/getBalanceOf/index.ts
```diff
@@ -16,15 +17,15 @@ export type GetBalanceOfOutput = {
 const getBalanceOf = async ({
   web3,
   accountAddress,
-  tokenId,
+  token,
 }: GetBalanceOfInput): Promise<GetBalanceOfOutput> => {
   let balanceWei: BigNumber;
 
-  if (tokenId === 'bnb') {
+  if (token.isNative) {
```
coreyar, line 24: I think this isNative property is nice

src/clients/api/queries/useGetUserMarketInfo/__snapshots__/index.spec.tsx.snap
```diff
@@ -126,6 +126,7 @@ Object {
         "asset": "bnb.svg",
         "decimals": 18,
         "id": "bnb",
+        "isNative": true,
```
coreyar, line 129: I think this isNative property is nice

src/components/TokenIcon/index.stories.tsx
```diff
@@ -57,4 +57,4 @@ export const Default = () => (
   </>
 );
 
-export const ShowSymbol = () => <TokenIcon token={TOKENS.usdc} showSymbol />;
+export const WithShowSymbol = () => <TokenIcon token={TOKENS.usdc} showSymbol />;
```
coreyar: nit: with makes me think it is a story decorator and not a story itself

coreyar (approved):

src/components/TokenIcon/index.stories.tsx
therealemjy: Oops yes sorry, this change got lost when I closed the other PR, I should have rebased from it.

coreyar (approved):

PR #858, opened 2022-11-09
[VEN-726] Add tests to cover Swap (Part 2)

therealemjy:
## Jira ticket(s)

VEN-726

## Changes

- add tests to cover Swap page


src/components/SelectTokenTextField/index.tsx
```diff
@@ -63,13 +66,15 @@ export const SelectTokenTextField: React.FC<SelectTokenTextFieldProps> = ({
                 small
                 css={styles.getButton({ isTokenListShown })}
                 disabled={disabled}
+                data-testid={`${testId}${TEST_ID_FRAGMENTS.tokenSelectButton}`}
```
coreyar: Another option would be to make `tokenSelectButton a function that returns the formatted testId. This way it removes uncertainty about how the testId should be formatted

coreyar (approved):

src/components/SelectTokenTextField/index.tsx
therealemjy: Ah yes that's a nice suggestion, I'll update.

coreyar (approved):

PR #863, opened 2022-11-25
Use token and vToken instead of IDs in query and mutation functions

therealemjy:
In order to prepare the dApp for isolated lending pools, we need to move away from the current way of getting token information, which is through generated token IDs and now deprecated functions (`unsafelyGetToken` and `unsafelyGetVToken`).
This PR removes token IDs usage from query and mutation functions, which in turn requires a lot of updates onto the rest of the dApp.

## Other changes

- merge `repayBnb` and `repayNonBnbVToken` into a single `repay` mutation function
- rename `borrowVToken` to `borrow`
- create `useGetUserAsset` hook to fetch a single asset. Once the subgraph updated, this hook will also need to be updated so it fetches the data from subgraph rather than user `useGetUserMarketInfo` (in the the future we can no longer assume `useGetUserMarketInfo` will always fetch all the assets supported on Venus)
- lowercase addresses when comparing them to one another


src/clients/api/mutations/repay/index.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+import repay, { REPAYMENT_BNB_BUFFER_PERCENTAGE } from '.';
+
+const fakeAmountWei = new BigNumber(10000000000000000);
+
+jest.mock('clients/contracts');
+
+describe('api/mutation/repay', () => {
+  describe('repay non-BNB loan', () => {
+    test('throws an error when request fails', async () => {
+      const fakeWeb3 = {} as any;
```
coreyar: Could we use an automatic module mock from jest here?

therealemjy: We could yes, but it won't be used. That's why I passed a plain object with the `any` type: to indicate this isn't being used in the test

src/hooks/useBorrowRepayModal/Modal/Repay/index.tsx
```diff
@@ -207,24 +207,27 @@ export const RepayForm: React.FC<RepayFormProps> = ({
...
 
-const Repay: React.FC<RepayProps> = ({ asset, onClose, includeXvs }) => {
+const Repay: React.FC<RepayProps> = ({ token, vToken, onClose, includeXvs }) => {
   const { t } = useTranslation();
   const { account } = React.useContext(AuthContext);
 
-  const vBepTokenContractAddress = unsafelyGetVToken(asset.token.id).address;
+  const {
+    data: { asset },
+  } = useGetUserAsset({ token });
```
coreyar, line 222: In some places we switched from asset to token but in others we still use asset? Do they have different definitions?

src/hooks/useBorrowRepayModal/Modal/Repay/index.tsx
```diff
@@ -207,24 +207,27 @@ export const RepayForm: React.FC<RepayFormProps> = ({
...
 export interface RepayProps {
-  asset: Asset;
+  token: Token;
+  vToken: Token;
   includeXvs: boolean;
   onClose: () => void;
 }
 
-const Repay: React.FC<RepayProps> = ({ asset, onClose, includeXvs }) => {
+const Repay: React.FC<RepayProps> = ({ token, vToken, onClose, includeXvs }) => {
```
coreyar, line 216: nit: seems like we could use just the token asset and get the related vToken from the asset object but I didn't dig deep into that idea

src/pages/Market/index.tsx
```diff
@@ -364,40 +364,42 @@ export const MarketUi: React.FC<MarketUiProps> = ({
...
 
+// TODO: pass vTokenAddress and poolAddress instead of IDs
 export type MarketProps = RouteComponentProps<{ vTokenId: string; poolId: string }>;
 
 const Market: React.FC<MarketProps> = ({
   match: {
     params: { vTokenId },
   },
 }) => {
+  // TODO: update to use token address
```
coreyar, line 375: Should this be done in this PR?

src/types/index.ts
```diff
@@ -11,6 +11,7 @@ export interface User {
 }
 
 export interface Asset {
+  // TODO: add vToken
```
coreyar, line 14: If we did this I think we could reduce passing tokens and vToken as props in repay and borrow components

src/clients/api/mutations/repay/index.spec.ts
coreyar: Yeah, could get rid of the any type though. It would also be a more complete mock where it does get used. Those also receive any types

src/hooks/useBorrowRepayModal/Modal/Repay/index.tsx
therealemjy: Right now, an asset represents a specific type we have on the front end and that contains a token, its corresponding vToken and some financial values associated to it (reserves, token price etc..). You can find the type definition in the main type file: `src/types/index.ts`

src/hooks/useBorrowRepayModal/Modal/Repay/index.tsx
therealemjy: I want to keep this as modular as possible, and asset has a lot of fields that aren't all relevant to this component, so we're better off passing token and vToken and letting the component fetch the data it needs

src/pages/Market/index.tsx
therealemjy: Impossible without updating the `Table` component so it's more modular, which is what I'm currently working on.

src/types/index.ts
therealemjy: We'd still pass them individually to improve modularity, but yes this will be done as the refactors are happening

src/clients/api/mutations/repay/index.spec.ts
therealemjy: It works when the fake Web3 instance isn't used by the function it's passed to, but it doesn't when we need to mock the `sendTransaction` method from web3.

I made the test and the mocked Web3 instance is an empty object:
Screen Shot 2022-11-29 at 11 32 31

I can replace the `as any` with `as unknown as Web3` as a compromise, like we do for contracts?

src/hooks/useBorrowRepayModal/Modal/Repay/index.tsx
coreyar: I mistyped and meant to just write token. I was suggesting removing a prop rather than adding one

coreyar: but yeah from your previous comment it sounds like asset would work here too in place of both if it didn't have those extra fields

src/clients/api/mutations/repay/index.spec.ts
coreyar: Ahh looks like you'd set it up manually on that mock like we do now.  Either options seems good.

coreyar (approved):

PR #874, opened 2022-12-01
[VEN-810] Added type checked subgraph queries

gleiser-oliveira:
## Jira ticket(s)

VEN-810

## Changes

- Using [Graph Client](https://github.com/graphprotocol/graph-client), this PR adds support for strongly typed subgraph queries.
- Made adjustments to the build process so the related files get generated on the postinstall hook
- Added an example of a typed query using `AssetsInAccount`


.env.template
```diff
@@ -4,3 +4,4 @@ TSC_COMPILE_ON_ERROR=true
 #REACT_APP_CHAIN_ID should be 97 for testnet
 REACT_APP_CHAIN_ID=97
 REACT_APP_GOOGLE_ANALYTICS_ID=
+SUBGRAPH_GRAPHQL_ENDPOINT=https://api.thegraph.com/subgraphs/name/venusprotocol/venus-isolated-pools-chapel
```
coreyar: nit: newline at the end of a file is nice

src/constants/endpoints.ts
```diff
@@ -5,6 +5,12 @@ export const API_ENDPOINT_URLS = {
   [BscChainId.TESTNET]: 'https://testnetapi.venus.io/api',
 };
 
+export const SUBGRAPH_ENDPOINT_URLS = {
+  [BscChainId.MAINNET]: 'https://api.thegraph.com/subgraphs/name/venusprotocol/venus-subgraph',
```
coreyar:
```suggestion
  [BscChainId.MAINNET]: 'https://api.thegraph.com/subgraphs/name/venusprotocol/venus-isolated-pools',
```

.env.template
coreyar: If you set this in ci.yaml. The tests should pass.

src/config/index.ts
```diff
@@ -20,12 +21,14 @@ const isOnTestnet = chainId === BscChainId.TESTNET;
 const rpcUrl = sample(RPC_URLS[chainId]) as string;
 const apiUrl = API_ENDPOINT_URLS[chainId];
 const bscScanUrl = BSC_SCAN_URLS[chainId];
+const subgraphUrl = SUBGRAPH_ENDPOINT_URLS[chainId];
```
therealemjy: This seems to be unused with your solution, so we should remove it

src/constants/endpoints.ts
```diff
@@ -5,6 +5,13 @@ export const API_ENDPOINT_URLS = {
   [BscChainId.TESTNET]: 'https://testnetapi.venus.io/api',
 };
 
+export const SUBGRAPH_ENDPOINT_URLS = {
+  [BscChainId.MAINNET]:
+    'https://api.thegraph.com/subgraphs/name/venusprotocol/venus-isolated-pools',
+  [BscChainId.TESTNET]:
+    'https://api.thegraph.com/subgraphs/name/venusprotocol/venus-isolated-pools-chapel',
+};
+
```
therealemjy: These can be removed since we use an environment variable to grab the endpoint URL

src/clients/subgraph/queries/getAssetsInAccount/assetsInAccountQuery.ts
```diff
@@ -0,0 +1,12 @@
...
+import FunctionKey from 'constants/functionKey';
+
+import subgraphClient from '../../subgraphClient';
+
+const useGetAssetsInAccountSubgraph = (accountAddress: string) =>
+  useQuery([FunctionKey.GET_ASSETS_IN_ACCOUNT, { accountAddress }], () =>
+    subgraphClient.AssetsInAccountQuery({ accountAddress }),
+  );
+
+export default useGetAssetsInAccountSubgraph;
```
therealemjy: I'd rename this file to `useGetAssetsInAccountSubgraph` since it only exports the hook of the same name.

scripts/run-generate-subgraph-types.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+if [[ "${REACT_APP_CHAIN_ID}" == "56" ]]; then
+  export SUBGRAPH_GRAPHQL_ENDPOINT="https://api.thegraph.com/subgraphs/name/venusprotocol/venus-isolated-pools"
+else
+  export SUBGRAPH_GRAPHQL_ENDPOINT="https://api.thegraph.com/subgraphs/name/venusprotocol/venus-isolated-pools-chapel"
+fi
```
therealemjy, line 7: Nice 👍

.env.template
```diff
@@ -4,3 +4,4 @@ TSC_COMPILE_ON_ERROR=true
...
```
therealemjy: We can remove this then

therealemjy (approved):
Added a minor comment about removing `SUBGRAPH_GRAPHQL_ENDPOINT` from `.env.template` since we now define the subgraph endpoint from the Yaml file.

I tested it locally and it works seamlessly. Well done!

PR #877, opened 2022-12-02
[VEN-812] Fixed condition to cancel proposals

gleiser-oliveira:
## Jira ticket(s)

[VEN-812](https://jira.toolsfdg.net/browse/VEN-812)

## Changes
Proposals can now be cancelled if either one of the following conditions are true:
- The connected account is also the proposer
- If the proposer's voting power is less than the proposal threshold, meaning anyone could cancel the proposal if this condition is true

### DIFF
--- a/src/pages/Proposal/ProposalSummary/index.tsx
+++ b/src/pages/Proposal/ProposalSummary/index.tsx
@@ -272,14 +272,15 @@ const ProposalSummary: React.FC<ProposalSummaryUiProps> = ({ className, proposal
     proposalId: proposal.id,
   });
 
-  const { data: currentVotesData } = useGetCurrentVotes(
-    { accountAddress },
+  const { data: proposerVotesData } = useGetCurrentVotes(
+    { accountAddress: proposal.proposer },
     { enabled: !!accountAddress },
   );
 
   const canCancelProposal =
-    proposalThresholdData?.thresholdWei &&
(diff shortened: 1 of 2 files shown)


src/pages/Proposal/ProposalSummary/index.tsx
```diff
@@ -272,14 +272,15 @@ const ProposalSummary: React.FC<ProposalSummaryUiProps> = ({ className, proposal
...
+    { accountAddress: proposal.proposer },
     { enabled: !!accountAddress },
   );
 
   const canCancelProposal =
-    proposalThresholdData?.thresholdWei &&
-    currentVotesData?.votesWei.isGreaterThanOrEqualTo(proposalThresholdData?.thresholdWei);
+    proposal.proposer === account?.address ||
+    (proposalThresholdData?.thresholdWei &&
+      proposerVotesData?.votesWei.isLessThan(proposalThresholdData?.thresholdWei));
```
therealemjy: The logic looks correct to me.

A very small thing, which comes from the previous code and not yours:
```suggestion
    proposal.proposer === account?.address ||
    (proposalThresholdData?.thresholdWei &&
      proposerVotesData?.votesWei.isLessThan(proposalThresholdData.thresholdWei));
```

therealemjy (approved):

therealemjy (approved):
Nice, looks good 👍

PR #881, opened 2022-12-07
[VEN-821] Display token balances on swap

therealemjy:
## Jira ticket(s)

VEN-821

## Changes

This PR looks scary but the high amount of added lines is mainly made of snapshots and mocks.

- remove token balance display from `SelectTokenTextfield` (this is in preparation for when we'll integrate the swap the borrow and supply modals using this component)
- create `getTokenBalances` query function
- update `SelectTokenTextfield` to display token balances
- fetch token balances from Swap to display on `SelectTokenTextfield`


src/pages/Swap/index.tsx
```diff
@@ -66,14 +66,40 @@ const SwapPageUi: React.FC<SwapPageUiProps> = ({
...
+  tokenBalances,
 }) => {
   const styles = useStyles();
-  const { t } = useTranslation();
+  const { t, Trans } = useTranslation();
 
   const { openSuccessfulTransactionModal } = useSuccessfulTransactionModal();
 
+  const { fromTokenUserBalanceWei, toTokenUserBalanceWei } = useMemo(() => {
+    console.log(tokenBalances, formValues.fromToken.address);
```
gleiser-oliveira: Stray log?

therealemjy: Indeed! I removed it

gleiser-oliveira (approved):

PR #884, opened 2022-12-14
[VEN-810] Fix: Now uses @graphql-codegen/cli to generate subgraph types

gleiser-oliveira:
## Jira ticket(s)

VEN-810

## Changes

- Replaced [Graph Client](https://github.com/graphprotocol/graph-client) with [GraphQL Code Generator](https://the-guild.dev/graphql/codegen), aiming to fix the runtime issues found.
- Just like before, it will read the SUBGRAPH_GRAPHQL_ENDPOINT to reach out our subgraph and schema and generate TypedDocumentNodes corresponding to each of the .graphql files present in the project.
- Main difference is we now use `graphql-request` during runtime. This avoids issues with .mjs files included in the previously generated code.


scripts/run-generate-subgraph-types.sh
```diff
@@ -6,4 +6,4 @@ else
   export SUBGRAPH_GRAPHQL_ENDPOINT="https://api.thegraph.com/subgraphs/name/venusprotocol/venus-isolated-pools-chapel"
 fi
 
-yarn graphclient build --dir ./src/types/subgraph/
+yarn graphql-code-generator --config ./src/types/subgraph/codegen.ts
```
therealemjy: Because we now need to access `SUBGRAPH_GRAPHQL_ENDPOINT` from within the app in order to call the subgraph using `graphql-request`, I'd make it available from the config directly (which you can find at `src/config/index.ts`) using the same logic as for the API and RPC URLs.

  This way it can then be imported from `codegen.ts` using the config directly (and doesn't need to be set in the `.env` file, since it's anyway static).
  
  This means this script file can be removed, and instead the command `yarn graphql-code-generator --config ./src/types/subgraph/codegen.ts` can be used directly inside the `package.json` file without needing to set any environment variable.

src/clients/subgraph/queries/getAssetsInAccount/useGetAssetsInAccountSubgraph.ts
```diff
@@ -1,12 +1,12 @@
...
 
 import FunctionKey from 'constants/functionKey';
-
-import subgraphClient from '../../subgraphClient';
+import { AssetsInAccountQueryDocument, subgraphEndpoint } from 'types/subgraph/gql/queries';
 
 const useGetAssetsInAccountSubgraph = (accountAddress: string) =>
   useQuery([FunctionKey.GET_ASSETS_IN_ACCOUNT, { accountAddress }], () =>
-    subgraphClient.AssetsInAccountQuery({ accountAddress }),
+    request(subgraphEndpoint, AssetsInAccountQueryDocument, { accountAddress }),
```
therealemjy: In order to keep everything contained within modules, we could create a subgraph client just like before except using `graphql-request` and the generated types.
This way everything would be imported from one place (and mocked in tests when necessary).

src/types/subgraph/codegen.ts
```diff
@@ -0,0 +1,23 @@
...
+const config: CodegenConfig = {
+  schema: process.env.SUBGRAPH_GRAPHQL_ENDPOINT,
+  documents: ['src/clients/subgraph/**/*.graphql'],
+  generates: {
+    './src/types/subgraph/gql/queries.ts': {
+      plugins: [
+        {
+          add: {
+            placement: 'append',
+            content: `export const subgraphEndpoint = '${process.env.SUBGRAPH_GRAPHQL_ENDPOINT}';`,
```
therealemjy: Following my previous comment, the subgraph endpoint would be imported from the config:
```suggestion
            content: `export const subgraphEndpoint = '${config.subgraphEndpointUrl}';`,
```

src/types/subgraph/codegen.ts
```diff
@@ -0,0 +1,23 @@
...
+        },
+        'typescript',
+        'typed-document-node',
+        'typescript-operations',
+      ],
+    },
+  },
+};
+
+export default config;
```
therealemjy: I'd move this file and the generated types inside the subgraph client folder (`src/clients/subgraph`) then create an index file that would export all the generated types and documents as well as the subgraph client. This way everything would be contained within the same module and can be imported using one absolute import.

```ts
import { subgraphClient, AssetsInAccountQueryDocument } from 'clients/subgraph';
```

.eslintignore
```diff
@@ -5,3 +5,4 @@ build
 storybook-static
 i18next-parser.config.js
 coverage
+src/types/subgraph/gql/queries.ts
```
therealemjy: That's a detail, but we can probably ignore the whole folder here
```suggestion
src/types/subgraph/gql
```

therealemjy (commented):
Looks really good, I like that approach much better than using The Graph's client.

src/types/subgraph/codegen.ts
gleiser-oliveira: I think I have to keep this as it was, because if I try to import config in the codegen file, it won't understand the imports inside `config/index.ts`:

```
import { BscChainId } from 'types';
import { BSC_SCAN_URLS } from 'constants/bsc';
import { API_ENDPOINT_URLS, RPC_URLS } from 'constants/endpoints';
```

Because those are aliased imports as set in tsconfig, and the CLI needs relative paths. Going down this route means changing all the related imports to be like:

```
import { BscChainId } from '../types';
import { BSC_SCAN_URLS } from '../constants/bsc';
import { API_ENDPOINT_URLS, RPC_URLS } from '../constants/endpoints';
```
I think it's just simpler to keep the script and the env var, what do you think?

therealemjy (approved):

PR #888, opened 2022-12-18
[FIX] Fix overflow issue with large values on dashboard

therealemjy:
For wallets handling large amount of funds, values tend to overflow. Example:
Screen Shot 2022-12-16 at 3 55 57 PM

Given the dashboard works as an overview of the account health, we don't need to be extra precise with the values we display and we actually gain in clarity by shortening them:
Screen Shot 2022-12-18 at 15 22 18


gleiser-oliveira (commented):
@therealemjy This looks very nice, I'm wondering if we shouldn't get a diff from the regressions, maybe it's not covered?

therealemjy:
@gleiser-oliveira That's a good point. It might not get detected as a change that's significant enough by `reg-viz` (we've set a certain threshold for changes)

gleiser-oliveira (approved):

PR #893, opened 2022-12-20
Remove XVS APR inclusion toggle

therealemjy:
## Context

After speaking with Pavel, the choice has been made to support the "Include XVS distribution APR" toggle from the Account page. Since isolated pools will now yield different reward tokens, keeping this toggle doesn't make sense anymore and could confuse users since that's a UX that was unique to us.

## Changes

- remove `includeXvsContext` and everything linked to it


gleiser-oliveira (approved):
Not only it simplifies the UX, it also makes the code cleaner, nice change

PR #894, opened 2022-12-20
[VEN-831] feat: add proposal type to create proposal form

coreyar:
## Jira ticket(s)

VEN-831

## Changes

This PR updates the create proposal form to include the proposal type select field


src/pages/Vote/CreateProposalModal/index.tsx
```diff
@@ -49,6 +50,17 @@ export const CreateProposal: React.FC<CreateProposalProps> = ({
         title: t('vote.pages.proposalInformation'),
         Component: () => (
           <>
+            <FormikSelectField
+              name="proposalType"
+              title={t('vote.createProposalForm.proposalType')}
+              ariaLabel={t('vote.createProposalForm.proposalType')}
+              css={styles.formBottomMargin}
```
therealemjy, line 58: You'll need to add a label saying "Proposal type" here.
Screen Shot 2022-12-23 at 12 23 08

src/types/index.ts
```diff
@@ -100,6 +100,7 @@ export interface Proposal {
   againstVotesWei: BigNumber;
   createdDate: Date | undefined;
   description: DescriptionV1 | DescriptionV2;
+  proposalType: 'Normal' | 'FastTrack' | 'Critical';
```
therealemjy: Instead of redeclaring this type in multiple places, you can use an enum:

```ts
enum ProposalTypeName {
    Normal,
    FastTrack,
    Critical
}
```

There currently is a type called `proposalTypeName` at `src/types/index.tsx`, this enum would replace it.

src/utilities/formatToProposal.ts
```diff
@@ -96,6 +98,11 @@ const formatToProposal = ({
   const againstVotesWei = new BigNumber(againstVotes || 0);
   const forVotesWei = new BigNumber(forVotes || 0);
 
+  const proposalTypeName = ['Normal', 'FastTrack', 'Critical'][proposalType] as
+    | 'Normal'
+    | 'FastTrack'
+    | 'Critical';
```
therealemjy: Following my previous comment, you'd then be able to use the enum here:
```ts
const proposalTypeName = ProposalTypeName[proposalType];
```

src/pages/Vote/CreateProposalModal/proposalSchema.ts
```diff
@@ -77,6 +77,7 @@ const proposalSchema = yup.object({
   forDescription: yup.string().required(ErrorCode.VALUE_REQUIRED),
   againstDescription: yup.string().required(ErrorCode.VALUE_REQUIRED),
   abstainDescription: yup.string().required(ErrorCode.VALUE_REQUIRED),
+  proposalType: yup.string().oneOf(['0', '1', '2']).required(ErrorCode.VALUE_REQUIRED),
```
therealemjy: Given the proposal type is represented as a number, I'd use a `number` type here instead of a `string`

src/pages/Vote/CreateProposalModal/index.tsx
coreyar: ahh yeah I missed the select only has a mobile header

src/utilities/formatToProposal.ts
coreyar: It can't be done this way because enums can't be indexed by an integer but I can add the type and improve on this. I was also trying to format this from a number to a descriptive string constant here. The enum would have to be 
```
enum ProposalTypeName {
    Normal = 'Normal',
    FastTrack = 'FastTrack',
    Critical = 'Critical'
}
```
and we'd still need a method to map the number  to this enum

therealemjy: Enums are indexed using an integer by default:
```ts
enum ProposalTypeName {
    Normal,
    FastTrack,
    Critical
}

ProposalTypeName[0] // This will return "Normal" as a string
```

coreyar: cool didn't know it worked both ways I thought it this only set default values. Thanks

therealemjy: No worries

therealemjy (approved):

gleiser-oliveira (approved):

PR #904, opened 2022-12-27
[VEN-833] Decoupled Supply and Withdraw modals

gleiser-oliveira:
## Jira ticket(s)

VEN-833

## Changes

- Refactored supply and withdraw to have their own components/modals/tests.
- This allows us to customize each modal on its own when integrating isolated lending.


src/hooks/useSupplyWithdrawModal/Modal/Withdraw/index.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+    (useGetMainAssets as jest.Mock).mockImplementation(() => ({
+      data: {
+        assets: [fakeAsset],
+        userTotalBorrowLimitCents: fakeUserTotalBorrowLimitDollars,
+        userTotalBorrowBalanceCents: fakeUserTotalBorrowBalanceDollars,
+      },
+      isLoading: false,
+    }));
+  });
```
therealemjy, line 51: Now that both tabs are divided, I'd add a test that checks that the withdrawable amount displayed is correct. What do you think?
Screen Shot 2022-12-29 at 10 53 39

I'd also add the same test as for the supply form to check that the submit button is disabled if no value has been provided ("submit is disabled with no amount").

src/hooks/useSupplyWithdrawModal/Modal/Supply/form.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+  const hypotheticalTokenSupplyBalance = amountValue
+    ? calculateNewBalance(asset.userSupplyBalanceTokens, amount)
+    : undefined;
+
+  const hypotheticalBorrowLimitCents = useMemo(() => {
+    const tokenPriceDollars = getBigNumber(asset?.tokenPriceDollars);
+    let updateBorrowLimitCents;
+
+    if (tokenPriceDollars && validAmount) {
```
therealemjy: I know that's from the previous code, but I think we should remove `getBigNumber` here. Its usage is a bit confusing, since it will return a 0 BigNumber if the parameter its given is `undefined`. Let me know your opinion, but I think we should get rid of the `getBigNumber` util from the codebase altogether (it's only used in the supply/withdraw modals).

 ```suggestion
    let updateBorrowLimitCents;

    if (asset?.tokenPriceDollars && validAmount) {
```

File: src/hooks/useSupplyWithdrawModal/Modal/Supply/form.tsx [hunk 3]
```diff
@@ -0,0 +1,281 @@
...
+        <ValueUpdate original={userTotalBorrowLimitCents} update={hypotheticalBorrowLimitCents} />
+      </LabeledInlineContent>
+
+      <LabeledInlineContent
+        label={t('supplyWithdraw.dailyEarnings')}
+        css={styles.getRow({ isLast: true })}
+        className="info-row"
+      >
+        <ValueUpdate original={dailyEarningsCents} update={hypotheticalDailyEarningCents} />
+      </LabeledInlineContent>
```
therealemjy: I'd put this in a separate `AccountData` component like for the borrow modal, so it can be shared between the Supply and Withdraw tabs

therealemjy (commented):
Really nice work 👌

I've added a few comments. Also now the `SupplyWithdrawForm.tsx` file (_src/hooks/useSupplyWithdrawModal/Modal/SupplyWithdrawForm.tsx_) can be removed

src/hooks/useSupplyWithdrawModal/Modal/Supply/form.tsx
gleiser-oliveira: Agreed and removed, as it was it was just making it a bit harder to follow the logic.

src/hooks/useSupplyWithdrawModal/Modal/Withdraw/index.spec.tsx
gleiser-oliveira: Added a test for the withdrawable amount. Testing "submit is disabled with no amount" is already done, line 178 in the supply test file.

therealemjy: Nice.

Indeed the test file of the Supply tab has this test, but the Withdraw tab doesn't which is why I think we should add it there as well. Does that make sense?

src/hooks/useSupplyWithdrawModal/Modal/Supply/form.tsx
therealemjy: Thanks!

therealemjy:
It looks really good!
It's ready to be merged for me (the last thing for me is I would add the "submit is disabled with no amount" test to the Withdraw tab).

Regarding the Typescript errors, they're due to these changes:
https://github.com/VenusProtocol/venus-protocol-interface/pull/899/files#diff-afd34f87d1748f6ef2ad9885bb81569bd465ec75de4ca5a9884a547145cded64L35-L36

https://github.com/VenusProtocol/venus-protocol-interface/pull/899/files#diff-251a40cde70df180e683a39f3fff11e1721f202e1eac1466dd8af7801b815e5fR13-R15

`useGetAsset` and `useGetMainAssets` now optionally return the `data` prop, where before it was always returned. I changed this to follow the same behavior as other queries that use `react-query`.

src/hooks/useSupplyWithdrawModal/Modal/Withdraw/index.spec.tsx
gleiser-oliveira: Oooohh, I misunderstood your first message, alright, adding it now

therealemjy (approved):

PR #905, opened 2022-12-30
[VEN-548] Wire up Account page

therealemjy:
## Jira ticket(s)

VEN-548

## Changes

- add `userWalletBalanceCents`, `userSupplyBalanceCents` and `userBorrowBalanceCents` props to `Asset` type
- update `userPercentOfLimit` prop type of `Asset` type to a number
- create `convertDollarsToCents` util
- fix columns on `MarketTable` component
- fix how `compareNumbers` handles `undefined` values
- wire up Account page


src/clients/api/queries/useGetMainAssets/index.ts
```diff
@@ -183,22 +194,23 @@ const useGetMainAssets = ({
...
-          supplyBalanceCents: new BigNumber(market?.totalSupplyUsd || 0)
-            .times(100)
-            .dp(0)
-            .toNumber(),
-          borrowBalanceCents: new BigNumber(market?.totalBorrowsUsd || 0)
-            .times(100)
-            .dp(0)
-            .toNumber(),
+          borrowBalanceCents: convertDollarsToCents(
+            market?.totalBorrowsUsd ? +market?.totalBorrowsUsd : 0,
```
gleiser-oliveira: `+market` is intentional?

therealemjy: It is yes, because `totalBorrowsUsd` is a string so we convert it to a number. However you highlighted that the second optional chaining is unnecessary since we've already assessed that `market.totalBorrowsUsd` isn't `undefined` hehe

I'll update that.

Edit: turns out all the optional chainings are unnecessary in this file, since `market` is always defined.

gleiser-oliveira (approved):

gleiser-oliveira (approved):

PR #919, opened 2023-01-12
[VEN-876] VAI repay fee is now calculated and shown (develop)

gleiser-oliveira:
## Jira ticket(s)

VEN-876

## Changes

- Added the repay fee to the `Repay VAI` component
- Fee is calculated after a 2 second debouce based on the repayment amount set by the user
- Added tests for the repay fee
- Updated VAIController ABI


src/pages/Vai/RepayVai/index.tsx
```diff
@@ -117,6 +120,8 @@ export const RepayVaiUi: React.FC<IRepayVaiUiProps> = ({
                   >
                     {readableRepayableVai}
                   </LabeledInlineContent>
+
```
therealemjy, line 123: On line 117, the `isLast` prop needs to be set to `false` since it's to the last row anymore

therealemjy (approved):

src/pages/Vai/RepayVai/index.tsx
gleiser-oliveira: Done, thanks!

PR #920, opened 2023-01-12
[VEN-859] Fetch isolated pools

therealemjy:
## Jira ticket(s)

VEN-859

## Changes

- point to temporary testnet graph endpoint
- update `PoolRiskRating` to use generated subgraph type
- index VTokens by address
- remove `id` field from tokens and vTokens
- remove unused `notNull` util
- remove `assetsInAccount` subgraph query
- create `isolatedPools` subgraph query
- wire `useGetIsolatedPools` up


src/clients/api/__mocks__/index.ts
```diff
@@ -115,6 +115,13 @@ export const useGetTreasuryTotals = jest.fn();
 
 export const useGetMainPoolTotalXvsDistributed = jest.fn();
 
+export const useGetIsolatedPools = jest.fn(() => ({
+  isLoading: false,
+  data: {
+    pools: [], // TODO: add isolated pools
```
gleiser-oliveira: Should we add mocks here now rather than later?

src/clients/api/queries/getIsolatedPools/index.ts
```diff
@@ -0,0 +1,135 @@
...
+import { COMPOUND_MANTISSA } from 'constants/compoundMantissa';
+import oracleAbi from 'constants/contracts/abis/oracle.json';
+
+import getTokenBalances, { GetTokenBalancesOutput } from '../getTokenBalances';
+import formatToPool from './formatToPool';
+import { FormatToPoolInput, GetIsolatedPoolsInput, GetIsolatedPoolsOutput } from './types';
+
+export type { GetIsolatedPoolsInput, GetIsolatedPoolsOutput } from './types';
+
+// TODO: add tests
```
gleiser-oliveira: And also making sure we got this under our radar like the mocks

therealemjy: Yeah good catch, I should have removed this comment after I added the tests

src/clients/api/__mocks__/index.ts
therealemjy: Definitely, I'll make the update

gleiser-oliveira (approved):

PR #931, opened 2023-01-26
[VEN-325] Replace web3-react with wagmi

therealemjy:
## Jira ticket(s)

VEN-325

## Changes

That's a huge PR, because it refactors the core logic for handling wallet connections, which in turn required changes to all the query and mutation functions. A lot of the changes can however safely be ignored when reviewing this PR, since they're repetitive. None of the original behaviors of the dApp are changed (that will come in the next PRs).
The most important changes to check are the ones made to the web3 client and the auth context, which you can respectively find at `src/clients/web3` and `src/context/AuthContext`.

List of all the changes:
- refactor web3 client
- refactor auth context
- handle wallet connection with `wagmi` instread of `web3-react`
- create `binanceWalletConnector` using PancakeSwap's code (https://github.com/pancakeswap/pancake-frontend)
- replace `web3` library with `ethers` (used by `wagmi`)
- update Typechain config to target ethers v5
- update contract hooks to use new web3 client 
- update query and mutation hooks to use new web3 client and remove unnecessary account address parameters that can be extracted from the contract's signer
- update query and mutation function tests and remove unnecessary tests checking that mutation functions throw errors


src/clients/api/mutations/repay/index.ts
```diff
@@ -1,67 +1,66 @@
...
+    throw new VError({ type: 'unexpected', code: 'somethingWentWrong' });
   }
 
   // Handle repaying full BNB loan
   if (isRepayingFullLoan) {
-    const maximillionContract = getMaximillionContract(web3);
-    const amountWithBuffer = amountWei.multipliedBy(1 + REPAYMENT_BNB_BUFFER_PERCENTAGE);
+    const maximillionContract = getMaximillionContract(signer);
+    const amountWithBufferWei = amountWei.multipliedBy(1 + REPAYMENT_BNB_BUFFER_PERCENTAGE);
+    const accountAddress = await signer!.getAddress();
```
coreyar, line 46: This emits a linter warning - Should we turn it off?

src/clients/api/mutations/supply/index.ts
```diff
@@ -1,48 +1,34 @@
...
 
-const supply = async ({
-  web3,
-  vToken,
-  accountAddress,
-  amountWei,
-}: SupplyInput): Promise<SupplyOutput> => {
+const supply = async ({ signer, vToken, amountWei }: SupplyInput): Promise<SupplyOutput> => {
   // Handle supplying BNB
+  // TODO: check isNative prop of underlying token instead (?)
```
coreyar, line 18: Do we need to do anything about this?

src/clients/api/queries/useGetVaults/useGetVestingVaults/useGetXvsVaultPools.ts
```diff
@@ -33,6 +33,7 @@ const useGetXvsVaultPools = ({
   >[] = [];
 
   // Fetch pool infos
+  // TODO: use multicall
```
coreyar, line 36: Do we need to address this?

src/clients/web3/useMulticall.ts
```diff
@@ -1,11 +1,22 @@
...
+import { useAuth } from 'context/AuthContext';
 
 const useMulticall = () => {
-  const web3 = useWeb3();
-  return useMemo(() => new Multicall({ web3Instance: web3, tryAggregate: true }), [web3]);
+  const { provider } = useAuth();
+  return useMemo(
+    () =>
+      new Multicall({
+        // @ts-expect-error The Provider type used by Multicall only has one
```
coreyar: It makes sense to add that property to the provider so that you don't unintentionally suppress other errors

src/context/AuthContext/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  const login = useCallback(async (connectorId: Connector) => {
+    // If user is attempting to connect their Infinity wallet but the dApp
+    // isn't currently running in the Infinity Wallet app, open it
+    if (connectorId === Connector.InfinityWallet && !isRunningInInfinityWalletApp()) {
+      openInfinityWallet(window.location.href, config.chainId);
+      return;
+    }
+
+    // TODO: determine based on connector ID passed
```
coreyar: Do we need to address this TODO?

therealemjy: No, it should have been removed (the logic to do that is on the next line).

src/clients/api/mutations/repay/index.ts
therealemjy: I'd keep the warning (and not put a comment to suppress it) as these should be used as rarely as possible, and deleted whenever the logic is refactored and the assertion is no longer needed.

src/clients/api/queries/useGetVaults/useGetVestingVaults/useGetXvsVaultPools.ts
therealemjy: Not for now, it's a minor upgrade

src/clients/api/mutations/supply/index.ts
therealemjy: Not right now. I added it since we might have to the day we want to handle multiple chains

src/clients/web3/useMulticall.ts
therealemjy: Not in this situation as it could lead to more errors if we ever use the modified version to send transactions

src/context/AuthContext/index.tsx
coreyar: ahh I see I misunderstood I thought this was an extra property that multicall needed. You can still solve this by upgrading ethers to match the version (5.7.2) used by multicall

src/clients/web3/useMulticall.ts
coreyar: ahh I see I misunderstood I thought this was an extra property that multicall needed. You can still solve this by upgrading ethers to match the version (5.7.2) used by multicall

therealemjy: Good point! I'll make the update

coreyar (approved):

coreyar (approved):

PR #957, opened 2023-02-10
[VEN-1074] Use WalletConnect V2 + upgrade packages (develop)

therealemjy:
## Jira ticket(s)

VEN-1074

## Changes

- upgrade `react-scripts` to `5.0.1` (this is necessary to support the new version of `wagmi`). Note that following this upgrade, `yarn start` throws a large amount of warnings concerning modules without a source map. This is a known issue of `react-scripts` (see [this Stack Overflow issue](https://stackoverflow.com/questions/70599784/failed-to-parse-source-map)) that for now doesn't have any fix
- upgrade `wagmi` to `0.11.5`
- install all minor package upgrades
- use version 2 of WalletConnect connector


gleiser-oliveira (commented):
@therealemjy changes look good, but there are some tests failing.

PR #968, opened 2023-02-28
[VEN-978] Removed wallet connector from local storage and modal

gleiser-oliveira:
## Jira ticket(s)

VEN-978

## Changes

- Removed `LS_KEY_CONNECTED_CONNECTOR` as we are not storing the wallet connector anymore
- Removed `getConnectedConnector` as it depended on `LS_KEY_CONNECTED_CONNECTOR`
- Refactored `useAuth` to comply with the new changes
- Updated account/wallet model to not display. which wallet is being used


src/translation/translations/en.json
```diff
@@ -58,7 +58,7 @@
     },
     "title": {
       "connectWallet": "Connect a wallet",
-      "yourWallet": "Your wallet"
```
gleiser-oliveira, line 61: @therealemjy let me know if you agree on changing this copy

src/context/AuthContext/index.tsx
```diff
@@ -3,30 +3,23 @@ import type { Provider } from '@wagmi/core';
...
 import { AuthModal } from 'components/AuthModal';
-import { LS_KEY_CONNECTED_CONNECTOR } from 'constants/localStorageKeys';
-import getConnectedConnector from 'context/AuthContext/getConnectedConnector';
 import useCopyToClipboard from 'hooks/useCopyToClipboard';
 import { isRunningInInfinityWalletApp } from 'utilities/walletDetection';
 
-export interface Account {
-  address: string;
-  connector?: Connector;
-}
```
gleiser-oliveira, line 20: @therealemjy I also went ahead and refactored this out since the connector is not present anymore. This change leads to most of the other changes in this PR, since I believe it is just simpler for us to reach out the `accountAddress` directly, as it would be the only key in this object. But let me know if you think this is still useful and I can revert this.

therealemjy: Good idea!

therealemjy (approved):

src/translation/translations/en.json
therealemjy: Definitely yeah

PR #971, opened 2023-02-28
[VEN-1134] Remove ID column from History table (master)

therealemjy:
## Jira ticket(s)

VEN-1134

## Changes

- remove ID column from History table


src/clients/api/queries/getTransactions/formatTransaction.ts
```diff
@@ -21,8 +19,6 @@ const formatTransaction = ({
   return {
     ...rest,
     amountWei: convertTokensToWei({ value: new BigNumber(amount), token }),
-    createdAt: new Date(createdAt),
```
coreyar, line 24: I would do this in 2 steps like you mentioned in slack.
First we can update the logic to use the timestamp and fallback to createdAt.
Once the backfill is complete we can remove createdAt and remove the fallback.

therealemjy: I don't think we should fallback to the `createdAt` field, as we will display partially correct data. We technically display wrong data all the time currently, but I'd rather make only one update to the frontend that fixes the issue definitely.

The idea is that this PR would only be deployed once the backend has been updated to backfill all the `null` timestamp fields.

coreyar: Got it. Sounds good

gleiser-oliveira (approved):

PR #975, opened 2023-03-01
[VEN-901] Market graph now uses market graph v2 endpoint

gleiser-oliveira:
## Jira ticket(s)

VEN-901

## Changes

- Adjusted types, hooks and chart component to take the new signature and amount of data returned by the v2 market graph endpoint


src/clients/api/queries/getMarketHistory/index.ts
```diff
@@ -0,0 +1,78 @@
...
+import { MarketSnapshot, VToken } from 'types';
+import { restService } from 'utilities';
+
+const ENTRIES_HOUR = 2;
+export const ENTRIES_DAY = ENTRIES_HOUR * 24;
+const ENTRIES_WEEK = ENTRIES_DAY * 7;
+const ENTRIES_30_DAYS = ENTRIES_DAY * 30;
+const ENTRIES_YEAR = ENTRIES_DAY * 365;
+
+export type MarketHistoryTypes = '1 day' | '1 week' | '1 month' | '1 year';
```
therealemjy: To follow the codebase naming convention:
```suggestion
export type MarketHistoryType = '1 day' | '1 week' | '1 month' | '1 year';
```

src/clients/api/queries/getMarketHistory/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  // @todo Add specific api error handling
+  if ('result' in response && response.result === 'error') {
+    throw new VError({
+      type: 'unexpected',
+      code: 'somethingWentWrong',
+      data: { message: response.message },
+    });
+  }
+
+  let sampleSize = response.data?.data.data.length || 0;
```
therealemjy, line 54: `.data.data.data` 😬
Do you think we could change the API endpoint so it doesn't nest the data returned three times?

src/clients/api/queries/getMarketHistory/index.ts
```diff
@@ -0,0 +1,78 @@
+import { VError } from 'errors';
+import { MarketSnapshot, VToken } from 'types';
+import { restService } from 'utilities';
+
+const ENTRIES_HOUR = 2;
+export const ENTRIES_DAY = ENTRIES_HOUR * 24;
+const ENTRIES_WEEK = ENTRIES_DAY * 7;
+const ENTRIES_30_DAYS = ENTRIES_DAY * 30;
+const ENTRIES_YEAR = ENTRIES_DAY * 365;
```
therealemjy: A small thing, but I'd add `PER_` to each constant name to make it extra clear what they represent. I'd also export all of them, incase they might be used from somewhere else.

```suggestion
export const ENTRIES_PER_HOUR = 2;
export const ENTRIES_PER_DAY = ENTRIES_PER_HOUR * 24;
export const ENTRIES_PER_WEEK = ENTRIES_PER_DAY * 7;
export const ENTRIES_PER_30_DAYS = ENTRIES_PER_DAY * 30;
export const ENTRIES_PER_YEAR = ENTRIES_PER_DAY * 365;
```

What do you think?

src/clients/api/queries/getMarketHistory/index.ts
gleiser-oliveira: Done, thanks for catching that up!

src/clients/api/queries/getMarketHistory/index.ts
gleiser-oliveira: Taken this, it's clearer. I also removed the export for `ENTRIES_PER_DAY`, my first idea was to use it in the chart, but it ended up feeling a bit strange to have this value coming out of the query.

src/clients/api/queries/getMarketHistory/index.ts
coreyar: yeah we can definitely improve this just not in the scope of the initial isolated lending release.

We first need to implement a sustainable versioning pattern and then refactor all our success and error responses to be cleaner and more useful. All of these items are on the roadmap

gleiser-oliveira: Well, that certainly looks redundant, it goes like this:

1st data is a frontend wrapper.
2nd data is how the backend returns all payloads.
3rd data is an actual field for this model in the database.

Number 1 is fine. Number 3 can be renamed to something like `result` instead, but that wouldn't really remove the nesting, it would just look different. Number 2 can be removed, yes, but removing it for this specific endpoint alone wouldn't be that nice. It would be a better idea to refactor how the API as a whole returns data, but that is a larger effort.

therealemjy: I see, thanks for the extra info. Let's keep things as they are for now

src/clients/api/queries/getMarketHistory/index.ts
therealemjy: Makes sense

therealemjy (approved):

PR #979, opened 2023-03-07
[VEN-1132] Market table now has sorting options when in mobile view (master)

gleiser-oliveira:
## Jira ticket(s)

VEN-1132

## Changes

- Mobile viewport now lets the user is sort markets by: total supply, supply APY, liquidity, total borrow and borrow APY.

### DIFF
--- a/src/components/Table/TableCards.tsx
+++ b/src/components/Table/TableCards.tsx
@@ -1,72 +1,132 @@
 /** @jsxImportSource @emotion/react */
 import { Paper, Typography } from '@mui/material';
-import React from 'react';
+import React, { useMemo } from 'react';
 import { Link } from 'react-router-dom';
+import { useTranslation } from 'translation';
 
 import { Delimiter } from '../Delimiter';
+import { Select, Option as SelectOption, SelectProps } from '../Select';
 import { useStyles } from './styles';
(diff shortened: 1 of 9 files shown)


src/components/Table/TableCards.tsx
```diff
@@ -1,72 +1,133 @@
...
+          label: column.label,
+        };
+
+        return [...acc, option];
+      }, [] as SelectOption[]),
+    [columns],
+  );
+
+  console.log(selectOptions);
+  console.log(order);
```
therealemjy: An oversight from coding hehe

therealemjy (changes requested):
It works great on mobile but you'll need to hide the select on higher resolutions:
Screen Shot 2023-03-08 at 11 28 18

src/components/Table/TableCards.tsx
gleiser-oliveira: 🙈

gleiser-oliveira:
> It works great on mobile but you'll need to hide the select on higher resolutions

Indeed, thanks for catching that! I was under the false impression it was hidden, but it was there all the way down the list. Fixed.

therealemjy (approved):

PR #991, opened 2023-03-15
[VEN-1103] Added OKX wallet connector

gleiser-oliveira:
## Jira ticket(s)

VEN-1103

## Changes

- Added OKX (injected connector) to list of supported wallets


therealemjy (approved):
Nice, at first glance I wasn't sure their wallet was injected the same way other wallets are. That's good news!

PR #1003, opened 2023-03-23
[VEN-864] Added XVS snapshot link to governance page

gleiser-oliveira:
## Jira ticket(s)

VEN-864

## Changes

- Added a link on the governance page that leading to the XVS snapshot

### DIFF
--- a/src/pages/Governance/VotingWallet/index.tsx
+++ b/src/pages/Governance/VotingWallet/index.tsx
@@ -1,7 +1,15 @@
 /** @jsxImportSource @emotion/react */
 import { Paper, Typography } from '@mui/material';
 import BigNumber from 'bignumber.js';
-import { Delimiter, InfoIcon, LinkButton, PrimaryButton, TokenIcon } from 'components';
+import {
+  Delimiter,
+  Icon,
+  InfoIcon,
+  LinkButton,
+  PrimaryButton,
+  SecondaryAnchorButton,
+  TokenIcon,
+} from 'components';
 import React, { useMemo, useState } from 'react';
 import { Link } from 'react-router-dom';
 import { useTranslation } from 'translation';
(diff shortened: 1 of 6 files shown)


src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -22,6 +30,8 @@ import DelegateModal from './DelegateModal';
 import { useStyles } from './styles';
 import TEST_IDS from './testIds';
 
+const XVS_SNAPSHOT_URL = 'https://snapshot.org/#/venus-xvs.eth';
```
therealemjy: Let's add this to a constant file inside `src/constants`.

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -174,6 +184,13 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
         </Typography>
       </Paper>
 
+      <a href={XVS_SNAPSHOT_URL} target="_blank" rel="noreferrer">
+        <SecondaryButton css={styles.snapshotButton}>
+          <Icon css={styles.snapshotButtonIcon} name="lightening" />
+          {t('vote.goToXvsSnapshot')}
+        </SecondaryButton>
+      </a>
```
therealemjy: We have a variant for buttons that are links: `AnchorButton` (`src/components/Button/index.tsx`). However, I've spotted two issues with it:
- right now `AnchorButton` renders a button with the `primary` variant, in your case here you need the `secondary` variant. So I'd rename `AnchorButton` to `PrimaryAnchorButton`, then create a `SecondaryAnchorButton` component that renders with the `secondary` variant and that you can use here
- I recently updated how anchors are rendered within the dApp and missed this one: when being hovered over, the `AnchorButton` displays an underline on its content. We need to update this so the underline isn't displayed

therealemjy (approved):

PR #1004, opened 2023-03-24
[VEN-1096] VAI mint and repay updates

gleiser-oliveira:
## Jira ticket(s)

VEN-1096

## Changes

- Added VAI wallet balance underneath the input
- Mint VAI now displays current APY
- Displaying borrow balance plus interests using `getVAIRepayAmount`
- Sending MAXUINT256 to fully pay the VAI borrow balance


src/__mocks__/contracts/multicall.ts
gleiser-oliveira, line 1: @therealemjy what is the right way to generate multicall mocks? I got this from logging the real response while testing, but I guess there is a better way.

src/pages/Vai/styles.ts
```diff
@@ -14,7 +14,7 @@ export const useStyles = () => {
       flex-direction: column;
     `,
     tabs: css`
-      max-width: ${theme.spacing(136)};
+      max-width: ${theme.spacing(140)};
```
gleiser-oliveira, line 17: @therealemjy this change seems random, but it comes from adding the new tooltip, I found that it looks better with a bit more width, let me know if that is fine.

src/__mocks__/contracts/multicall.ts
therealemjy: That's how I'm doing it as well. I think that's an acceptable way, as this way the mocked request represents realistic data.

src/pages/Vai/styles.ts
therealemjy: Ok sounds good, thanks for adding context

src/clients/api/index.ts
```diff
@@ -121,6 +121,10 @@ export { default as getVaiCalculateRepayAmount } from './queries/getVaiCalculate
 export * from './queries/getVaiCalculateRepayAmount';
 export { default as useGetVaiCalculateRepayAmount } from './queries/getVaiCalculateRepayAmount/useGetVaiCalculateRepayAmount';
 
+export { default as getVaiRepayAmountWithInterests } from './queries/getVaiRepayAmountWithInterests';
+export * from './queries/getVaiCalculateRepayAmount';
```
therealemjy: I think you meant to export `getVaiRepayAmountWithInterests`:
```suggestion
export * from './queries/getVaiRepayAmountWithInterests';
```

src/pages/Vai/textFieldWithBalance/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+            White: <span css={styles.whiteLabel} />,
+          }}
+          values={{
+            balance: formatTokensToReadableValue({
+              value: userBalanceWei,
+              token: TOKENS.vai,
+            }),
+          }}
+        />
+      </Typography>
```
therealemjy: You can use the `description` prop of `FormikTokenTextField` to display the wallet balance here. It also needs to be converted into tokens before being displayed (for this you can use the `convertWeiToTokens` util instead of `formatTokensToReadableValue `:
```suggestion
      <FormikTokenTextField
        name="amount"
        css={styles.textField}
        token={TOKENS.vai}
        max={maxValue}
        disabled={disabled}
        rightMaxButton={{
          label: maxButtonLabel,
          valueOnClick: maxValue,
        }}
        data-testid={TEST_IDS.repayTextField}
        description={
          <Trans
            i18nKey="borrowRepayModal.repay.walletBalance"
            components={{
              White: <span css={styles.whiteLabel} />,
            }}
            values={{
              balance: convertWeiToTokens({
                valueWei: userBalanceWei,
                token: TOKENS.vai,
                returnInReadableFormat: true,
                addSymbol: true,
              }),
            }}
          />
        }
      />
```

src/pages/Vai/RepayVai/index.tsx
```diff
@@ -60,7 +73,7 @@ export const RepayVaiUi: React.FC<IRepayVaiUiProps> = ({
 
   // Convert minted wei into VAI
```
therealemjy: It's a small thing but we should update this comment since this value no longer represents the minted VAI only

src/pages/Vai/RepayVai/index.tsx
```diff
@@ -99,32 +112,34 @@ export const RepayVaiUi: React.FC<IRepayVaiUiProps> = ({
...
+                  userBalanceWei={userBalanceWei}
+                  maxButtonLabel={t('vai.repayVai.rightMaxButtonLabel')}
+                />
 
+                <div css={styles.ctaContainer}>
                   <LabeledInlineContent
                     css={styles.getRow({ isLast: false })}
                     iconSrc={TOKENS.vai}
                     label={t('vai.repayVai.repayVaiBalance')}
+                    tooltip={t('vai.repayVai.repayFeeTooltip')}
```
therealemjy: We use this tooltip for the repay fee already, so we should use a different content here (perhaps "Total VAI balance to repay, including interests").

src/pages/Vai/RepayVai/index.tsx
```diff
@@ -99,32 +112,34 @@ export const RepayVaiUi: React.FC<IRepayVaiUiProps> = ({
...
                 </div>
 
+                {isPayingFullRepayBalance(values.amount, repayBalanceWei) && (
+                  <NoticeWarning description={<Trans i18nKey="vai.repayVai.fullRepayWarning" />} />
+                )}
+
                 <FormikSubmitButton
+                  css={styles.submitButton({
+                    isShowingWarning: isPayingFullRepayBalance(values.amount, repayBalanceWei),
+                  })}
```
therealemjy: Instead of conditionally applying a top margin on the submit button, you could always apply a bottom margin on the `NoticeWarning` instance above:

```suggestion
                {isPayingFullRepayBalance(values.amount, repayBalanceWei) && (
                  <NoticeWarning description={<Trans i18nKey="vai.repayVai.fullRepayWarning" />} css={styles.noticeWarning} />
                )}

                <FormikSubmitButton
```

Where `styles.noticeWarning` applies the bottom margin.

src/pages/Vai/RepayVai/index.tsx
```diff
@@ -33,19 +39,26 @@ export interface IRepayVaiUiProps {
...
   repayVai: (amountWei: BigNumber) => Promise<ContractReceipt | undefined>;
   userBalanceWei?: BigNumber;
+  repayBalanceWei?: BigNumber;
   userMintedWei?: BigNumber;
 }
 
+const isPayingFullRepayBalance = (amount: string, repayBalance: BigNumber | undefined): boolean => {
+  const amountWei = convertTokensToWei({ value: new BigNumber(amount), token: TOKENS.vai });
+  return repayBalance ? amountWei === repayBalance : false;
+};
```
therealemjy: - `new BigNumber(amount)` will return the string `NaN` when amount is an empty string (which is the case when the textfield is empty)
- I'd rename `repayBalance` to `repayBalanceWei` to follow the codebase pattern of adding the unit to variable names
- since you're comparing two `BigNumber`s, you can use the `isEqualTo` method

src/pages/Vai/textFieldWithBalance/index.tsx
therealemjy, line 1: The name of the folder this component is contained in should be capitalized: TextFieldWithBalance

src/components/Notice/styles.ts
```diff
@@ -8,11 +8,11 @@ export const useStyles = () => {
   return {
     root: css`
       border-radius: ${theme.shape.borderRadius.medium}px;
-      background-color: ${theme.palette.background.default};
     `,
     inner: css`
       border-radius: ${theme.shape.borderRadius.medium}px;
       border: 1px solid ${theme.palette.secondary.light};
+      background-color: ${theme.palette.background.default};
```
gleiser-oliveira, line 15: Changed this to support the margin

src/pages/Vai/RepayVai/index.tsx
gleiser-oliveira: At first I wanted to avoid this because I needed one extra change which could lead to other places using `NoticeWarning` to regress https://github.com/VenusProtocol/venus-protocol-interface/pull/1004#discussion_r1149567513

But now after considering and reviewing, it seems good

therealemjy (approved):

PR #1005, opened 2023-03-24
[VEN-1165] Fetch isolated pools directly from contracts

therealemjy:
## Jira ticket(s)

VEN-1165

## Changes

- update `getIsolatedPools` query function to fetch as much data as possible from contracts instead of the subgraph. Now only the `supplierCount` and `borrowerCount` fields of each asset are fetched from the subgraph as these can't be taken from contracts
- add `areAddressesEqual` utility function
- remove local record of PoolLens ABI in favor of ABI from `@venusprotocol/isolated-pools` package
- rename `formatToPool` util to `addUserPropsToPool`
- add `getIsolatedPoolParticipantsCount` subgraph query
- remove `getIsolatedPools` subgraph query
- update `PoolLens` contract testnet address
- add `PoolRegistry` contract testnet address
- update testnet tokens and vTokens
- fix maximum input value of supply, withdraw and borrow forms to prevent it from going below 0


gleiser-oliveira (approved):
This looks good, we can solve the places with console.error in https://github.com/VenusProtocol/venus-protocol-interface/pull/1010

PR #1008, opened 2023-03-27
[VEN-870] Updated high risk borrow warning

gleiser-oliveira:
## Jira ticket(s)

VEN-870

## Changes

- The high risk borrow warning has been updated to match the current designs

### DIFF
--- a/src/hooks/useBorrowRepayModal/Modal/Borrow/index.tsx
+++ b/src/hooks/useBorrowRepayModal/Modal/Borrow/index.tsx
@@ -1,14 +1,6 @@
 /** @jsxImportSource @emotion/react */
 import BigNumber from 'bignumber.js';
-import {
-  AccountData,
-  ConnectWallet,
-  EnableToken,
-  FormikSubmitButton,
-  FormikTokenTextField,
-  IsolatedAssetWarning,
-  Spinner,
-} from 'components';
+import { ConnectWallet, EnableToken, Spinner } from 'components';
 import { ContractReceipt } from 'ethers';
 import React, { useMemo } from 'react';
 import { useTranslation } from 'translation';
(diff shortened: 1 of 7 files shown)


src/hooks/useBorrowRepayModal/Modal/Borrow/index.tsx
```diff
@@ -87,67 +87,81 @@ export const BorrowForm: React.FC<BorrowFormProps> = ({
...
-          {pool.isIsolated && (
-            <IsolatedAssetWarning
+      {({ values, dirty, isValid, errors }) => {
+        const isHighRiskBorrow =
+          new BigNumber(values.amount).isGreaterThanOrEqualTo(safeLimitTokens) &&
+          new BigNumber(values.amount).isLessThanOrEqualTo(limitTokens);
+
+        const enableLabel = isHighRiskBorrow
+          ? t('borrowRepayModal.borrow.submitButtonHighRisk')
+          : t('borrowRepayModal.borrow.submitButton');
```
therealemjy: I'm generally not a fan of declaring variables inside the return statement of a component as I feel it obfuscates the logic. An alternative could be to use a memoized callback:
```tsx
const isHighRiskBorrow = useCallback(
  (amount: string) =>
    new BigNumber(amount).isGreaterThanOrEqualTo(safeLimitTokens) &&
    new BigNumber(amount).isLessThanOrEqualTo(limitTokens),
  [limitTokens, safeLimitTokens],
);
```

Then you can use it inside the return statement.
The two advantages I see are that we keep the logic out of the return statement and that these constants don't get redeclared with every render. Let me know your thoughts though in case you prefer the approach you took originally!

gleiser-oliveira: I created a new component and moved all the related logic to it, this way we avoid having this variable inside the return statement and `BorrowForm` is cleaner now, let me know if you'd like additional changes 👍🏻

therealemjy (approved):
Nice, that's clean 👍

PR #1010, opened 2023-03-28
[VEN-1066][VEN-1130] Add analytics and error logger

therealemjy:
## Jira ticket(s)

VEN-1066
VEN-1130

## Changes

- add analytics through PostHog
- add error logger through Sentry


gleiser-oliveira (commented):
I think these console.error calls can be removed now

https://github.com/VenusProtocol/venus-protocol-interface/blob/feature/posthog-sentry/src/clients/api/queries/getIsolatedPools/index.ts#L48

https://github.com/VenusProtocol/venus-protocol-interface/blob/feature/posthog-sentry/src/clients/api/queries/getIsolatedPools/formatOutput/formatToDistributions.ts#L18

therealemjy:
For now I think we should keep them in until Sentry is integrated to the dApp (https://github.com/VenusProtocol/venus-protocol-interface/pull/1010) so they can be replaced with sending an event to Sentry. What do you think?

gleiser-oliveira (approved):

PR #1014, opened 2023-04-07
[VEN-1272] Integrate swap feature to repay form

therealemjy:
## Jira ticket(s)

VEN-1272

## Main changes

- update `RepayForm` component inside `useBorrowRepayModal` hook to handle swap and repay flow. **Enabling the token to swap from and sending the transaction will be handled in the next PRs**
- update how libraries are initialized on launch of the dApp so that initialization is split by library and contained in a separate `initializeLibraries` function
- add mock for the config so that it can be tested too and isn't affected by the local environment variables
- create `isFeatureEnabled` utility function and use it everywhere in the dApp to make it easier to test feature flags

## Miscellaneous changes
 - move mocks related to Swap to reusable files
- create mock for `IsolatedPoolParticipantsCountQuery` subgraph query (so far the tests attempted to make a real request to the subgraph node)
- update `AccountData` to accept a `swap` prop that is then used to determine the values displayed
- add `active` prop and style to `Button` component
- update `rightMaxButton` prop of `TokenTextField` component to represent all the props of a `Button` component bar `variant`, `children`, and `small`
- add `isLoading` prop to the values returned by `useGetSwapInfo`
- add custom `positive` and `lowerThanOrEqualTo` rules to Yup
- add `maxAmount` prop to Mint VAI and Repay VAI tabs
- add `options` parameter to `useGetSwapTokenUserBalances` hook


src/hooks/useBorrowRepayModal/Modal/RepayForm/index.tsx
```diff
@@ -168,11 +261,120 @@ export interface RepayFormProps {
...
+    isRepayingFullLoan,
+  }: {
+    swap: Swap;
+    isRepayingFullLoan: boolean;
+  }) => {
+    // TODO: call relevant mutation function (see: VEN-1270)
+    console.log('TODO: execute swap and repay', {
+      swap,
+      isRepayingFullLoan,
+    });
```
gleiser-oliveira, line 285: nit: although this log is not needed, removing it will leave the arguments as unused

gleiser-oliveira (approved):

gleiser-oliveira:
@therealemjy just an issue with the storybook test

src/hooks/useBorrowRepayModal/Modal/RepayForm/index.tsx
gleiser-oliveira: Or even better, we can use the swap which is on testnet

therealemjy: Yeah. I can remove it and add an exception however

PR #1016, opened 2023-04-09
Remove duplicated % sign

therealemjy:
## Changes

- remove duplicated percentage sign in tooltip. Because the value passed to the tooltip already contains a percentage sign, the one present in the tooltip results in two % signs being displayed in the same sentence

### DIFF
--- a/src/translation/translations/en.json
+++ b/src/translation/translations/en.json
@@ -420,7 +420,7 @@
   },
   "myAccount": {
     "safeLimit": "Your safe limit:",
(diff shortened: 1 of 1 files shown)


therealemjy:
The tests will pass once https://github.com/VenusProtocol/venus-protocol-interface/pull/1017 merged

gleiser-oliveira (approved):
@therealemjy this looks good, I guess #1017 is going to be reopened?

therealemjy:
@gleiser-oliveira Thanks for the review.

Sort of. I ended up having to refactor the Repay form so it uses a React state (I ran into some limitations using Formik), so I'll open a PR that contains all the last changes needed for this task.

PR #1021, opened 2023-04-17
[VEN-1271] Add support for Swap and Supply flow (UI)

therealemjy:
## Jira ticket(s)

VEN-1271

## Changes

- refactor Supply form following the same file architecture as the Repay form
- integrate swap to Supply form

The last step is to connect the submit handler of the Supply form to the `SwapRouter` contract, which will be done in a next PR.


src/hooks/useSupplyWithdrawModal/Modal/SupplyForm/SubmitSection/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    if (formError === 'HIGHER_THAN_SUPPLY_CAP') {
+      return t('supplyWithdrawModal.supply.submitButtonLabel.amountHigherThanSupplyCap');
+    }
+
+    if (formError === 'HIGHER_THAN_WALLET_BALANCE') {
+      return t('supplyWithdrawModal.supply.submitButtonLabel.insufficientWalletBalance', {
+        tokenSymbol: fromToken.symbol,
+      });
+    }
```
gleiser-oliveira, line 62: nit: I feel this is better expressed as a `switch(formError) case...`

gleiser-oliveira (approved):
Just a minor comment on the formError if blocks, please feel free to update or to keep them, PR looks good!

src/hooks/useSupplyWithdrawModal/Modal/SupplyForm/SubmitSection/index.tsx
therealemjy: The issue is that `formError` isn't the only variables used to determine what to return here: some of the IF statements use `isSwapLoading`, `fromTokenAmountTokens` and `isFormValid`. That's why I'd keep the IF statements, otherwise in order to use a switch we'll have to make a separate function

PR #1039, opened 2023-05-09
[VEN-1434] Updating dApp with the latest IL deployment

gleiser-oliveira:
## Jira ticket(s)

VEN-1434

## Changes

- Added vTokens and corresponding addresses, logos and underlying tokens from testnet


src/constants/tokens/swap/testnet.ts
```diff
@@ -1,14 +1,5 @@
...
 import { TESTNET_TOKENS } from '../common/testnet';
 
 export const TESTNET_SWAP_TOKENS = {
-  wbnb: {
-    symbol: 'WBNB',
-    decimals: 18,
-    address: '0xae13d989daC2f0dEbFf460aC112a837C89BAa7cd',
-    asset:
-      'https://pancakeswap.finance/images/tokens/0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c.svg',
-  } as Token,
```
gleiser-oliveira, line 12: @chechu @therealemjy let's review this change, I moved wbnb to TESTNET_TOKENS

src/constants/tokens/vBep/testnet.ts
```diff
@@ -228,4 +228,191 @@ export const TESTNET_VBEP_TOKENS = {
...
+    symbol: 'vBIFI',
+    asset: vXvs, // TODO: add correct asset
+    underlyingToken: TESTNET_TOKENS.bifi,
+  } as VToken,
+  // Pool 2
+  '0x6d258734b003e803c1f67ccaa2ccddab93a60067': {
+    address: '0x6d258734b003e803c1f67ccaa2ccddab93a60067',
+    decimals: 8,
+    symbol: 'vUSDD',
+    asset: vXvs, // TODO: add correct asset
```
gleiser-oliveira: Do we still need vToken logos?

src/constants/tokens/swap/mainnet.ts
```diff
@@ -1,5 +1,2354 @@
+import { Token } from 'types';
+
 import { MAINNET_TOKENS } from '../common/mainnet';
-import { MAINNET_PANCAKE_SWAP_TOKENS } from './mainnetPancakeSwapTokens';
+
+const MAINNET_PANCAKE_SWAP_TOKENS = {
```
gleiser-oliveira: We shouldn't need to do this, the script was failing in CI

Dockerfile
```diff
@@ -19,6 +19,8 @@ COPY src/constants/contracts/abis ./src/constants/contracts/abis
 COPY src/clients/subgraph ./src/clients/subgraph
 COPY patches ./patches
 
+COPY venusprotocol-isolated-pools-0.1.0-local.1.tgz ./
```
therealemjy: Is this a temporary solution until the package gets updated?

src/utilities/convertDollarsToCents.ts
```diff
@@ -1,6 +1,14 @@
...
+const convertDollarsToCents = (
+  value: number | BigNumber,
+  { whole }: { whole: boolean } = { whole: true },
+) => {
+  const cents = new BigNumber(value).times(100);
+  if (whole) {
+    return cents.dp(0).toNumber();
+  }
+  return cents.toNumber();
+};
```
therealemjy: I'd remove the option to return rounded cents and always return fractional cents to keep things simple. The dApp will still only display 2 decimals for dollar values since `formatCentsToReadableValue` takes care of that.

src/utilities/getContractAddress.ts
```diff
@@ -1,8 +1,13 @@
...
 import config from 'config';
 
 import mainContractAddresses from 'constants/contracts/addresses/main.json';
 
 const getContractAddress = (contractId: keyof typeof mainContractAddresses) =>
   mainContractAddresses[contractId][config.chainId];
 
+export const getIsolatedPoolAddress = (
+  contractName: keyof typeof isolatedLendingTestnetDeployments.contracts,
+) => (config.isOnTestnet ? isolatedLendingTestnetDeployments.contracts[contractName].address : '');
```
therealemjy: Rather than using a separate util, how about keeping the existing logic and making a list of all the contracts (the ones we keep a reference of locally merged with the ones taken from the package)?

Something like:
```typescript
const isolatedLendingContractAddresses = Object.entries(
  isolatedLendingTestnetDeployments.contracts,
).map(([contractName, contractInfo]) => ({
  [contractName]: {
    97: contractInfo.address,
    56: '', // TODO: fill up once contracts have been deployed to mainnet
  },
}));

const contractAddress = {
  ...mainContractAddresses,
  ...isolatedLendingContractAddresses
};
```

This way `getContractAddress` can stay the same and the logic for retrieving contract addresses within the dApp can remain as it is.

src/constants/tokens/vBep/testnet.ts
therealemjy: We don't indeed, since they aren't used anywhere. I've created a ticket to get them removed: https://jira.toolsfdg.net/browse/VEN-1511

src/constants/tokens/swap/testnet.ts
therealemjy: Since we now have a pool that lists WBNB on testnet, this change makes sense to me

src/constants/tokens/swap/mainnet.ts
therealemjy: What was the error thrown by the process?

I'd like to keep the existing logic which generates `mainnetPancakeSwapTokens.ts` using PancakeSwap's list contained in their repo, because this way we will always be up to date with them (the list changes frequently).

Dockerfile
gleiser-oliveira: Correct, we need to release a new one

src/constants/tokens/swap/mainnet.ts
gleiser-oliveira: Indeed, I did this to unblock the deployment faster, but the script should be the right thing to go for. Here's the error, we can revisit it today:

```
ts-node --esm ./scripts/generatePancakeSwapTokens.mts && yarn prettier --write ./src/constants/tokens/swap/mainnetPancakeSwapTokens.ts
[91mReferenceError: exports is not defined in ES module scope
    at file:///usr/app/scripts/generatePancakeSwapTokens.mts:28:23
    at ModuleJob.run (node:internal/modules/esm/module_job:197:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:337:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)
[0m[91merror Command failed with exit code 1.
```

therealemjy: Ok I see. I'll have a look at this on my side

therealemjy: I found a fix for this: https://github.com/VenusProtocol/venus-protocol-interface/pull/1046
I went with the easy route and transformed the script into a plain JavaScript module. Since it's a small script, I think we're good to go with that for now

therealemjy:
All comments have been addressed, I'd say we're ready to merge this one into `develop`. @gleiser-oliveira What do you think?

therealemjy (approved):

gleiser-oliveira:
> All comments have been addressed, I'd say we're ready to merge this one into `develop`. @gleiser-oliveira What do you think?

@therealemjy  let's get this one merged https://github.com/VenusProtocol/venus-protocol-interface/pull/1072 and it will be good to go!

therealemjy:
Sounds good to me 👌

PR #1050, opened 2023-05-26
[VEN-1514] Added support for account authentication (develop)

gleiser-oliveira:
## Jira ticket(s)

VEN-1514

## Changes

- Added logic to support authenticating an account on the dApp


src/clients/api/index.ts
```diff
@@ -164,6 +164,10 @@ export { default as getPendingRewards } from './queries/getPendingRewards';
 export * from './queries/getPendingRewards';
 export { default as useGetPendingRewards } from './queries/getPendingRewards/useGetPendingRewards';
 
+export { default as getAuthentication } from './queries/getAuthentication';
+export * from './queries/getAllowance';
```
therealemjy:
```suggestion
export * from './queries/getAuthentication';
```

src/clients/api/queries/getAuthentication/index.ts
```diff
@@ -0,0 +1,17 @@
+import { restService } from 'utilities';
+
+const getAuthentication = async (accountAddress: string | undefined): Promise<boolean> => {
```
therealemjy: To follow how other query functions define their arguments, I'd suggest using an object. I also think `accountAddress` should never be `undefined`, as otherwise the logic says that an `undefined` address is not authorized which is technically incorrect. Instead, we can just disable the hook when the address passed is `undefined` (more on that later).
```suggestion
const getAuthentication = async ({ accountAddress }: { accountAddress: string }): Promise<boolean> => {
```

therealemjy: Just for the sake of making the name extra clear, I'd got with something like: `getIsAddressAuthorized`. What do you think?

src/clients/api/queries/getAuthentication/index.ts
```diff
@@ -0,0 +1,17 @@
...
+    const response = await restService({
+      endpoint: `/authentication/${accountAddress}`,
+      method: 'GET',
+    });
+    const authorized = response.status === 200;
+
+    return authorized;
+  }
+
+  return false;
```
therealemjy: Also to follow how other query functions are written, I'd return an object:
```suggestion
  if (accountAddress) {
    const response = await restService({
      endpoint: `/authentication/${accountAddress}`,
      method: 'GET',
    });
    const authorized = response.status === 200;

    return {
      authorized,
    };
  }

  return {
    authorized: false,
  };
```

src/clients/api/__mocks__/index.ts
```diff
@@ -12,6 +12,10 @@ import FunctionKey from 'constants/functionKey';
 import { GetBalanceOfInput } from '../queries/getBalanceOf';
 
 // Queries
+export const getAuthentication = jest.fn(accountAddress => fakeAddress !== accountAddress);
+export const useGetAuthentication = (accountAddress: string | undefined) =>
+  useQuery(FunctionKey.GET_IS_ADDRESS_AUTHORIZED, () => getAuthentication(accountAddress));
+
```
therealemjy:
```suggestion
export const getIsAddressAuthorized = jest.fn(accountAddress => fakeAddress !== accountAddress);
export const useGetIsAddressAuthorized = (accountAddress: string | undefined) =>
  useQuery(FunctionKey.GET_IS_ADDRESS_AUTHORIZED, () => getIsAddressAuthorized(accountAddress));

```

therealemjy (approved):
Looks good, just a small variable name change in the mock needed

src/clients/api/queries/getIsAddressAuthorized/index.ts
```diff
@@ -0,0 +1,29 @@
+import { restService } from 'utilities';
+
+export type GetAuthenticationOutput = {
```
therealemjy:
```suggestion
export type GetIsAddressAuthorizedOutput = {
```

therealemjy (approved):

PR #1060, opened 2023-05-30
Added preview env var and API URL

gleiser-oliveira:
## Changes

- Checks if the app is running in the preview environment and uses the corresponding API URL

### DIFF
--- a/src/constants/endpoints.ts
+++ b/src/constants/endpoints.ts
@@ -1,8 +1,9 @@
-import { BscChainId } from 'types';
+import { BscChainId, Environment } from 'types';
 
-export const API_ENDPOINT_URLS = {
-  [BscChainId.MAINNET]: 'https://api.venus.io/api',
-  [BscChainId.TESTNET]: 'https://testnetapi.venus.io/api',
+export const API_ENDPOINT_URLS: Record<Environment, string> = {
+  mainnet: 'https://api.venus.io/api',
+  preview: 'https://api-preview.venus.io/api',
+  testnet: 'https://testnetapi.venus.io/api',
 };
 
 export const RPC_URLS: {
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -6,7 +6,7 @@ export enum BscChainId {
   'TESTNET' = 97,
 }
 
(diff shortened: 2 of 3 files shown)


src/constants/endpoints.ts
```diff
@@ -5,6 +5,12 @@ export const API_ENDPOINT_URLS = {
   [BscChainId.TESTNET]: 'https://testnetapi.venus.io/api',
 };
 
+export const getApiEndpointUrl = (chainId: BscChainId, isOnPreview = false) => {
+  if (isOnPreview) return 'https://api-preview.venus.io/api';
+
+  return API_ENDPOINT_URLS[chainId];
+};
```
gleiser-oliveira: Can be either this, or changing the `API_ENDPOINT_URLS` to use the new env var as the key, instead of the `chainId`, since preview and mainnet have the same `chainId`

coreyar: I would refactor the `API_ENDPOINT_URLS` because they will need to get refactored if we launch on other networks

src/types/index.ts
```diff
@@ -6,6 +6,8 @@ export enum BscChainId {
   'TESTNET' = 97,
 }
 
+export type Environments = 'testnet' | 'preview' | 'mainnet';
```
therealemjy: Don't we have `staging` and `production` as well? I'm referring to Corey's message in the frontend channel.
Nit: it should be singularized: `Environment`

src/constants/endpoints.ts
therealemjy: I second Corey. If we launch on other networks, we'll very likely provide a chain ID to the API endpoint so it knows what data to load but its URL would be the same for each environment.

gleiser-oliveira: Done, thanks guys!

src/types/index.ts
gleiser-oliveira: In my opinion it's just confusing to add more than the 3 we use on the frontend right now. I'd update as needed.

Renamed to `Environment`

therealemjy (approved):

PR #1067, opened 2023-06-02
[HOTFIX] Wallet connection status (develop)

therealemjy:
## Changes

- update rest service so status is returned correctly
- prevent `useGetIsAddressAuthorized` hook from ever refetching after fetching successfully once
- reverse the check, to consider valid more cases, including cache responses
- add `REACT_APP_ENVIRONMENT` variable to .env.example file


src/clients/api/queries/getIsAddressAuthorized/useGetIsAddressAuthorized.ts
```diff
@@ -19,7 +18,10 @@ const useGetIsAddressAuthorized = (accountAddress: string, options?: Options) =>
     FunctionKey.GET_IS_ADDRESS_AUTHORIZED,
     () => getIsAddressAuthorized({ accountAddress }),
     {
-      refetchInterval: DEFAULT_REFETCH_INTERVAL_MS,
+      refetchOnMount: false,
+      refetchOnReconnect: false,
+      refetchOnWindowFocus: false,
+      staleTime: Infinity,
```
gleiser-oliveira: I'd say this is too permissive. It can be the case that DEFAULT_REFETCH_INTERVAL_MS refetches too frequently and we don't want it, but I think we need to check it from time to time.

therealemjy: Agreed, Jesús also thinks the same. Perhaps once every hour then?

gleiser-oliveira: That would work!

gleiser-oliveira (approved):

PR #1078, opened 2023-06-14
fix: environment is set by REACT_APP_ENVIRONMENT

gleiser-oliveira:
## Changes

- Environment should be set by REACT_APP_ENVIRONMENT env var exclusively.

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -2,7 +2,6 @@ import sample from 'lodash/sample';
 import { BscChainId, Environment } from 'types';
 
 import { BSC_SCAN_URLS } from 'constants/bsc';
-import { DAPP_HOSTS } from 'constants/dAppHosts';
 import { API_ENDPOINT_URLS, RPC_URLS } from 'constants/endpoints';
 
 export interface Config {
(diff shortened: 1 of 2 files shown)


therealemjy:
Ah cool you did the changes already. We can remove the `dappHosts.ts` file too then since it was only used for that.

src/config/index.ts
```diff
@@ -24,17 +23,8 @@ export interface Config {
...
-  environment = 'mock';
-} else if (DAPP_HOSTS.testnet.includes(window.location.host)) {
-  environment = 'testnet';
-} else if (DAPP_HOSTS.preview.includes(window.location.host)) {
-  environment = 'preview';
-} else if (DAPP_HOSTS.mainnet === window.location.host) {
-  environment = 'mainnet';
-}
+const environment: Environment =
+  (process.env.REACT_APP_ENVIRONMENT as Environment | undefined) || 'testnet';
```
therealemjy: Just for safety, I'd default to `mainnet`:
```suggestion
  (process.env.REACT_APP_ENVIRONMENT as Environment | undefined) || 'mainnet';
```

therealemjy (approved):

PR #1082, opened 2023-06-14
Fix initial redirection to Account page

therealemjy:
## Changes

- update auth context to expose `status`
- fix logic to determine whether a user should be redirected to the Account page on launch

### DIFF
--- a/src/context/AuthContext/index.tsx
+++ b/src/context/AuthContext/index.tsx
@@ -27,7 +27,7 @@ export interface AuthContextValue {
   openAuthModal: () => void;
   closeAuthModal: () => void;
   provider: Provider;
-  isReconnecting: boolean;
+  status: 'disconnected' | 'connecting' | 'connected' | 'reconnecting';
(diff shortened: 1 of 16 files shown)


src/context/AuthContext/index.tsx
```diff
@@ -49,6 +49,9 @@ export const AuthProvider: React.FC = ({ children }) => {
   const { data: signer } = useSigner();
   const provider = useProvider();
   const { address, status } = useAccount();
+
+  console.log(status, address);
```
gleiser-oliveira: Log can be removed

gleiser-oliveira: Pushed a fix

gleiser-oliveira (approved):

gleiser-oliveira (approved):

src/context/AuthContext/index.tsx
therealemjy: Nice, thank you!

PR #1092, opened 2023-06-15
[VEN-1620] Add WBETH Token/VToken

gleiser-oliveira:
## Jira ticket(s)

VEN-1620

## Changes

- Add WBETH token and VToken addresses for testnet and mainnet


therealemjy (approved):

therealemjy:
Looks good to me, just some minor linting issues.

One thing worth noting: we need to make sure not to deploy any update to prod before tomorrow, since that means we'd deploy WBETH to mainnet.

PR #1131, opened 2023-06-29
[VEN-1638][VEN-1166][VEN-923] Update isolated lending addresses

therealemjy:
## Jira ticket(s)

VEN-1638
VEN-1166
VEN-923

## Changes

- update `@venusprotocol/isolated-pools` package
- add missing mainnet isolated tokens
- add mainnet isolated vTokens
- add mainnet swap router address of isolated pool

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@testing-library/react": "^12.1.4",
     "@uiw/react-markdown-preview": "^4.1.13",
     "@uiw/react-md-editor": "^3.14.3",
-    "@venusprotocol/isolated-pools": "1.0.0-dev.18",
+    "@venusprotocol/isolated-pools": "^1.0.0",
     "bignumber.js": "^9.0.0",
     "copy-to-clipboard": "^3.3.1",
     "date-fns": "^2.28.0",
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@testing-library/react": "^12.1.4",
     "@uiw/react-markdown-preview": "^4.1.13",
     "@uiw/react-md-editor": "^3.14.3",
-    "@venusprotocol/isolated-pools": "1.0.0-dev.13",
+    "@venusprotocol/isolated-pools": "1.0.0-dev.18",
     "bignumber.js": "^9.0.0",
     "copy-to-clipboard": "^3.3.1",
     "date-fns": "^2.28.0",
--- a/src/clients/api/queries/getIsolatedPools/index.ts
+++ b/src/clients/api/queries/getIsolatedPools/index.ts
@@ -15,7 +15,7 @@ import { GetIsolatedPoolsInput, GetIsolatedPoolsOutput } from './types';
 
 export type { GetIsolatedPoolsInput, GetIsolatedPoolsOutput } from './types';
 
-const POOL_REGISTRY_ADDRESS = getContractAddress('PoolRegistry');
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -45,7 +45,7 @@
     "@testing-library/react": "^12.1.4",
     "@uiw/react-markdown-preview": "^4.1.13",
     "@uiw/react-md-editor": "^3.14.3",
-    "@venusprotocol/isolated-pools": "1.0.0-dev.13",
+    "@venusprotocol/isolated-pools": "1.0.0-dev.18",
```
gleiser-oliveira: Should we use the stable package instead? https://www.npmjs.com/package/@venusprotocol/isolated-pools/v/1.0.0

src/clients/api/queries/getIsolatedPools/index.ts
```diff
@@ -15,7 +15,7 @@ import { GetIsolatedPoolsInput, GetIsolatedPoolsOutput } from './types';
 
 export type { GetIsolatedPoolsInput, GetIsolatedPoolsOutput } from './types';
 
-const POOL_REGISTRY_ADDRESS = getContractAddress('PoolRegistry');
+const POOL_REGISTRY_ADDRESS = getContractAddress('PoolRegistry_Proxy');
```
gleiser-oliveira, line 18: I'd keep PoolRegistry, both should return the same address

package.json
gleiser-oliveira: An additional comment here is that the backend and subgraph are using 1.0.0, so it would make them all be in sync

therealemjy: Yep, good point. Let me make the update

src/clients/api/queries/getIsolatedPools/index.ts
therealemjy: They should yeah, but my understanding is that using the proxy should be safer since if a new Pool Registry contract is deployed then the proxy will automatically point to it, meaning we won't have to redeploy the frontend.
Perhaps I'm wrong?

gleiser-oliveira: AFAIK PoolRegistry_Proxy is autogenerated and has the same address as PoolRegistry, meaning that whenever we update the package, the address returned by both of them will be the same. I think it's worth confirming that with contracts to have clarity.

src/constants/tokens/vBep/mainnet.ts
```diff
@@ -172,5 +172,25 @@ export const MAINNET_VBEP_TOKENS = {
...
     underlyingToken: MAINNET_TOKENS.wbeth,
   } as VToken,
-  // TODO: add tokens from isolated pools
+
+  // Isolated assets
+  // Pool 1 – Stable Coins
+  '0xca2d81aa7c09a1a025de797600a7081146dceed9': {
+    address: '0xca2d81aa7c09a1a025de797600a7081146dceed9',
+    decimals: 8,
+    symbol: 'vHAY',
```
chechu, line 181: Where do we use this symbol? Because the real symbol of the vToken is `vHAY_Stablecoins`. The rule for the symbols in IL is: `v` + `<underlying token symbol>` + `_<pool name>`. Do we need to consider it?

therealemjy: It's only used for display on the frontend, we never actually use the symbol in any transaction (we always use the address)

gleiser-oliveira (approved):

PR #1162, opened 2023-07-10
[VEN-1673] Allowance revoking feature

therealemjy:
## Jira ticket(s)

VEN-1673

## Changes

- create `SpendingLimit` component
- display spending limit on Supply form and let user revoke it
- display spending limit on Repay form and let user revoke it
- display spending limit on Swap page and let user revoke it
- display spending limit on Repay VAI tab of the VAI page and let user revoke it
- display spending limit on Stake modal of the Vault page and let user revoke it
- update `SwapDetails` so it can be shared with Swap page
- add `useRevokeSpendingLimit` mutation function
- update how spending allowance is refetched after certain mutations


src/hooks/useOperationModal/Modal/RepayForm/__tests__/__snapshots__/indexIntegratedSwap.spec.tsx.snap
```diff
@@ -2,59 +2,59 @@
...
   <div
-    class="css-u34kla-SwapDetails"
+    class="css-s1vrvo-SwapDetails"
   >
     <div
       class="css-wkfae5-LabeledInlineContent"
     >
       <span
-        class="MuiTypography-root MuiTypography-body1 css-1degdl8-MuiTypography-root-LabeledInlineContent"
+        class="MuiTypography-root MuiTypography-body1 css-16extv5-MuiTypography-root-LabeledInlineContent"
```
coreyar: If we are using storybooks snapshots correctly we don't need to snapshot UIs and can avoid noisy diffs like this one.

src/hooks/useOperationModal/Modal/SupplyForm/__tests__/__snapshots__/indexIntegratedSwap.spec.tsx.snap
```diff
@@ -2,59 +2,59 @@
...
   <div
-    class="css-u34kla-SwapDetails"
+    class="css-s1vrvo-SwapDetails"
   >
     <div
       class="css-wkfae5-LabeledInlineContent"
     >
       <span
-        class="MuiTypography-root MuiTypography-body1 css-1degdl8-MuiTypography-root-LabeledInlineContent"
+        class="MuiTypography-root MuiTypography-body1 css-16extv5-MuiTypography-root-LabeledInlineContent"
```
coreyar: Same comment on this diff, it isn't as useful as snapshots because you can't understand the impact of the css changes from this diff

src/components/SpendingLimit/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const handleRevoke = async () => {
+    try {
+      await onRevoke();
+    } catch (error) {
+      let { message } = error as Error;
+
+      if (error instanceof VError) {
+        message = formatVErrorToReadableString(error);
+      }
```
coreyar, line 46: Since message isn't guaranteed to be user friendly you could refactor this to always show the user the value from `formatVErrorToReadableString` and return a generic message if the error can't be parsed.

```suggestion
        const message = formatErrorToReadableString(error);
```

src/hooks/useOperationModal/Modal/WithdrawForm/index.tsx
```diff
@@ -127,17 +127,17 @@ export const WithdrawFormUi: React.FC<WithdrawFormUiProps> = ({
...
-              White: <span css={sharedStyles.whiteLabel} />,
-            }}
-            values={{ amount: readableWithdrawableAmountTokens }}
-          />
-        }
       />
 
+      <LabeledInlineContent
+        label={t('operationModal.withdraw.withdrawableAmount')}
+        css={sharedStyles.getRow({ isLast: true })}
```
coreyar, line 134: You could avoid the javascript and do this in pure css using last child selector

src/hooks/useOperationModal/Modal/BorrowForm/index.tsx
```diff
@@ -164,6 +161,15 @@ export const BorrowFormUi: React.FC<BorrowFormUiProps> = ({
         )}
       </div>
 
+      <LabeledInlineContent
+        label={t('operationModal.borrow.borrowableAmount')}
+        css={sharedStyles.getRow({ isLast: true })}
```
coreyar, line 166: Could be done in pure css with last child selector

src/hooks/useOperationModal/Modal/WithdrawForm/index.tsx
therealemjy: Not in this instance, the name might be misleading but what this really means is it applies a larger bottom margin, but that rule isn't always applied to the last element.
I'll change the name for something that makes more sense.

src/hooks/useOperationModal/Modal/BorrowForm/index.tsx
therealemjy: Same as previously

src/hooks/useOperationModal/Modal/RepayForm/__tests__/__snapshots__/indexIntegratedSwap.spec.tsx.snap
therealemjy: Good catch. In this case, we forgot to use to snapshot the text content rather than the UI.

The goal here isn't to test the UI actually, but the behavior of the component. I'll make the change so we test the text content.

src/hooks/useOperationModal/Modal/SupplyForm/__tests__/__snapshots__/indexIntegratedSwap.spec.tsx.snap
therealemjy: Same as before, we indeed should not test css differences

src/hooks/useOperationModal/Modal/SupplyForm/SubmitSection/index.tsx
```diff
@@ -72,12 +84,36 @@ export const SubmitSection: React.FC<SubmitSectionProps> = ({
       });
     }
 
+    if (!isFormSubmitting && formError === 'HIGHER_THAN_WALLET_SPENDING_LIMIT') {
+      return t('operationModal.supply.submitButtonLabel.amountHigherThanWalletWalletSpendingLimit');
+    }
+
+    if (!isFormSubmitting && formError === 'PRICE_IMPACT_TOO_HIGH') {
```
coreyar, line 91: Some ideas to improve this 
- Use yup or zod to handle form validation and return the error message. This removes the need for all the error condiationals
- This could be abstracted to a function that does a single check for `isFormSubmitting`

coreyar (approved):

src/components/SpendingLimit/index.tsx
therealemjy: It makes sense from a user standpoint, however at the moment our error handling isn't good enough to show generic error messages: we just disabled Sentry so we depend on users to report errors to us, which they do through screenshots... It's dirty. Hopefully we can activate Sentry again soon and improve our error handling.

src/hooks/useOperationModal/Modal/SupplyForm/SubmitSection/index.tsx
therealemjy: Agreed, that would make things cleaner. I've created a ticket (VEN-1732) for this task.

gleiser-oliveira (approved):

PR #1163, opened 2023-07-10
hotfix: skip BIFI from listed assets

gleiser-oliveira:
## Changes

- Skips listing BIFI

### DIFF
--- a/src/clients/api/queries/getIsolatedPools/formatOutput/index.ts
+++ b/src/clients/api/queries/getIsolatedPools/formatOutput/index.ts
@@ -12,6 +12,7 @@ import {
 
 import { BLOCKS_PER_DAY } from 'constants/bsc';
 import { COMPOUND_DECIMALS } from 'constants/compoundMantissa';
+import { MAINNET_TOKENS } from 'constants/tokens';
 import { logError } from 'context/ErrorLogger';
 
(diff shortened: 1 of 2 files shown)


src/clients/api/queries/getIsolatedPools/formatOutput/index.ts
```diff
@@ -71,6 +71,11 @@ const formatToPools = ({
         return acc;
       }
 
+      // Skip BIFI
+      if (vToken.address.toLowerCase() === '0xc718c51958d3fd44f5f9580c9ffac2f89815c909') {
```
therealemjy: I'd recommend making this more generic and adding a comment to explain why we do that:

```typescript
import { MAINNET_TOKENS } from 'constants/tokens';
import { areTokensEqual } from 'utilities';

// Temporary hotfix following multichain issue
if (areTokensEqual(vToken.underlyingToken, MAINNET_TOKENS.bifi)) {
```

therealemjy (changes requested):

src/clients/api/queries/getIsolatedPools/formatOutput/index.ts
gleiser-oliveira: Good call, made the change

therealemjy (approved):

PR #1172, opened 2023-07-12
[VEN-1708] Calculate distributions using underlying decimals

gleiser-oliveira:
## Jira ticket(s)

VEN-1708

## Changes

- Takes the underlying token decimals into account when calculating the distributions, instead of always using `COMPOUND_DECIMALS`, which is a constant with value 18

### DIFF
--- a/src/clients/api/queries/getIsolatedPools/formatOutput/formatToDistributions.ts
+++ b/src/clients/api/queries/getIsolatedPools/formatOutput/formatToDistributions.ts
@@ -32,12 +32,18 @@ const formatToDistributions = (rewardsDistributorsResults: ContractCallReturnCon
         const {
           apyPercentage: supplyApyPercentage,
           dailyDistributedTokens: supplyDailyDistributedTokens,
-        } = calculateApy(supplySpeedMantissa);
+        } = calculateApy({
+          ratePerBlockMantissa: supplySpeedMantissa,
+          decimals: rewardToken.decimals,
+        });
 
         const {
           apyPercentage: borrowApyPercentage,
           dailyDistributedTokens: borrowDailyDistributedTokens,
-        } = calculateApy(borrowSpeedMantissa);
+        } = calculateApy({
+          ratePerBlockMantissa: borrowSpeedMantissa,
+          decimals: rewardToken.decimals,
+        });
 
         // Initialize with an empty array if necessary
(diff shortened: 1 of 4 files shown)


src/utilities/calculateApy.ts
```diff
@@ -4,17 +4,20 @@ import { BLOCKS_PER_DAY } from 'constants/bsc';
 import { COMPOUND_DECIMALS, COMPOUND_MANTISSA } from 'constants/compoundMantissa';
 import { DAYS_PER_YEAR } from 'constants/daysPerYear';
 
-const calculateApy = (ratePerBlockMantissa: BigNumber | string | number) => {
+const calculateApy = (
+  ratePerBlockMantissa: BigNumber | string | number,
+  decimals = COMPOUND_DECIMALS,
+) => {
```
therealemjy: Just to make it easier to understand what needs to be passed to this function, I'd use an object to define the arguments:
```suggestion
export interface CalculateApyInput {
  ratePerBlockMantissa: BigNumber | string | number,
  decimals = COMPOUND_DECIMALS,
};

const calculateApy = ({
  ratePerBlockMantissa,
  decimals,
}: CalculateApyInput) => {
```

This way when using it we can avoid some confusion as to what needs to be passed and in what order.

src/clients/api/queries/getIsolatedPools/formatOutput/formatToDistributions.ts
```diff
@@ -27,17 +37,19 @@ const formatToDistributions = (rewardsDistributorsResults: ContractCallReturnCon
...
 
+      const underlyingDecimal = underlyingDecimals[r / 2];
+
       // Only add distribution if one of the speeds is not 0
       if (supplySpeedMantissa.isGreaterThan(0) || borrowSpeedMantissa.isGreaterThan(0)) {
         const {
           apyPercentage: supplyApyPercentage,
           dailyDistributedTokens: supplyDailyDistributedTokens,
-        } = calculateApy(supplySpeedMantissa);
+        } = calculateApy(supplySpeedMantissa, underlyingDecimal);
```
therealemjy: Can't you use `rewardToken.decimals` directly?

chechu (commented):
lgtm

therealemjy (approved):

PR #1191, opened 2023-07-18
[VEN-1741] Update mainnet and testnet RPC URLs

gleiser-oliveira:
## Jira ticket(s)

VEN-1741

## Changes

- Updated RPC URLs for mainnet and testnet
- Added `VITE_RPC_URL` env var, meant to be used during local development


src/config/index.ts
```diff
@@ -56,12 +63,31 @@ export const ENV_VARIABLES = {
 const environment: Environment =
   (ENV_VARIABLES.VITE_ENVIRONMENT as Environment | undefined) || 'mainnet';
 
+const isLocalServer =
+  (window.location.hostname === 'localhost' || window.location.hostname === '127.0.0.1') &&
+  environment !== 'ci' &&
+  environment !== 'storybook';
```
therealemjy: I'd suggest using Vite's `DEV` environment variable for that, as it will cover all cases (localhost, but also if the dApp is running locally with the `--host` flag enable, in which case it would be hosted on an ip address):
```suggestion
const isLocalServer =
  import.meta.env.DEV &&
  environment !== 'ci' &&
  environment !== 'storybook';
```

src/config/index.ts
```diff
@@ -56,12 +63,31 @@ export const ENV_VARIABLES = {
...
+  case 'preview':
+    localRpcUrl = ENV_VARIABLES.VITE_RPC_URL_MAINNET;
+    break;
+  case 'testnet':
+    localRpcUrl = ENV_VARIABLES.VITE_RPC_URL_TESTNET;
+    break;
+  default:
+    localRpcUrl = RPC_URLS[chainId];
+    break;
+}
```
therealemjy: Given `localRpcUrl` will only ever be used with the local environment variables, I think you could simplify this logic:

```suggestion
const localRpcUrl = environment ===  'mainnet' || environment ===  'preview' ? ENV_VARIABLES.VITE_RPC_URL_MAINNET : ENV_VARIABLES.VITE_RPC_URL_TESTNET;
```

therealemjy (approved):

therealemjy (approved):

PR #1195, opened 2023-07-19
[VEN-1723] Fix distribution APYs of isolated assets

therealemjy:
…dd @venusprotocol/oracle package## Jira ticket(s)

VEN-1723

## Changes

- remove hotfix to ignore distribution APYs of isolated assets
- update how distribution APYs of isolated assets are calculated to take token prices in consideration
- add `multiplyMantissaDaily` util
- simplify `calculateApy` util
- fetch token prices from ResilientOracle contract
- rename `borrowRatePerBlockTokens` property of `Asset` type to `borrowPercentageRatePerBlock`
- rename `supplyRatePerBlockTokens` property of `Asset` type to `supplyPercentageRatePerBlock`


src/utilities/calculateDailyDistributedTokens/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export interface CalculateDailyDistributedTokensInput {
+  ratePerBlockMantissa: BigNumber | string | number;
+}
+
+const calculateDailyDistributedTokens = ({
+  ratePerBlockMantissa,
+}: CalculateDailyDistributedTokensInput) =>
+  new BigNumber(ratePerBlockMantissa)
+    .div(COMPOUND_MANTISSA)
```
therealemjy, line 14: @kkirka @chechu 
I'm not sure whether this is correct. I asked in the Contract channel and Manmeet confirmed vToken rates from contracts are expressed with 18 decimals all the time.

@gleiser-oliveira However I know you made some updates previously so that token distributions would be calculated using the decimals of the reward token.

Can you guys confirm whether rates for vToken (`supplyRatePerBlock` and `borrowRatePerBlock`) always use 18 decimals, or if rates for distributions follow a different rule please?

src/clients/api/queries/getIsolatedPools/formatOutput/formatDailyDistributedRewardTokensMapping.ts
```diff
@@ -0,0 +1,84 @@
...
+      for (let r = 0; r < speedResults.length - 1; r += 2) {
+        const vTokenAddress = speedResults[r].methodParameters[0].toLowerCase();
+        const supplySpeedMantissa = new BigNumber(speedResults[r].returnValues[0].hex);
+        const borrowSpeedMantissa = new BigNumber(speedResults[r + 1].returnValues[0].hex);
+
+        // Only add distribution if one of the speeds is not 0
+        if (supplySpeedMantissa.isGreaterThan(0) || borrowSpeedMantissa.isGreaterThan(0)) {
+          const supplyDailyDistributedRewardTokens = calculateDailyDistributedTokens({
+            ratePerBlockMantissa: supplySpeedMantissa,
+            decimals: rewardToken.decimals,
```
therealemjy, line 60: @kkirka @chechu 
Is it correct here to use the decimals of the reward token, and not 18 decimals all the time?
I'm a bit confused with what decimals are used for what values. Manmeet confirmed me the `supplyRatePerBlock` and `borrowRatePerBlock` values of a VToken (returned in the `getAllPools` response of the PoolLens contract for exampe) always use 18 decimals, yet we recently implemented a fix to calculate distribution APYs using the decimals of the reward token.

kkirka: I think the confusion comes from the fact we mix up "rate", "speed", etc. terms.
1) For a value that is inherently a _percentage_ like interest rates (supplyRatePerBlock) or factors (collateral factor, liquidation threshold, etc.) we use 18 decimals.
2) For values that are expressed in _tokens_ (e.g., distribution speed — how many tokens are distributed per block) we use token decimals.
3) For values that are in vTokens (e.g., vToken.totalSupply(), vToken.balanceOf(user)) we use vToken decimals which is 8 for all vTokens currently, and we'll most likely follow this convention going forward.

I'm not sure how we can fix this confusion though. Maybe we have this info missing in docs somewhere? We can add it, can you list the contract functions you think this info should be added to?

therealemjy: I see. I remember we had that discussion but I still wasn't sure this was correct. Thanks for the explanation.

Adding this to a doc somewhere would be great yeah: in Notion perhaps?
For now I'm clear, these 3 rules should be enough.

kkirka: We could use custom types for this, e.g.
```
type Exp is uint256; // 18 decimals
type UnderlyingAmount is uint256; // underlying decimals
type VTokenAmount is uint256; // vToken decimals
type Price is uint256; // (36 - underlying decimals) decimals
type ExchangeRate is uint256; // (underlying decimals + 18 - vToken decimals) decimals 
```

...and then have conversion functions in the code.

I experimented with this, but it would require quite a huge refactoring of the codebase, so I figured it wasn't a priority.

kkirka: Answering your question, you should use reward token decimals here because `supplySpeedMantissa` is in reward tokens. Not sure it's correct to call it `ratePerBlockMantissa` because rate feels like a _percentage_

therealemjy: Your proposition with types makes sense to me, but if it involves heavy refactors I agree with you that's not a priority.

The ideal scenario would be to have the contacts fully documented with explanations on what the return types correspond to (I haven't had a chance to look at the new documentation thoroughly: maybe that's included in it already?).
Otherwise having the 3 rules you wrote somewhere in Notion would also be helpful.

I agree with your last comment about the name of the parameter. To make things less confusing, I've renamed it to `mantissa`.

Thanks for responding and giving me all the explanations: I feel more confident with this PR now!

src/clients/api/queries/getIsolatedPools/formatOutput/formatDailyDistributedRewardTokensMapping.ts
```diff
@@ -0,0 +1,84 @@
...
+      const results = rewardsDistributorsResult.callsReturnContext;
+      const rewardTokenAddress = results[0].returnValues[0].toLowerCase();
+      const rewardToken = getTokenByAddress(rewardTokenAddress);
+
+      if (!rewardToken) {
+        logError(`Record missing for reward token: ${rewardTokenAddress}`);
+        return acc;
+      }
+
+      const rewardTokenPriceDollars = tokenPricesDollars[rewardToken.address.toLowerCase()];
```
chechu: AFAIU, we could reuse here `rewardTokenAddress`, that is already in lower case

src/utilities/calculateApy/index.ts
```diff
@@ -0,0 +1,19 @@
+import BigNumber from 'bignumber.js';
+
+import { COMPOUND_DECIMALS } from 'constants/compoundMantissa';
+import { DAYS_PER_YEAR } from 'constants/daysPerYear';
+
+export interface CalculateApyInput {
+  dailyDistributedTokens: BigNumber;
+}
+
+const calculateApy = ({ dailyDistributedTokens }: CalculateApyInput) =>
```
chechu: The param is not a number of tokens, but a daily percentage. Maybe the param name can generate some confusion

package.json
```diff
@@ -48,6 +48,7 @@
     "@uiw/react-markdown-preview": "^4.1.13",
     "@uiw/react-md-editor": "^3.14.3",
     "@venusprotocol/isolated-pools": "^1.1.0",
+    "@venusprotocol/oracle": "^1.7.0-dev.5",
```
chechu: New main release will be published soon with the needed changes. See https://github.com/VenusProtocol/oracle/pull/112

chechu:
LGTM

src/clients/api/queries/getIsolatedPools/formatOutput/formatDailyDistributedRewardTokensMapping.ts
therealemjy: Good catch

src/utilities/calculateApy/index.ts
therealemjy: Are you sure?
If you look at these lines: https://github.com/VenusProtocol/venus-protocol-interface/pull/1195/files#diff-9c59ea2aace70b47e0e69728f671a18140352c5b43e537c697349500de8dad5bR148-R154 the value we first calculate is `supplyDailyDistributedTokens`, which is one part of the value we display on the Market page (we add `borrowDailyDistributedTokens` to it to get the total amount of tokens distributed per day): 
Screen Shot 2023-07-21 at 13 23 13

This value is expressed in tokens, and it's this value we then pass to `calculateApy` to determine the distribution APY.

package.json
therealemjy: Ok great, thanks for the heads up

chechu: oracles 1.7.0 is ready: https://github.com/VenusProtocol/oracle/releases/tag/v1.7.0

therealemjy: Nice, thanks. I've made the update

chechu (approved):

gleiser-oliveira (approved):

PR #1196, opened 2023-07-19
fix: crash when transaction.tokenAddress is null

gleiser-oliveira:
## Changes

- This attempts to fix a crash in the history page when listing transactions that have their tokenAddress as null
- It is expected that a transaction comes with a tokenAddress null from the API, so this has been ajusted in the corresponding type
- Added checks to avoid trying to lowercase a null address, thus avoiding the crash
- Defaults to BNB in `getTokenByAddress` if the address is null

### DIFF
--- a/src/utilities/getTokenByAddress.ts
+++ b/src/utilities/getTokenByAddress.ts
@@ -1,15 +1,18 @@
 import { Token } from 'types';
+import { areAddressesEqual } from 'utilities';
 
 import { TOKENS } from 'constants/tokens';
 
-const getTokenByAddress = (address: string) => {
+const getTokenByAddress = (address?: string | null) => {
   let token: Token | undefined;
 
   Object.keys(TOKENS)
     .filter(key => Object.prototype.hasOwnProperty.call(TOKENS, key))
     .forEach(tokenId => {
       const currentToken = TOKENS[tokenId as keyof typeof TOKENS];
-      if (currentToken?.address.toLowerCase() === address.toLowerCase()) {
+      if (!address) {
+        token = TOKENS.bnb;
+      } else if (areAddressesEqual(currentToken?.address, address)) {
         token = currentToken;
       }
     });
--- a/src/utilities/getVTokenByAddress.ts
+++ b/src/utilities/getVTokenByAddress.ts
@@ -2,8 +2,8 @@ import { VToken } from 'types';
 
 import { VBEP_TOKENS } from 'constants/tokens';
 
-const getVTokenByAddress = (address: string) =>
-  address.toLowerCase() in VBEP_TOKENS
+const getVTokenByAddress = (address?: string | null) =>
+  !!address && address.toLowerCase() in VBEP_TOKENS
     ? (VBEP_TOKENS[address.toLowerCase() as keyof typeof VBEP_TOKENS] as VToken)
     : undefined;
 
--- a/src/clients/api/queries/getTransactions/types.ts
+++ b/src/clients/api/queries/getTransactions/types.ts
@@ -8,5 +8,5 @@ export interface TransactionResponse {
   to: string;
   transactionHash: string;
   logIndex: string;
-  tokenAddress: string;
+  tokenAddress: string | null;
 }


coreyar (approved):

src/utilities/getVTokenByAddress.ts
```diff
@@ -2,8 +2,8 @@ import { VToken } from 'types';
 
 import { VBEP_TOKENS } from 'constants/tokens';
 
-const getVTokenByAddress = (address: string) =>
-  address.toLowerCase() in VBEP_TOKENS
+const getVTokenByAddress = (address: string | null) =>
+  !!address && address.toLowerCase() in VBEP_TOKENS
     ? (VBEP_TOKENS[address.toLowerCase() as keyof typeof VBEP_TOKENS] as VToken)
     : undefined;
```
coreyar, line 8: How are you getting vBnb if there is no underlying address for it?

src/utilities/getTokenByAddress.ts
```diff
@@ -2,14 +2,16 @@ import { Token } from 'types';
...
   let token: Token | undefined;
 
   Object.keys(TOKENS)
     .filter(key => Object.prototype.hasOwnProperty.call(TOKENS, key))
     .forEach(tokenId => {
       const currentToken = TOKENS[tokenId as keyof typeof TOKENS];
-      if (currentToken?.address.toLowerCase() === address.toLowerCase()) {
+      if (!address) {
+        token = TOKENS.bnb;
+      } else if (currentToken?.address.toLowerCase() === address.toLowerCase()) {
```
therealemjy: You could use the areTokensEqual util here (it does exactly the same thing)

src/utilities/getVTokenByAddress.ts
```diff
@@ -2,8 +2,8 @@ import { VToken } from 'types';
 
 import { VBEP_TOKENS } from 'constants/tokens';
 
-const getVTokenByAddress = (address: string) =>
-  address.toLowerCase() in VBEP_TOKENS
+const getVTokenByAddress = (address: string | null) =>
```
therealemjy: In this case we could just make address optional, so that undefined could be passed to the function as well

src/utilities/getVTokenByAddress.ts
gleiser-oliveira: This function won't receive an underlying address, it receives the address for vBNB, which is a valid key in the VBEP_TOKENS object. `getTokenByAddress` is the one that needs the fallback, since it deals with underlying tokens

src/utilities/getTokenByAddress.ts
gleiser-oliveira: `areTokensEqual` wouldn't be compatible with `address`, as it is a string here (maybe we want another util for these cases?)

therealemjy: Sorry, I mistyped. The util I was thinking about is `areAddressesEqual`

gleiser-oliveira: There you go, I was just looking at the list of utils for it, done!

src/utilities/getVTokenByAddress.ts
gleiser-oliveira: Updated

src/utilities/getVTokenByAddress.ts
coreyar: If the address is null it returns undefined but you it should return the token for vBnb

therealemjy (approved):

coreyar (approved):

src/utilities/getVTokenByAddress.ts
```diff
@@ -1,10 +1,17 @@
...
-import { VBEP_TOKENS } from 'constants/tokens';
+import { MAINNET_VBEP_TOKENS, TESTNET_VBEP_TOKENS, VBEP_TOKENS } from 'constants/tokens';
 
-const getVTokenByAddress = (address: string) =>
-  address.toLowerCase() in VBEP_TOKENS
+const getVTokenByAddress = (address?: string | null) => {
+  // get the vBNB of the current network to serve as a fallback
+  const vBnb = config.isOnTestnet
+    ? TESTNET_VBEP_TOKENS['0x2e7222e51c0f6e98610a1543aa3836e092cde62c']
+    : MAINNET_VBEP_TOKENS['0xa07c5b74c9b40447a954e1466938b865b6bbea36'];
```
therealemjy: I wouldn't fallback to vBNB, as I see two issues with that:
- vBNB only exists in the main pool right now, but technically we could have a vBNB token in an isolated pool in which case defaulting to the main pool token could be misleading
- it will cause issues once we switch to being a multichain dApp

I'd return `undefined` when address is `null` instead. Let me know your thoughts

gleiser-oliveira: Also in the specific case for transactions there is this

```
  const token =
    getTokenByAddress(tokenAddress) ||
    getVTokenByAddress(tokenAddress)?.underlyingToken ||
    TOKENS.xvs;
```

Which covers the undefined case

therealemjy: Yeah that part makes sense to me. Since `getVTokenAddress` is a utility function, it is bound to be used in other places so I'd stick to returning `undefined` and letting logics that use it handle undefined cases.

PR #1210, opened 2023-07-27
[VEN-1671] Update transactions to use the next endpoint version

gleiser-oliveira:
## Jira ticket(s)

VEN-1671

## Changes

- Updated restService logic to be able to send the accept-version header during requests
- Updated signature of the transactions endpoint and transactions model type
- Updated tests


src/clients/api/queries/getTransactions/formatTransaction.ts
```diff
@@ -1,31 +1,41 @@
...
+    : // else get the token from tokenAddress
+      getTokenByAddress(tokenAddress);
+
+  // if neither is found, use XVS
+  const token = transactionToken || TOKENS.xvs;
 
   return {
     ...rest,
-    amountWei: convertTokensToWei({ value: new BigNumber(amount), token }),
+    amountMantissa: new BigNumber(amountMantissa),
```
therealemjy, line 33: Why the name change from wei to mantissa?

src/utilities/restService.ts
```diff
@@ -57,14 +76,18 @@ export async function restService<D>({
...
     .then(async response => {
       const { status } = response;
 
       let data: undefined;
 
       try {
         data = await response.json();
+        const warning = response.headers.get('Warning');
+        if (warning) {
+          console.warn(warning);
```
therealemjy: I'd call the `logError` function (`import { logError } from 'context/ErrorLogger';`), this way this gets reported in Sentry so we get notified that we need to update the endpoint used.

therealemjy (approved):

src/clients/api/queries/getTransactions/formatTransaction.ts
gleiser-oliveira: Particularly, it feels slightly better to use the same term from the payload. Otherwise it looks like we are dealing with two different amount values. Either that or we just use wei/mantissa interchangeably

src/clients/api/queries/getTransactions/formatTransaction.ts
```diff
@@ -1,31 +1,41 @@
...
   timestamp,
+  from,
   ...rest
 }: TransactionResponse) => {
-  const token =
-    getTokenByAddress(tokenAddress) ||
-    getVTokenByAddress(tokenAddress)?.underlyingToken ||
-    TOKENS.xvs;
+  // check if the tokenAddress is from a VToken
+  const vToken = getVTokenByAddress(tokenAddress);
```
coreyar: You might want to reverse the logic since the majority of the time the address will be a token not a vToken

gleiser-oliveira: I actually prefer looking for a VToken first here, so I can construct a Token object like in lines 22-23 without having to null check getTokenByAddress before.

src/clients/api/queries/getTransactions/formatTransaction.ts
therealemjy: I see, that makes sense. We can start using mantissa on the frontend too: I think for now we'll consider wei/mantissa to be interchangeable as you mentioned and progressively move towards using mantissa instead of wei.

therealemjy (changes requested):
Toggling the "My transaction" toggle on the History page seems to make it hang because it throws a JS error:
Screen Shot 2023-09-08 at 09 08 57

therealemjy:
Looks good to me 👍

therealemjy (approved):

PR #1211, opened 2023-07-29
fix: use the correct condition for last rewards block

kkirka:
## Changes

- Fix the condition for accounting last rewards block in distribution APY


therealemjy (approved):
Thank you! Sorry for introducing this bug, this is totally on me

chechu (approved):

PR #1214, opened 2023-08-01
[VEN-1742] Import proposals from JSON files

gleiser-oliveira:
## Jira ticket(s)

VEN-1742

## Changes

- This PR implements the automatic import of proposals using a JSON formatted file, for this we have:
    - Changed the create proposal modal to display two buttons, one leading to the file import, another to the already existing manual process
    - If the automatic import is selected, we ask the user to upload a JSON file containing the specs for the proposal
    - We validate the file, raising any errors we might find
    - If all is correct, we show the user a preview of the proposal to confirm it should be created
- Updated the parsing and encoding of actions when validating them. Previously the tuple type was being parsed incorrectly, which would lead to any tuple argument signature to not have its inner types
- Updated prettier to v3, linting the new and existing files. This was required due to a bug that caused
`type ProposalFormContext = ReturnType<typeof useFormikContext<FormValues>>;` to throw a syntax error


public/maintainance.html
therealemjy, line 1: I didn't even know we had this file... I'll create a ticket to get it removed, as we've never used it

src/errors/importProposalErrorPhrases.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const importProposalErrorPhrases = {
+  noMetaKey: t('errors.noMetaKey'),
+  noSignaturesKey: t('errors.noSignaturesKey'),
+  noTypeKey: t('errors.noTypeKey'),
+  noTargetsKey: t('errors.noTargetsKey'),
+  noParamsKey: t('errors.noParamsKey'),
+  missingTitle: t('errors.missingTitle'),
+  missingTargetForSignature: t('errors.missingTargetForSignature'),
+  missingParamsForSignature: t('errors.'),
```
therealemjy: This key seems incomplete

src/utilities/importJsonProposal/index.ts
therealemjy, line 1: This logic and is specific to the JSON VIP feature and is unlikely to be reused in the dApp, so I'd move it to the Governance page folder instead.

Also, how about using the Yup validation schema of the Create Proposal form (https://github.com/VenusProtocol/venus-protocol-interface/blob/6f802ae11067b7c4b353fffb735baf2a246f21a4/src/pages/Governance/ProposalList/CreateProposalModal/proposalSchema.ts) to validate the uploaded JSON?
This way both flows would follow the same logic.

src/utilities/importJsonProposal/samples/vip-123.json
therealemjy, line 1: I'd move this file to the `assets` folder.

src/utilities/importJsonProposal/__tests__/index.spec.ts
```diff
@@ -0,0 +1,158 @@
+import importJsonProposal from '..';
+import TEST_VIP from '../samples/vip-123.json';
+
+describe('utilities/getSmartDecimalPlaces', () => {
```
therealemjy:
```suggestion
describe('utilities/importJsonProposal', () => {
```

src/pages/Governance/ProposalList/CreateProposalModal/index.tsx
```diff
@@ -40,86 +44,162 @@ export const CreateProposal: React.FC<CreateProposalProps> = ({
...
+        } = await importJsonProposal(file);
+
+        setFieldValue('actions', actions);
+        setFieldValue('title', title);
+        setFieldValue('proposalType', proposalType);
+        setFieldValue('description', description);
+        setFieldValue('forDescription', forDescription);
+        setFieldValue('abstainDescription', abstainDescription);
+        setFieldValue('againstDescription', againstDescription);
+        await validateForm();
```
therealemjy:
```suggestion
        const proposal = await importJsonProposal(file);
        setFieldValues(proposal);
```

`setFieldValues` will automatically trigger a revalidation of the form.

src/pages/Governance/ProposalList/CreateProposalModal/index.tsx
```diff
@@ -40,86 +44,162 @@ export const CreateProposal: React.FC<CreateProposalProps> = ({
...
-  );
+    ];
+    if (proposalMode === 'file') {
+      return initialSteps.concat(uploadFileSteps);
+    }
+    if (proposalMode === 'manual') {
+      return initialSteps.concat(manualProposalSteps);
+    }
+    return initialSteps;
+  }, [proposalMode]);
```
therealemjy: Seeing the logic get more complicated, how about using nested routing to handle the navigation within the Create Proposal modal? https://v5.reactrouter.com/web/example/nesting

src/pages/Governance/ProposalList/CreateProposalModal/index.tsx
```diff
@@ -200,21 +280,30 @@ export const CreateProposal: React.FC<CreateProposalProps> = ({
...
+                ) : (
+                  <PrimaryButton
+                    fullWidth
+                    disabled={getErrorsByStep(currentStep)}
+                    onClick={() => setCurrentStep(currentStep + 1)}
+                  >
+                    {buttonText}
+                  </PrimaryButton>
+                ))}
             </Form>
```
therealemjy: I think this logic is becoming misleading: `CurrentFields` would imply to me that it only contains the form fields, but it could actually contain the initial UI which contains the two buttons to choose what flow to follow. The logic for displaying the submit button is also getting quite complicate in my opinion.

Thinking about it, I'd follow a different approach and instead make every step into a component. Coupled with nested routing, we could get rid of a lot of extra logic while being able to share common steps between both flows (the initial UI and the final step showing a preview of the proposal).

therealemjy (changes requested):
- The description doesn't seem to appear in the VIP preview when uploading one via JSON:
Screen Shot 2023-08-01 at 15 45 24

- After uploading a JSON file, then going back to the previous screen to click on the "Create manually" button, the data from the JSON file is still in the form state:
Screen Shot 2023-08-01 at 15 46 01

I think we should reset the state of the form after uploading a JSON file but going back to the previous step. What do you think?

- I get an error when trying to submit the sample VIP via JSON:
Screen Shot 2023-08-01 at 15 52 25

src/utilities/importJsonProposal/index.ts
gleiser-oliveira: The validation is in fact used when the values are set, this only formats the data from the JSON file to be passed to the form. I'll move it to the Governance folder.

gleiser-oliveira:
> * The description doesn't seem to appear in the VIP preview when uploading one via JSON:

Problem comes from the markdown in the description. I've updated the JSON asset with a valid markdown description if you used that one.

src/pages/Governance/checkImportErrors/index.ts
therealemjy, line 1: Since this file is used by the `CreateProposalModal` only, I'd move it inside its directory.

src/pages/Governance/ProposalList/CreateProposalModal/ProposalWizard/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+        </Route>
+        <Route path={`${path}/proposal-descriptions`}>
+          <VotingDescriptions />
+        </Route>
+        <Route path={`${path}/proposal-actions`}>
+          <ActionAccordion />
+        </Route>
+        <Route path={`${path}/proposal-preview`}>
+          <ProposalPreview />
+        </Route>
```
therealemjy: Let's add these routes to the `routing` [file](https://github.com/VenusProtocol/venus-protocol-interface/blob/18f5d7c790c9c75fb3611244e3cf581d528be72a/src/constants/routing.ts) and import them from there. This way you'll also be able to check what the current step is by using the path directly

src/pages/Governance/ProposalList/CreateProposalModal/UploadOrManualProposal/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+          id="proposalFile"
+          ref={inputFile}
+          css={styles.fileInput}
+          data-testid={TEST_IDS.fileInput}
+          onChange={e => {
+            e.stopPropagation();
+            e.preventDefault();
+            const file = e.target.files ? e.target.files[0] : null;
+            onClickUploadFile(file);
+          }}
```
therealemjy: I'm generally not a fan of having inline functions in the return statement of a component, I feel it makes it harder to read. How about moving this to a function outside the return statement?

src/pages/Governance/ProposalList/CreateProposalModal/UploadOrManualProposal/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+        />
+        <SecondaryButton css={styles.option} onClick={() => inputFile.current?.click()}>
```
therealemjy: Same thing here, I'd move inline function into a function outside the return statement

src/pages/Governance/importJsonProposal/index.ts
therealemjy, line 1: Since this is used by the `CreateProposalModal` component only, I'd move it inside its directory.

src/pages/Governance/importJsonProposal/index.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  const formattedProposal = {
+    title,
+    description,
+    forDescription,
+    abstainDescription,
+    againstDescription,
+    proposalType,
+    actions,
+  };
```
therealemjy: Wouldn't you be able to leverage the validation schema of the form to handle the validation of this object too?

You could shape the data you get from the uploaded JSON, like you do here, then pass this object to the validation schema. Depending on the errors it returns, you can then decide on what error message to throw/display.
The advantage I see is that you'd centralize validation and use the same set of rules for both the upload and manual flows. Does that make sense?

therealemjy (commented):
This looks so much cleaner. Thanks for refactoring this form to use a nested router (on top of the feature you're adding). We'll be able to use this form as an example for future similar flows.

therealemjy (approved):

therealemjy:
Looks good to me 👍

PR #1218, opened 2023-08-02
feat: make the dApp resilient to the subgraph being down

therealemjy:
## Changes

- update `getIsolatedPools` query function so it does not throw an error if the request to the subgraph to fetch the participant counts fails


src/clients/api/queries/getIsolatedPools/index.ts
```diff
@@ -29,6 +29,17 @@ export type { GetIsolatedPoolsInput, GetIsolatedPoolsOutput } from './types';
 const POOL_REGISTRY_ADDRESS = getContractAddress('PoolRegistry_Proxy'); // TODO: get from package (see VEN-1746)
 const RESILIENT_ORACLE_ADDRESS = getContractAddress('oracle'); // TODO: get from package (see VEN-1746)
 
+// Since the borrower and supplier counts aren't essentially information, we make the logic so the
```
gleiser-oliveira:
```suggestion
// Since the borrower and supplier counts aren't essential information, we make the logic so the
```

therealemjy: Good catch

gleiser-oliveira (approved):

PR #1298, opened 2023-08-22
[VEN-1718] Capture analytic events

therealemjy:
## Jira ticket(s)

VEN-1718

## Changes

- update the `enterMarkets` mutation function to `enterMarket` so it only take one address as parameter
- add `useAnalytics` hook
- capture analytic events


src/clients/api/mutations/vote/useCastVote.ts
```diff
@@ -29,6 +31,18 @@ const useCastVote = (options?: Options) => {
       onSuccess: (...onSuccessParams) => {
         const { proposalId, voteType } = onSuccessParams[1];
 
+        let voteTypeWord: AnalyticEventProps<'Vote casted'>['voteType'] = 'abstain';
+        if (voteType === 0) {
```
coreyar: Could also be defined and used as an enum

coreyar (approved):

src/clients/api/mutations/vote/useCastVote.ts
therealemjy: Good point, that would be cleaner too. I'll make the change

PR #1344, opened 2023-08-31
fix: await setValues when importing a VIP file

gleiser-oliveira:
## Changes

- Some imports were failing validation, surfacing that the async function `setValues` was missing an await


chechu:
LGTM

therealemjy (approved):

PR #1477, opened 2023-09-25
refactor: optimize contracts package

therealemjy:
## Changes

The main goal of this PR is to update the contracts package with the goal of reducing the final bundle size. The ABIs and contract addresses are pulled from different Venus packages, which contain large JSON files. We end up adding a lot of unnecessary weight to the dApp as only part of these JSON files are being used.

What this updated package does is it extract ABIs from the Venus packages and creates a JSON file for each contract ABI. It also extracts their addresses and saves them into an `addresses.ts` file, generates types for each contract using the extracted ABIs and finally it generates, for each contract, a function and a hook to create a contract instance. e.g.: for the PoolLens contract, the`getPoolLensContract` function and `useGetPoolLensContract` hook are created at `src/packages/contractsNew/getters/venusLens.ts`.

The dApp can then import just the functions/hooks it needs. The package has also been built in such a way that each contract function/hook pair is independent, so importing one will only import this contract's ABI and not all the contract ABIs (which is the main reason why this package will drastically reduce the final bundle size).

In order to limit the amount of changes to review (although there's still quite a lot of them...), this PR creates a new contracts package (placed in the directory `src/packages/contractsNew`) without being used. A next PR will replace the existing contracts packages and its usage with the new one.


src/clients/api/queries/getVTokens/useGetVTokens.ts
```diff
@@ -28,9 +29,7 @@ const useGetVTokens = (options?: Options) => {
   const { chainId } = useAuth();
   const tokens = useGetTokens();
 
-  const venusLensContract = useGetUniqueContract({
-    name: 'venusLens',
-  });
+  const venusLensContract = useGetVenusLensContract();
```
therealemjy, line 32: This is an example of how the new package is used

gleiser-oliveira (approved):

PR #1484, opened 2023-09-26
refactor: optimize request refetches

therealemjy:
The goal of this PR is to reduce the overall amount of requests made and only reduce the refetch interval of requests for which that is really relevant.

## Changes
- remove refetch interval from queries that don't require to be refetched often
- replace the default refetch interval of 9 seconds, used on queries that need to be refetched often, with a pseudo random refetch interval of 9 to 15 seconds (3 to 5 blocks). The reason for using a pseudo random interval is to avoid the [thundering herd problem](https://en.wikipedia.org/wiki/Thundering_herd_problem#:~:text=In%20computer%20science%2C%20the%20thundering,but%20only%20one%20will%20win.)
- update auth context so the account address is returned even if the signer is `undefined` or the status of the wallet is not "connected". This means that for returning users the account address is instantly returned upon opening the dApp, reducing the amount of total requests made, as in the current implementation a query that uses the account address would be fired on mount, then fired again after the account address was returned

### DIFF
--- a/src/clients/api/queries/getIsolatedPools/useGetIsolatedPools.ts
+++ b/src/clients/api/queries/getIsolatedPools/useGetIsolatedPools.ts
@@ -1,11 +1,10 @@
 import { QueryObserverOptions, useQuery } from 'react-query';
-import { callOrThrow } from 'utilities';
+import { callOrThrow, generatePseudoRandomRefetchInterval } from 'utilities';
 
 import getIsolatedPools, {
   GetIsolatedPoolsInput,
   GetIsolatedPoolsOutput,
 } from 'clients/api/queries/getIsolatedPools';
-import { DEFAULT_REFETCH_INTERVAL_MS } from 'constants/defaultRefetchInterval';
 import FunctionKey from 'constants/functionKey';
 import { useAuth } from 'context/AuthContext';
 import useGetTokens from 'hooks/useGetTokens';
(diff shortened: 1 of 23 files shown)


src/clients/api/queries/getIsolatedPools/useGetIsolatedPools.ts
```diff
@@ -29,6 +28,10 @@ type Options = QueryObserverOptions<
   [FunctionKey.GET_ISOLATED_POOLS, TrimmedInput]
 >;
 
+const refetchInterval = generatePseudoRandomRefetchInterval();
+
+console.log(refetchInterval);
```
gleiser-oliveira: console.log

src/clients/api/queries/getMainPool/useGetMainPool.ts
```diff
@@ -31,6 +30,10 @@ type Options = QueryObserverOptions<
   [FunctionKey.GET_MAIN_POOL, TrimmedInput]
 >;
 
+const refetchInterval = generatePseudoRandomRefetchInterval();
+
+console.log(refetchInterval);
```
gleiser-oliveira: console.log

src/clients/api/queries/getPendingRewards/useGetPendingRewards.ts
```diff
@@ -35,6 +34,10 @@ type Options = QueryObserverOptions<
   [FunctionKey.GET_PENDING_REWARDS, TrimmedGetPendingRewardsInput]
 >;
 
+const refetchInterval = generatePseudoRandomRefetchInterval();
+
+console.log(refetchInterval);
```
gleiser-oliveira: console.log

gleiser-oliveira (approved):

PR #1504, opened 2023-09-28
refactor: use new contracts package

therealemjy:
## Jira ticket(s)

VEN-2008

## Changes

Migrate to using the new contracts package. With this PR, the old contracts package is no longer used and we've removed 2mb of unused files. The next PR will remove the old contracts package from the codebase.


gleiser-oliveira (approved):
Very, very nitpicky comment: `contractsNew` feels a bit weird as the package name. I'd use a version number.

PR #1591, opened 2023-10-15
fix: remove pow function usage

therealemjy:
## Changes

- replace usage of the `pow` function from bignumber.js with `**` (`Math.pow()`), as the `pow` function is too expensive to run and ends up blocking the main thread quite significantly when called many times in a row (such as in the `getMainPool` and `getIsolatedPools` functions)


src/clients/api/queries/getIsolatedPools/__tests__/__snapshots__/index.spec.ts.snap
gleiser-oliveira, line 1: Some results are `Infinity`

src/clients/api/queries/getIsolatedPools/__tests__/__snapshots__/index.spec.ts.snap
```diff
@@ -94,13 +94,13 @@ exports[`api/queries/getIsolatedPools > returns isolated pools in the correct fo
...
         {
-          "borrowApyPercentage": "2.014488268288307401",
+          "borrowApyPercentage": "2.014488268285719",
           "borrowBalanceCents": "0",
           "borrowBalanceTokens": "0",
           "borrowCapTokens": "400000",
           "borrowDistributions": [
             {
-              "apyPercentage": "9.043693025262964313386387663825371300504104188785330896893565789987545579127055429069323635317387647909529452260341021253946764927771680348464445898439384525165998607391664168111161332361457382484356298080225495666764015575308933862550916891578065520226087412437428029555679042896644062315930188946310835086919038925876599304407561953491660141829283562214983175072577507413505264909735341975263567052662333003543379441477606078483092391158302886569231609113478749386614271074231851701401628318399465705490915827283170241214244615293752252032691043903504205999241905301306055286770517315811324821862294542079922467097443741700021587052520981326419063555759994546260795733927905348612198852958046780884967914111088710992984703292022e+711",
+              "apyPercentage": "Infinity",
```
gleiser-oliveira, line 103: Like this

src/clients/api/queries/getVaiRepayApy/__snapshots__/index.spec.ts.snap
```diff
@@ -2,6 +2,6 @@
...
+  "repayApyPercentage": "Infinity",
```
gleiser-oliveira, line 5: `Infinity`

src/clients/api/queries/getVTokenApySimulations/__tests__/__snapshots__/index.spec.ts.snap
```diff
@@ -4,503 +4,503 @@ exports[`api/queries/getVTokenApySimulations > returns the APY simulations in th
 {
   "apySimulations": [
     {
-      "borrowApyPercentage": 1.0483200000054798e-9,
-      "supplyApyPercentage": 2.096640000021919e-9,
+      "borrowApyPercentage": 1.05072e-9,
+      "supplyApyPercentage": 2.09335e-9,
```
gleiser-oliveira, line 8: This might be minor than the infinity values, but is the precision loss acceptable here?

src/clients/api/queries/getIsolatedPools/__tests__/__snapshots__/index.spec.ts.snap
therealemjy: Yeah, that's actually normal. That's because JavaScript can't handle values bigger than 1.7976931348623157^308. In some of the tests, calculated APYs reach unrealistically huge values that are higher than this max, so `Infinity` is returned instead.

In real case scenarios, we won't ever have this issue since we have safe guards anyway for values that are too big (`Infinity` would be displayed as > 10 000% when it comes to percentages for exemple).

src/clients/api/queries/getVTokenApySimulations/__tests__/__snapshots__/index.spec.ts.snap
therealemjy: I'd say it is. If you check this snapshot: https://github.com/VenusProtocol/venus-protocol-interface/pull/1591/files/adb1abee8898455248302b33dd335d07effcdbf7#diff-6bc877593c72b804d11cf819f5dc5df18ede9281ed7bdb59e1d4e044b4e3df29R8
We only lose precision after the 11th decimals. I think that's fine for a frontend dApp.

gleiser-oliveira (approved):

PR #1603, opened 2023-10-16
feat: update PrimeStatusBanner to show amount of prime tokens left

gleiser-oliveira:
## Jira ticket(s)

VEN-2058

## Changes

- Condionally displays the amount of prime tokens left above the PrimeStatusBanner, conditions are:
  - User is not prime
  - The amount of prime tokens left is above 0 and below 5% of the total amount

### DIFF
--- a/src/containers/PrimeStatusBanner/PrimeTokensLeft/index.tsx
+++ b/src/containers/PrimeStatusBanner/PrimeTokensLeft/index.tsx
@@ -1,0 +1,26 @@
+import { useTranslation } from 'translation';
+
+interface PrimeTokensLeftProps {
+  tokensLeft: number;
+}
+
+const PrimeTokensLeft = ({ tokensLeft }: PrimeTokensLeftProps) => {
+  const { Trans } = useTranslation();
+  return (
+    <div className="flex flex-row rounded-full border border-yellow bg-[#2E2C2A] px-3 py-1 text-xs font-semibold uppercase lining-nums proportional-nums leading-[15px] tracking-[0.5px] text-yellow [font-variant:all-small-caps]">
+      <p>
+        <Trans
+          i18nKey="primeStatusBanner.tokensLeft"
+          components={{
+            WhiteText: <span className="text-offWhite" />,
+          }}
+          values={{
+            tokensLeft,
+          }}
+        />
+      </p>
+    </div>
+  );
+};
+
+export default PrimeTokensLeft;
--- a/src/containers/PrimeStatusBanner/index.tsx
+++ b/src/containers/PrimeStatusBanner/index.tsx
@@ -144,12 +144,12 @@ export const PrimeStatusBannerUi: React.FC<PrimeStatusBannerUiProps> = ({
       className={cn(
         'relative flex flex-col content-center md:flex-row md:justify-between',
         isUserXvsStakeHighEnoughForPrime && 'items-start md:items-center',
-        shouldShowPrimeTokensLeftIndicator && 'mt-[12.5px]',
+        shouldShowPrimeTokensLeftIndicator && 'mt-3',
         className,
       )}
     >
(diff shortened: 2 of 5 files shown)


src/containers/PrimeStatusBanner/index.tsx
```diff
@@ -137,11 +143,17 @@ export const PrimeStatusBannerUi: React.FC<PrimeStatusBannerUiProps> = ({
...
     <Card
       className={cn(
-        'flex flex-col content-center md:flex-row md:justify-between',
+        'relative flex flex-col content-center md:flex-row md:justify-between',
         isUserXvsStakeHighEnoughForPrime && 'items-start md:items-center',
         className,
       )}
     >
+      {shouldShowPrimeTokensLeftIndicator && (
+        <div className="absolute top-0 -mt-4 max-sm:right-4 sm:left-18 md:left-20">
```
therealemjy: I'd recommend sticking to a mobile-first only approach to avoid confusion:
```suggestion
        <div className="absolute top-0 -mt-4 right-4 sm:right-0 sm:left-18 md:left-20">
```

src/containers/PrimeStatusBanner/PrimeTokensLeft/index.tsx
```diff
@@ -0,0 +1,22 @@
+import { useTranslation } from 'translation';
+
+const PrimeTokensLeft = ({ tokensLeft = 13 }: { tokensLeft: number }) => {
```
therealemjy: I think you forgot to remove the default value:
```suggestion
const PrimeTokensLeft = ({ tokensLeft }: { tokensLeft: number }) => {
```

Also, to stay consistent with the codebase I'd recommend declaring the component prop types through an interface that can then be exported.

therealemjy (commented):
Screen Shot 2023-10-17 at 11 12 31

A few minor things:
- the font size of the text should be 12px (currently 16px) with a line-height of 15px (currently 18px)
- the top and bottom paddings of the container should be 24px on `sm` breakpoint and up (currently 24px on `md` and up)
- in order to compensate for the height of the indicator component, I think we should add a top margin to the `PrimeStatusBanner` component equal to half the height of the indicator component (including the borders) when the indicator is displayed

src/containers/PrimeStatusBanner/index.tsx
gleiser-oliveira: If we do `sm:right-0`, this would mean adding `right: 0` instead of just not having it above a certain breakpoint

Screen Shot 2023-10-17 at 0 34 46

therealemjy: You're right, my bad. It's `sm:right-auto` that can be used to reset the `right` property on `sm` and up:
```suggestion
		<div className="absolute top-0 -mt-4 right-4 sm:right-auto sm:left-18 md:left-20">
```

src/containers/PrimeStatusBanner/index.tsx
```diff
@@ -137,11 +142,18 @@ export const PrimeStatusBannerUi: React.FC<PrimeStatusBannerUiProps> = ({
   return (
     <Card
       className={cn(
-        'flex flex-col content-center md:flex-row md:justify-between',
+        'relative flex flex-col content-center md:flex-row md:justify-between',
         isUserXvsStakeHighEnoughForPrime && 'items-start md:items-center',
+        shouldShowPrimeTokensLeftIndicator && 'mt-[12.5px]',
```
therealemjy: That's a very small thing: given you've applied a negative top margin of 3rem to the div wrapping `PrimeTokensLeft`, you can apply a top margin of 3rem here as well:
```suggestion
        shouldShowPrimeTokensLeftIndicator && 'mt-3',
```

therealemjy (approved):

PR #1604, opened 2023-10-17
feat: show scroll to top button on mobile/tablet

gleiser-oliveira:
## Jira ticket(s)

VEN-2061

## Changes

- On mobile and tablet breakpoints, after scrolling down a screen (100vh), a scroll back to the top button will appear in the lower right corner of the screen

### DIFF
--- a/src/containers/Layout/ScrollToTop/index.tsx
+++ b/src/containers/Layout/ScrollToTop/index.tsx
@@ -1,0 +1,25 @@
+import { ButtonWrapper, Icon } from 'components';
+import { cn } from 'utilities';
+
+import { PAGE_CONTAINER_ID } from 'constants/layout';
+
+import { store } from '../store';
+
+const ScrollToTop = () => {
+  const isVisible = store.use.isScrollToTopVisible();
+  const scrollElem = document.getElementById(PAGE_CONTAINER_ID);
+
+  return (
+    <ButtonWrapper
+      className={cn(
+        'shadow-[0px 4px 15px 0px #0D1017] fixed bottom-16 right-2 h-10 w-10 rounded-full border-0 bg-lightGrey p-0 transition-opacity lg:hidden',
+        isVisible ? 'opacity-1' : 'pointer-events-none opacity-0',
+      )}
+      onClick={() => scrollElem?.scrollTo({ behavior: 'smooth', top: 0 })}
+    >
+      <Icon className="h-3 w-[10px]" name="arrowUpFull" />
+    </ButtonWrapper>
+  );
+};
+
+export default ScrollToTop;
--- a/src/containers/Layout/index.tsx
+++ b/src/containers/Layout/index.tsx
@@ -1,25 +1,52 @@
+import React from 'react';
+
 import { PAGE_CONTAINER_ID } from 'constants/layout';
 
 import { Footer } from './Footer';
 import { Header } from './Header';
 import { Menu } from './Menu';
+import ScrollToTop from './ScrollToTop';
+import { store } from './store';
 
 export interface LayoutProps {
   children?: React.ReactNode;
 }
 
-export const Layout: React.FC<LayoutProps> = ({ children }) => (
-  <div className="flex h-screen flex-col md:flex-row">
-    <Menu />
+export const Layout: React.FC<LayoutProps> = ({ children }) => {
+  const setScrollToTopVisible = store.use.setScrollToTopVisible();
+  const isScrollToTopVisible = store.use.isScrollToTopVisible();
+  const viewportHeight = window.innerHeight;
+  const handleScroll = (event: React.UIEvent<HTMLDivElement>) => {
+    const scrollElem = event.currentTarget;
+    const shouldShowScrollToTopVisible =
+      scrollElem?.scrollTop && scrollElem.scrollTop > viewportHeight;
 
-    <div className="flex flex-1 flex-col overflow-y-auto" id={PAGE_CONTAINER_ID}>
-      <Header />
+    if (shouldShowScrollToTopVisible && !isScrollToTopVisible) {
+      setScrollToTopVisible(true);
+    } else if (!shouldShowScrollToTopVisible && isScrollToTopVisible) {
+      setScrollToTopVisible(false);
+    }
+  };
 
-      <main className="w-full flex-1 px-4 pb-4 md:px-6 xl:mx-auto xl:max-w-[1360px] xl:px-10">
-        {children}
-      </main>
+  return (
+    <div className="flex h-screen flex-col md:flex-row">
+      <Menu />
 
-      <Footer />
+      <div
+        className="flex flex-1 flex-col overflow-y-auto"
+        id={PAGE_CONTAINER_ID}
+        onScroll={handleScroll}
+      >
+        <Header />
+
+        <main className="w-full flex-1 px-4 pb-4 md:px-6 xl:mx-auto xl:max-w-[1360px] xl:px-10">
+          {children}
+        </main>
+
+        <ScrollToTop />
+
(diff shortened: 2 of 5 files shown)


src/theme/index.ts
```diff
@@ -60,6 +60,9 @@ export const theme = {
       },
     ],
   },
+  boxShadow: {
+    scrollToTop: '0px 4px 15px 0px #0D1017',
```
therealemjy: Given this box shadow is specific to the scroll to top button, I wouldn't add it to the theme since it won't be reused. I'd just add the box shadow to the button instance through classes.

src/containers/Layout/ScrollToTop/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+    return () => scrollElem?.removeEventListener('scroll', toggleVisibility);
+  }, []);
+
+  return (
+    isVisible && (
+      <ButtonWrapper
+        className="fixed bottom-16 right-2 h-10 w-10 p-0 rounded-full border-0 bg-lightGrey shadow-scrollToTop lg:hidden"
+        onClick={() => scrollElem?.scrollTo({ behavior: 'smooth', top: 0 })}
+      >
+        <Icon name="arrowUpFull" />
```
therealemjy: According to the designs, the icon should be 12px high

src/containers/Layout/ScrollToTop/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+        setIsVisible(true);
+      } else {
+        setIsVisible(false);
+      }
+    };
+
+    scrollElem?.addEventListener('scroll', toggleVisibility);
+
+    return () => scrollElem?.removeEventListener('scroll', toggleVisibility);
+  }, []);
```
therealemjy: Because `scrollElem` is `null` on mount, you'll need to add it to the list of dependencies of the effect in order for it to work on scroll:
```suggestion
  }, [scrollElem]);
```

src/containers/Layout/ScrollToTop/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    isVisible && (
+      <ButtonWrapper
+        className="fixed bottom-16 right-2 h-10 w-10 p-0 rounded-full border-0 bg-lightGrey shadow-scrollToTop lg:hidden"
+        onClick={() => scrollElem?.scrollTo({ behavior: 'smooth', top: 0 })}
+      >
```
therealemjy: A small design tweak: you could use css to show/hide the button in order to animate its entrance and exit, while using `pointer-events-none` when it's hidden to prevent it from being clickable.

```suggestion
      <ButtonWrapper
      className={cn(
        'shadow-scrollToTop fixed bottom-16 right-2 h-10 w-10 rounded-full border-0 bg-lightGrey p-0 transition-opacity lg:hidden',
        isVisible ? 'opacity-1' : 'pointer-events-none opacity-0',
      )}
      onClick={() => scrollElem?.scrollTo({ behavior: 'smooth', top: 0 })}
    >
```

src/containers/Layout/ScrollToTop/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <ButtonWrapper
+        className={cn(
+          'shadow-[0px 4px 15px 0px #0D1017] fixed bottom-16 right-2 h-10 w-10 rounded-full border-0 bg-lightGrey p-0 transition-opacity lg:hidden',
+          isVisible ? 'opacity-1' : 'pointer-events-none opacity-0',
+        )}
+        onClick={() => scrollElem?.scrollTo({ behavior: 'smooth', top: 0 })}
+      >
+        <Icon className="h-3 w-[10px]" name="arrowUpFull" />
+      </ButtonWrapper>
+    )
```
therealemjy: If you want the animation and since you're disabling the element using css, you should remove the `isVisible` check:
```suggestion
      <ButtonWrapper
        className={cn(
          'shadow-[0px 4px 15px 0px #0D1017] fixed bottom-16 right-2 h-10 w-10 rounded-full border-0 bg-lightGrey p-0 transition-opacity lg:hidden',
          isVisible ? 'opacity-1' : 'pointer-events-none opacity-0',
        )}
        onClick={() => scrollElem?.scrollTo({ behavior: 'smooth', top: 0 })}
      >
        <Icon className="h-3 w-[10px]" name="arrowUpFull" />
      </ButtonWrapper>
```

src/containers/Layout/ScrollToTop/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+        setIsVisible(true);
+      } else {
+        setIsVisible(false);
+      }
+    };
+
+    scrollElem?.addEventListener('scroll', toggleVisibility);
+
+    return () => scrollElem?.removeEventListener('scroll', toggleVisibility);
+  }, [scrollElem]);
```
therealemjy: I thought this would solve the issue with the effect not working, but unfortunately it's still there. The issue is that `scrollElem` is `null` when the component initially renders, and because it never re-renders then the effect does not work.

An alternative solution to the problem would be to detect scroll events on the page container by passing and `onScroll` prop to it. You could then use a zustand store (see example [here](https://github.com/VenusProtocol/venus-protocol-interface/blob/962b991337414b940c39f15a0e4247dd8f6b6b28/src/pages/Dashboard/PrimePromotionalBanner/store.ts)) to store whether the scroll to top button should show or not and listen to changes to that store in scroll to top button itself.
Let me know your thoughts.

src/containers/Layout/store.ts
```diff
@@ -0,0 +1,22 @@
...
+const useStore = create<State>()(
+  persist(
+    set => ({
+      isScrollToTopVisible: false,
+      setScrollToTopVisible: (v: boolean) => set(() => ({ isScrollToTopVisible: v })),
+    }),
+    {
+      name: 'scroll-to-top-visibility',
+    },
+  ),
```
therealemjy: The `persist` function is used to save the store in local storage so that its state persist between sessions. In this case I think we should remove it as this doesn't need to be persisted.
```suggestion
    set => ({
      isScrollToTopVisible: false,
      setScrollToTopVisible: (v: boolean) => set(() => ({ isScrollToTopVisible: v })),
    }),
```

src/containers/Layout/index.tsx
```diff
@@ -1,25 +1,48 @@
...
+  const setScrollToTopVisible = store.use.setScrollToTopVisible();
+  const viewportHeight = window.innerHeight;
+  const handleScroll = (event: React.UIEvent<HTMLDivElement>) => {
+    const scrollElem = event.currentTarget;
+    if (scrollElem?.scrollTop && scrollElem.scrollTop > viewportHeight) {
+      setScrollToTopVisible(true);
+    } else {
+      setScrollToTopVisible(false);
+    }
+  };
```
therealemjy: That's going to trigger a lot of re-renders, because on every scroll event the store state gets updated. You can limit the amount of updates/re-renders by only updating the store's state when needed:
```suggestion
  const isScrollToTopVisible = store.use.isScrollToTopVisible();

  const handleScroll = (event: React.UIEvent<HTMLDivElement>) => {
    const scrollElem = event.currentTarget;
    const shouldShowScrollToTopVisible = scrollElem?.scrollTop && scrollElem.scrollTop > viewportHeight;
    
    if (shouldShowScrollToTopVisible && !isScrollToTopVisible) {
      setScrollToTopVisible(true);
    } else if (!shouldShowScrollToTopVisible && isScrollToTopVisible) {
      setScrollToTopVisible(false);
    }
  };
```

therealemjy (approved):
Looks good to me

PR #1617, opened 2023-10-19
feat: update flow for handling transactions

therealemjy:
The goal of this PR is to update how transactions are sent and tracked so the main UI isn't blocked while a transaction is being confirmed. In this iteration, only the mutations related to supplying, withdrawing, borrowing and repaying have been updated to use the new flow. Other mutations will be updated in the next PR so this one remains manageable.

Demo: https://www.loom.com/share/478db802ed6d462ea16be95613ebbc11?sid=122a074f-80ce-41a7-a1c3-34ee81e94af9

## Changes

- create `useSendTransaction` hook to handle sending and tracking transactions
- update `approveToken` query to use `useSendTransaction`
- update `supply` query to use `useSendTransaction`
- update `swapTokensAndSupply` query to use `useSendTransaction`
- update `repay` query to use `useSendTransaction`
- update `swapTokensAndRepay` query to use `useSendTransaction`
- update `redeem` query to use `useSendTransaction`
- update `redeemUnderlying` query to use `useSendTransaction`
- update `borrow` query to use `useSendTransaction`
- update operation forms following changes to mutations
- fix cache invalidation for main pool query


gleiser-oliveira (approved):
This one is really cool! 🚀

PR #1619, opened 2023-10-19
feat: fetch account Prime status

therealemjy:
## Jira ticket(s)

VEN-2076
VEN-2072

## Changes

- add Prime contract testnet record
- fix issue with `getContractName` and `getUniqueContractAddress` functions where contracts were assumed to exist on all chains
- add `getIsAccountPrime` query function
- remove `isPrime` property from auth context in favor of fetching Prime status of user using query function


src/packages/contracts/utilities/getUniqueContractAddress/index.ts
```diff
@@ -7,5 +7,10 @@ export type GetUniqueContractAddressInput = {
   chainId: ChainId;
 };
 
-export const getUniqueContractAddress = ({ name, chainId }: GetUniqueContractAddressInput) =>
-  addresses[name][chainId];
+export const getUniqueContractAddress = ({ name, chainId }: GetUniqueContractAddressInput) => {
+  const contractAddresses = addresses[name];
+
+  return Object.prototype.hasOwnProperty.call(contractAddresses, chainId)
```
gleiser-oliveira: `hasOwn` [can be used here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn)

```suggestion
  return Object.hasOwn(contractAddresses, chainId)
```

therealemjy: Nice, I didn't know that existed! Thanks

gleiser-oliveira (approved):

PR #1654, opened 2023-10-25
feat: wire up Prime banner fetching data from contract

gleiser-oliveira:
## Jira ticket(s)

VEN-2075

## Changes

- Gets the waiting period from the Prime contract
- Gets the limit of claimable Prime tokens from the Prime contract
- Gets the minimum amount of staked XVS a user has to have to claim a Prime token
- Gets the count of claimed Prime tokens from the Prime contract


src/containers/PrimeStatusBanner/index.tsx
```diff
@@ -317,9 +344,9 @@ const PrimeStatusBanner: React.FC<PrimeStatusBannerProps> = props => {
   return (
     <PrimeStatusBannerUi
       xvs={xvs!}
-      claimedPrimeTokenCount={claimedPrimeTokenCount}
-      primeTokenLimit={primeTokenLimit}
-      primeClaimWaitingPeriodSeconds={primeClaimWaitingPeriodSeconds}
+      claimedPrimeTokenCount={claimedPrimeTokenCount || 0}
+      primeTokenLimit={primeTokenLimit || 0}
+      primeClaimWaitingPeriodSeconds={primeClaimWaitingPeriodSeconds || 0}
```
therealemjy: We shouldn't display any UI until all the data necessary has been fetched, otherwise we're at risk of displaying incorrect temporary states. For example say every finished fetching but `primeClaimWaitingPeriodSeconds` did not, then we might show to users that they can claim their Prime token just before the UI automatically gets updated to say they need to wait 90 days before they can claim it.

To avoid the issue, I'd recommend merging all the query hooks of this PR into one (wrapping all the contract calls with `Promise.all` so they get automatically batched by multicall). Something like `getPrimeStatus`.

src/containers/PrimeStatusBanner/index.tsx
```diff
@@ -299,12 +306,32 @@ const PrimeStatusBanner: React.FC<PrimeStatusBannerProps> = props => {
...
-  const primeClaimWaitingPeriodSeconds = 90 * 24 * 60 * 60; // 90 days in seconds
-  const userStakedXvsTokens = new BigNumber('100');
-  const minXvsToStakeForPrimeTokens = new BigNumber('1000');
+
+  const { data: primeClaimWaitingPeriodSeconds } = useGetPrimeClaimWaitingPeriod();
+
+  const { data: userStakedXvsTokensData } = useGetXvsVaultUserInfo(
+    {
+      accountAddress: accountAddress || '',
+      poolIndex: 1,
```
therealemjy: This will be environment-dependent. It might be worth asking the contract team what the best way for us is to get the index of the vault users will have to stake in for Prime: I think we'd be safer fetching this from a contract (perhaps the contract team could even come up with a new method on the Prime contract for us to check the user stake).

src/containers/PrimeStatusBanner/__tests__/index.spec.tsx
```diff
@@ -17,6 +17,9 @@ describe('PrimeStatusBanner', () => {
         isPrime: true,
       },
     }));
+    (useGetPrimeClaimWaitingPeriod as Vi.Mock).mockImplementation(() => ({
+      data: 600,
+    }));
```
therealemjy: I'd recommend adding some tests to cover the different scenarios, now that we're fetching everything from contracts

src/clients/api/queries/getPrimeStatus/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    primeContract.totalRevocable(),
+    primeContract.revocableLimit(),
+    primeContract.xvsVault(),
+    primeContract.xvsVaultPoolId(),
+    primeContract.xvsVaultRewardToken(),
+    primeContract.claimTimeRemaining(accountAddress),
+  ]);
+
+  return {
+    claimWaitingPeriod: claimWaitingPeriod.toNumber(),
```
therealemjy: To align with the codebase standard, I'd add the unit to the property name:
```suggestion
    claimWaitingPeriodSeconds: claimWaitingPeriod.toNumber(),
```
I assumed it is expressed in seconds, but I'm not entirely sure it is

src/clients/api/queries/getPrimeStatus/index.ts
```diff
@@ -0,0 +1,56 @@
+import BigNumber from 'bignumber.js';
+import { Prime } from 'packages/contracts';
+
+export interface GetPrimeClaimWaitingPeriodInput {
+  accountAddress: string;
+  primeContract: Prime;
+}
+
+export interface GetPrimeClaimWaitingPeriodOutput {
```
therealemjy:
```suggestion
export interface GetPrimeStatusInput {
  accountAddress: string;
  primeContract: Prime;
}

export interface GetPrimeStatusOutput {
```

src/clients/api/queries/getPrimeStatus/useGetPrimeStatus.ts
```diff
@@ -0,0 +1,38 @@
...
+import { GetPrimeClaimWaitingPeriodOutput, getPrimeStatus } from 'clients/api';
+import FunctionKey from 'constants/functionKey';
+import { useIsFeatureEnabled } from 'hooks/useIsFeatureEnabled';
+
+type Options = QueryObserverOptions<
+  GetPrimeClaimWaitingPeriodOutput,
+  Error,
+  GetPrimeClaimWaitingPeriodOutput,
+  GetPrimeClaimWaitingPeriodOutput,
+  FunctionKey.GET_PRIME_STATUS
```
therealemjy: Since we're passing an account address as input, it should be added to the query key as well so it gets cached for this user only:
```suggestion
[FunctionKey.GET_PRIME_STATUS, GetPrimeStatusInput]
```

src/clients/api/queries/getPrimeStatus/useGetPrimeStatus.ts
```diff
@@ -0,0 +1,38 @@
...
+>;
+
+export interface GetPrimeStatusInput {
+  accountAddress: string;
+}
+
+const useGetPrimeClaimWaitingPeriod = (
```
therealemjy:
```suggestion
const useGetPrimeStatus = (
```

src/containers/PrimeStatusBanner/index.tsx
```diff
@@ -297,29 +310,63 @@ const PrimeStatusBanner: React.FC<PrimeStatusBannerProps> = props => {
...
-  const primeTokenLimit = 1000;
-
-  const claimPrimeToken = async () => fakeContractReceipt;
-  const isClaimPrimeTokenLoading = false;
+  const { data: primeStatusData, isLoading: isLoadingPrimeStatus } = useGetPrimeStatus(
+    {
+      accountAddress: accountAddress || '',
+    },
+    {
+      enabled: !isAccountPrime,
```
therealemjy:
```suggestion
      enabled: !isAccountPrime && !!accountAddress,
```

src/containers/PrimeStatusBanner/index.tsx
```diff
@@ -297,29 +310,63 @@ const PrimeStatusBanner: React.FC<PrimeStatusBannerProps> = props => {
...
+  );
+  const { data: userStakedXvsTokensData, isLoading: isLoadingXvsVaultUserInfo } =
+    useGetXvsVaultUserInfo(
+      {
+        accountAddress: accountAddress || '',
+        poolIndex: primeStatusData?.xvsVaultPoolId || 0,
+        rewardTokenAddress: primeStatusData?.rewardTokenAddress || '',
+      },
+      {
+        enabled: !!accountAddress && !!xvs && !!primeStatusData,
```
therealemjy: `xvs` isn't being passed so no need to make sure it exists.
```suggestion
        enabled: !!accountAddress && !!primeStatusData,
```

src/containers/PrimeStatusBanner/index.tsx
```diff
@@ -139,8 +146,9 @@ export const PrimeStatusBannerUi: React.FC<PrimeStatusBannerUiProps> = ({
   const displayClaimButton =
     !haveAllPrimeTokensBeenClaimed &&
     isUserXvsStakeHighEnoughForPrime &&
-    primeClaimWaitingPeriodSeconds <= 0;
+    userPrimeClaimWaitingPeriodSeconds <= 0;
 
+  console.log('kidsgr', displayStakeButton);
```
therealemjy: A small oversight: this should be removed.

src/containers/PrimeStatusBanner/index.tsx
```diff
@@ -297,29 +310,63 @@ const PrimeStatusBanner: React.FC<PrimeStatusBannerProps> = props => {
...
+  });
+
+  const minXvsToStakeForPrimeTokens = convertWeiToTokens({
+    valueWei: primeMinimumStakedXvsMantissa || new BigNumber('0'),
+    token: xvs,
+  });
+
+  const highestPrimeSimulationApyBoostPercentage = new BigNumber('3.14');
+
+  const claimPrimeToken = async () => fakeContractReceipt;
```
therealemjy, line 393: Would you mind adding a `TODO` comment here to remember to wire this function to contracts? Or perhaps you're already working on another PR for it.

gleiser-oliveira: No worries here, it's the next PR in the pipeline

therealemjy: Awesome

therealemjy (approved):
Looks good to me. Nice one!

PR #1657, opened 2023-10-25
refactor: use .hasOwnProperty of object instead of .hasOwn

therealemjy:
## Changes

- use `hasOwnProperty` method of Object instead of `hasOwn` as the former offers better browser compatibility

### DIFF
--- a/src/constants/routing.ts
+++ b/src/constants/routing.ts
@@ -63,7 +63,7 @@ type Routes = {
 
 export const routes = Object.keys(routeSubdirectories).reduce<Routes>(
   (obj, key) =>
-    Object.hasOwn(routeSubdirectories, key)
+    Object.prototype.hasOwnProperty.call(routeSubdirectories, key)
       ? {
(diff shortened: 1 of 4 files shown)


gleiser-oliveira (commented):
@therealemjy do we want to support a specific browser or old version that does not support this yet? I'd say it has been widely adopted by now: 
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn
- https://caniuse.com/mdn-javascript_builtins_object_hasown

therealemjy:
The reason I wanted to make the change is because I opened the dApp on Safari and got a blank page due to hasOwn not being supported. I then checked CanIUse (which you linked) and saw that version 15.3 did not support it, despite being fairly recent (2022).

However, after checking the actual user percentages (https://caniuse.com/usage-table) Safari 15.3 and below has a very low user base. I'll close this PR then.

PR #1696, opened 2023-11-02
feat: add claim Prime token mutation

gleiser-oliveira:
## Jira ticket(s)

VEN-2071

## Changes

- Add claim Prime token mutation
- PrimeStatusBanner handles claiming a Prime token


src/containers/PrimeStatusBanner/index.tsx
```diff
@@ -359,6 +360,17 @@ const PrimeStatusBanner: React.FC<PrimeStatusBannerProps> = props => {
     userStakedXvsTokensData.pendingWithdrawalsTotalAmountWei,
   );
 
+  const { mutateAsync: claimPrimeToken, isLoading: isClaimPrimeTokenLoading } =
+    useClaimPrimeToken();
+
+  const handleClaimPrimeToken = async () => {
+    if (!accountAddress) {
+      throw new VError({ type: 'unexpected', code: 'somethingWentWrong' });
+    }
```
therealemjy: I think you can remove this condition, as you're not passing `accountAddress` to `claimPrimeToken`. The responsibility of defining when `handleClaimPrimeToken` can be called belongs to the component itself then, which won't show the "Claim Prime token" button if the right conditions aren't met.

therealemjy (approved):

PR #1716, opened 2023-11-04
fix: show the eta date for queued proposals

gleiser-oliveira:
## Changes

- This PR fixes queued proposals showing the queued date instead of its eta date


src/pages/Governance/ProposalList/GovernanceProposal/index.tsx
```diff
@@ -113,6 +113,7 @@ interface GovernanceProposalProps {
   endDate?: Date;
   cancelDate?: Date;
   queuedDate?: Date;
```
therealemjy, line 115: This is now unused so we should get rid of it

therealemjy (approved):

PR #1736, opened 2023-11-08
feat: add create proposal feature flag

gleiser-oliveira:
## Jira ticket(s)

VEN-2131

## Changes

- Added a feature flag for creating proposals
- Conditionally render the create proposal CTA and modal depending on the feature flag

### DIFF
--- a/src/pages/Governance/index.spec.tsx
+++ b/src/pages/Governance/index.spec.tsx
@@ -2,6 +2,7 @@ import { fireEvent, waitFor } from '@testing-library/react';
 import BigNumber from 'bignumber.js';
 import _cloneDeep from 'lodash/cloneDeep';
 import React from 'react';
+import { ChainId } from 'types';
 import Vi from 'vitest';
 
 import fakeAccountAddress, { altAddress } from '__mocks__/models/address';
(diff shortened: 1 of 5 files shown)


src/pages/Governance/index.spec.tsx
```diff
@@ -268,4 +285,60 @@ describe('pages/Governance', () => {
...
+      accountAddress: fakeAccountAddress,
+      chainId: ChainId.SEPOLIA,
+    }));
+    const { queryAllByTestId } = renderComponent(<Governance />, {
+      routerOpts: {
+        routerInitialEntries: ['/governance/proposal-create', '/governance'],
+        routePath: '/governance/*',
+      },
+    });
+    expect(queryAllByTestId(TEST_IDS.createProposal)).toHaveLength(0);
```
therealemjy, line 342: Nit: you should be able to retrieve the button using its text:
```tsx
getByText(en.vote.createProposalPlus)
```

therealemjy (approved):

src/pages/Governance/index.spec.tsx
gleiser-oliveira: Using `getByText` or `getByTestId` in the test cases where it won't be rendered will throw an error, so I opted to add an ID to the container and query any instances of it in the document.

PR #1742, opened 2023-11-08
fix: Prime

therealemjy:
## Changes

- fix calculation of hypothetical Prime APYs
- subtract pending withdrawals from amount staked by user in vault shown

### DIFF
--- a/src/clients/api/queries/useGetVaults/useGetVestingVaults/index.ts
+++ b/src/clients/api/queries/useGetVaults/useGetVestingVaults/index.ts
@@ -150,7 +150,9 @@ const useGetVestingVaults = ({
         (acc, _item, poolIndex) => {
           const totalStakedWeiData = poolBalances[poolIndex];
           const lockingPeriodMs = poolData[poolIndex]?.poolInfos.lockingPeriodMs;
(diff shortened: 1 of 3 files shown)


src/clients/api/queries/useGetVaults/useGetVestingVaults/index.ts
```diff
@@ -150,7 +150,9 @@ const useGetVestingVaults = ({
         (acc, _item, poolIndex) => {
           const totalStakedWeiData = poolBalances[poolIndex];
           const lockingPeriodMs = poolData[poolIndex]?.poolInfos.lockingPeriodMs;
-          const userStakedWei = poolData[poolIndex]?.userInfos?.stakedAmountWei;
+          const userStakedWei = poolData[poolIndex]?.userInfos?.stakedAmountWei.minus(
```
gleiser-oliveira, line 153: Question: from a UX perspective, wouldn't this change potentially confuse users? It might be a bigger discussion, but I'd add the amount of pending withdrawals to the XVS (if greater than 0) vault card along with this change

gleiser-oliveira (approved):

PR #1745, opened 2023-11-08
feat: identify chainId on PostHog

gleiser-oliveira:
## Jira ticket(s)

VEN-2136

## Changes

- Adds `chainId` when identifying a user on PostHog

### DIFF
--- a/src/packages/analytics/index.tsx
+++ b/src/packages/analytics/index.tsx
@@ -6,17 +6,17 @@ import { version as APP_VERSION } from 'constants/version';
 import { useAuth } from 'context/AuthContext';
 
 const UserIdentifier: React.FC = () => {
-  const { accountAddress } = useAuth();
+  const { accountAddress, chainId } = useAuth();
   const posthog = usePostHog();
 
   // Identify user by their account address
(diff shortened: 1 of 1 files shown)


src/packages/analytics/index.tsx
```diff
@@ -6,13 +6,13 @@ import { version as APP_VERSION } from 'constants/version';
 import { useAuth } from 'context/AuthContext';
 
 const UserIdentifier: React.FC = () => {
-  const { accountAddress } = useAuth();
+  const { accountAddress, chainId } = useAuth();
   const posthog = usePostHog();
 
   // Identify user by their account address
   useEffect(() => {
```
therealemjy, line 13: You'll need to pass `chainId` to the list of dependencies so that the effect triggers when it changes

therealemjy (changes requested):

src/packages/analytics/index.tsx
gleiser-oliveira: Oops, fixed it

therealemjy (approved):

PR #1750, opened 2023-11-08
feat: add chainId tag to Sentry

gleiser-oliveira:
## Jira ticket(s)

VEN-2135

## Changes

- Adds a `chainId` tag to Sentry. The main benefits of tags are that they are searchable and appear in Sentry's UI much like the `environment` where an event has happened

More info can be found here: https://docs.sentry.io/platforms/javascript/enriching-events/tags/

### DIFF
--- a/src/context/AuthContext/index.tsx
+++ b/src/context/AuthContext/index.tsx
@@ -1,4 +1,5 @@
 import { openInfinityWallet } from '@infinitywallet/infinity-connector';
+import * as Sentry from '@sentry/react';
 import { VError, displayMutationError } from 'errors';
 import { Signer, getDefaultProvider } from 'ethers';
 import noop from 'noop-ts';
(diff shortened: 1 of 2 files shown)


src/context/AuthContext/index.tsx
```diff
@@ -47,6 +48,7 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
 
   const defaultChainId = chains[0].id;
   const chainId = chain?.id || defaultChainId;
+  Sentry.setTag('chainId', chainId);
```
therealemjy: This will be triggered every time the component renders, not just when `chainId` changes.
A quick solution for now would be to wrap the call in an effect and set `chainId` as its dependency.

However, another approach could be to make a separate error package, which would export an `ErrorProvider` that renders the Sentry provider as well as the effect for automatically updating the chainId in Sentry (I think we should [add the account address](https://docs.sentry.io/platforms/javascript/session-replay/configuration/) as well). We'll then be able to move all the functions and hooks we currently have in the error folder into this error package and keep everything related to errors in it.

Let me know your thoughts though, since that's more work.

therealemjy (approved):

PR #1761, opened 2023-11-10
fix: add chainId to the dependencies of memoized menuItems

gleiser-oliveira:
## Changes

- Fix for the menuItems list not getting refreshed when the chainId changes

### DIFF
--- a/src/containers/Layout/useGetMenuItems.tsx
+++ b/src/containers/Layout/useGetMenuItems.tsx
@@ -126,6 +126,7 @@ const useGetMenuItems = () => {
     accountAddress,
     convertVrtRouteEnabled,
     corePoolRouteEnabled,
+    swapRouteEnabled,
     historyRouteEnabled,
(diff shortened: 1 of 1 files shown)


src/containers/Layout/useGetMenuItems.tsx
```diff
@@ -118,7 +118,7 @@ const useGetMenuItems = () => {
     }
 
     return menuItems;
-  }, [accountAddress]);
+  }, [accountAddress, chainId]);
```
therealemjy: I think the actual dependencies are `corePoolRouteEnabled`, `historyRouteEnabled`, `convertVrtRouteEnabled` and `vaiRouteEnabled`. Technically they do rely on `chainId`, but there could be instances where even if the chain changes these don't (for example if we were to switch from BSC mainnet to BSC testnet).

therealemjy (approved):

PR #1764, opened 2023-11-10
refactor: add eslint rule to check hook dependencies

therealemjy:
## Changes

- add eslint rule to check hook dependencies
- fix most hook dependencies: I left the warnings related to function dependencies that change often and styles, which also change often and will disappear as we move towards using Tailwind instead of styled components


gleiser-oliveira (approved):
Nice addition! Impressive how many of those slip through silently

therealemjy:
Yeah, to be honest hooks aren't the most straight forward thing to work with, despite how powerful they are :/

PR #1781, opened 2023-11-14
feat: use voteProposal to conditionally render delegate section

gleiser-oliveira:
## Jira ticket(s)

VEN-2119

## Changes

- The `voteProposal` feature flag now controls if the user sees the delegate votes section, along with a warning about votes being disabled in the `VotingWallet` component


therealemjy (commented):
It looks good.
I've realized we are missing a piece of UX that should have been described in the ticket and the designs (sorry for that): the "delegate" feature should be disabled when the `voteProposal` feature is disabled.

To reflect that, can you prevent rendering the "Delegate" (or "Redelegate") button  (as well as the modal associated with it) when the `voteProposal` feature is disabled please?
Screen Shot 2023-11-14 at 10 26 08

therealemjy (approved):

therealemjy (changes requested):
It looks good, however there's one small design issue: when the `voteProposal` feature is not enabled we end up with additional blank space at the bottom of the `VotingWallet` component container:

Screen Shot 2023-11-15 at 09 18 08

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -57,6 +61,8 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
   delegateModelIsOpen,
   setDelegateModelIsOpen,
 }) => {
+  const { switchChain } = useAuth();
+  const voteProposalFeatureEnabled = useIsFeatureEnabled({ name: 'voteProposal' });
   const { t, Trans } = useTranslation();
   const styles = useStyles();
```
therealemjy: Let's get rid of styled components completely here: you can replace the existing styles with Tailwind classes, delete the `styles.ts` file completely and remove the first line of this file (`/** @jsxImportSource @emotion/react */`).

therealemjy (commented):
Nice, one step further towards converting the codebase to Tailwind 🥳
I think we can go even further and convert the `VotingWallet` component entirely to Tailwind.

Beside that, I spotted a minor bug. When launching the dApp and while the voting weight of the user is loading, the "Deposit XVS" button shows in the UI without any top margin.
Screen Shot 2023-11-21 at 10 10 29

A solution can be to just hide it while any of the initial queries are loading.

src/theme/index.ts
```diff
@@ -26,6 +26,7 @@ export const theme = {
     // The next colors are not in the UI kit, but are used throughout the designs
     mediumBlue: '#265ACC',
     darkBlue: '#1B4398',
+    delimiter: '#21293A',
```
therealemjy: This color is only used for one element in the `VotingWallet` component, so I wouldn't add it to the theme.
Instead I'd use its hex code as arbitrary value for Tailwind classes: `border-r-[#21293A]`

therealemjy (commented):
We're getting very close. There's only two minor issues I spotted:

1) The UI on tablet has some slight discrepancies with the designs. The item in the middle is slightly pushed to the bottom. Also, the first two cells ("Voting weight" and "Total locked" should be stuck to each other with a fixed spacing in between them):
Screen Shot 2023-11-22 at 11 01 55

2) We still have extra space at the bottom while the UI is loading:
Screen Shot 2023-11-22 at 11 08 46

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -84,47 +89,69 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
+  const showDelegateButton =
+    !isDataLoading && connectedWallet && userHasLockedXVS && voteProposalFeatureEnabled;
+  const shouldApplyMarginToTotalLocked =
+    voteProposalFeatureEnabled || !connectedWallet || showDepositXvs;
+
   return (
-    <div css={styles.root}>
-      <Typography variant="h4">{t('vote.votingWallet')}</Typography>
+    <div className="flex flex-1 flex-col lg:ml-4">
+      <h4 className="m-0 text-lg font-semibold leading-normal">{t('vote.votingWallet')}</h4>
```
therealemjy: `leading-normal` corresponds to the line-height that's already applied to all text elements, so you don't need to add it here

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -84,47 +89,69 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
+          }
+        />
+      )}
 
-      <Paper css={styles.votingWalletPaper}>
-        <div css={styles.votingWeightContainer}>
-          <Typography variant="body2" css={styles.subtitle} className="text-grey">
+      <Card className="mt-6 flex flex-col items-start justify-start p-6 sm:flex-row sm:items-center sm:justify-between lg:flex-col lg:items-start">
+        <div className="border-r-delimiter pb-4 sm:border-r sm:pb-0 sm:pr-10 lg:border-r-0 lg:pb-4">
+          <p className="m-0 text-base font-semibold leading-normal tracking-[0.3px] text-grey sm:text-sm md:text-base">
```
therealemjy: No need for the  `leading-normal` and `m-0` classes. Also, I'd remove the `tracking-[0.3px]` props (the designs don't seem to include it, and I'd rather keep texts consistent).

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -84,47 +89,69 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
-        <div css={styles.totalLockedContainer}>
-          <div css={styles.totalLockedTitle}>
-            <Typography
-              variant="body2"
-              className="text-grey"
-              css={[styles.subtitle, styles.totalLockedText]}
-            >
+        <div className={cn('mt-4', shouldApplyMarginToTotalLocked && 'mb-5 sm:mb-0 lg:mb-5')}>
+          <div className="mb-1 flex flex-row items-end">
+            <p className="m-0 mr-2 text-base font-semibold leading-normal tracking-[0.3px] text-grey sm:text-sm md:text-base">
```
therealemjy: No need for the  `leading-normal` class

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -134,81 +161,86 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
+      </Card>
 
-      <Paper css={[styles.votingWalletPaper, styles.voteSection]}>
-        <Typography variant="body2" color="textPrimary" css={styles.toVote}>
-          {t('vote.toVoteYouShould')}
-        </Typography>
+      {voteProposalFeatureEnabled && (
+        <>
+          <Card className="mt-6 flex flex-col items-start justify-start p-6 lg:justify-between">
+            <p className="mb-4 font-semibold leading-normal tracking-[0.3px]" color="textPrimary">
```
therealemjy: No need for the  `leading-normal` and `tracking-[0.3px]` classes. The `color` prop should also be removed as it's not relevant to a `p` tag.

src/pages/Governance/VotingWallet/testIds.ts
```diff
@@ -1,6 +1,8 @@
 export default {
   votingWeightValue: 'vote-voting-wallet-voting-weight-value',
   totalLockedValue: 'vote-voting-wallet-total-locked-value',
+  delegateButton: 'delegate-button',
   delegateYourVoting: 'vote-voting-wallet-delegate-your-voting',
   depositYourTokens: 'vote-voting-wallet-deposit-your-tokens',
+  votingDisabledWarning: 'voting-disabled-warning',
```
therealemjy: I recommend adding the same prefixes as other test IDs as they need to be unique within the entire app:
```suggestion
  delegateButton: 'vote-voting-wallet-delegate-button',
  delegateYourVoting: 'vote-voting-wallet-delegate-your-voting',
  depositYourTokens: 'vote-voting-wallet-deposit-your-tokens',
  votingDisabledWarning: 'vote-voting-wallet-voting-disabled-warning',
```

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -134,81 +161,86 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
-            }}
-          />
-        </Typography>
+            <span className="mb-3 text-sm" color="textPrimary">
+              <Trans
+                i18nKey="vote.depositYourTokens"
+                components={{
+                  Link: (
+                    <Link
+                      className="text-blue hover:cursor-pointer hover:underline"
```
therealemjy: That's the default style of the `Link` component, so you don't need any of these classes

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -134,81 +161,86 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
-      </Paper>
+      </Card>
 
-      <Paper css={[styles.votingWalletPaper, styles.voteSection]}>
-        <Typography variant="body2" color="textPrimary" css={styles.toVote}>
-          {t('vote.toVoteYouShould')}
-        </Typography>
+      {voteProposalFeatureEnabled && (
+        <>
+          <Card className="mt-6 flex flex-col items-start justify-start p-6 lg:justify-between">
```
therealemjy: `items-start` and `justify-start` are the default values, so you don't need these classes here

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -84,47 +89,69 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
 
-        <div css={styles.totalLockedContainer}>
-          <div css={styles.totalLockedTitle}>
-            <Typography
-              variant="body2"
-              className="text-grey"
-              css={[styles.subtitle, styles.totalLockedText]}
-            >
+        <div className={cn('mt-4', shouldApplyMarginToTotalLocked && 'mb-5 sm:mb-0 lg:mb-5')}>
+          <div className="mb-1 flex flex-row items-end">
```
therealemjy: `flex-row` is the default value, so you don't need this class here

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -84,47 +89,69 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
+            className="m-0 text-xl sm:text-lg md:text-xl"
+            data-testid={TEST_IDS.votingWeightValue}
+          >
             {readableVoteWeight}
-          </Typography>
+          </h3>
         </div>
 
-        <Delimiter css={styles.delimiter} />
+        <Delimiter className="block w-full sm:hidden lg:block" />
```
therealemjy: `block` is the default value for this component, so you don't need this class here (you still need `lg:block`)

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -84,47 +89,69 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
+      <Card className="mt-6 flex flex-col items-start justify-start p-6 sm:flex-row sm:items-center sm:justify-between lg:flex-col lg:items-start">
+        <div className="border-r-delimiter pb-4 sm:border-r sm:pb-0 sm:pr-10 lg:border-r-0 lg:pb-4">
+          <p className="m-0 text-base font-semibold leading-normal tracking-[0.3px] text-grey sm:text-sm md:text-base">
             {t('vote.votingWeight')}
-          </Typography>
+          </p>
 
-          <Typography variant="h3" css={styles.value} data-testid={TEST_IDS.votingWeightValue}>
+          <h3
+            className="m-0 text-xl sm:text-lg md:text-xl"
```
therealemjy: Titles don't have any margins by default with Tailwind, so you can remove the `m-0` class here

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -134,81 +161,86 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
-                <Link
-                  to={routes.vaults.path}
-                  css={styles.clickableText}
-                  data-testid={TEST_IDS.depositYourTokens}
-                />
-              ),
-            }}
-          />
-        </Typography>
+            <span className="mb-3 text-sm" color="textPrimary">
```
therealemjy: No need for the `color` prop

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -134,81 +161,86 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
-                  tabIndex={0}
-                  onClick={() => setDelegateModelIsOpen(true)}
-                  data-testid={TEST_IDS.delegateYourVoting}
-                />
-              ),
-            }}
-          />
-        </Typography>
-      </Paper>
+            <span className="text-sm" color="textPrimary">
```
therealemjy: No need for the `color` prop

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -84,47 +89,69 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
+              {t('vote.multichain.switchToBsc')}
+            </Button>
+          }
+        />
+      )}
 
-      <Paper css={styles.votingWalletPaper}>
-        <div css={styles.votingWeightContainer}>
-          <Typography variant="body2" css={styles.subtitle} className="text-grey">
+      <Card className="mt-6 flex flex-col items-start justify-start p-6 sm:flex-row sm:items-center sm:justify-between lg:flex-col lg:items-start">
```
therealemjy: No need for the `items-start` and `justify-start` classes (these are the default CSS values)

src/pages/Governance/VotingWallet/index.tsx
```diff
@@ -84,47 +89,69 @@ export const VotingWalletUi: React.FC<VotingWalletUiProps> = ({
...
           </div>
 
-          <div css={styles.totalLockedValue}>
-            {xvs && <TokenIcon token={xvs} css={styles.tokenIcon} />}
+          <div className={cn('flex flex-row items-center', voteProposalFeatureEnabled && 'pb-2')}>
+            {xvs && <TokenIcon className="mr-3 h-[26px] w-[26px]" token={xvs} />}
 
-            <Typography variant="h3" css={styles.value} data-testid={TEST_IDS.totalLockedValue}>
+            <h3
+              className="m-0 text-xl sm:text-lg md:text-xl"
```
therealemjy: No need for the `m-0` class

src/pages/Governance/VotingWallet/index.tsx
gleiser-oliveira: Just to be sure: `tracking-[0.3px]` - `letter-spacing: 0.3px` is currently set in this element in the prod app. As you said, this is not present in the designs, so I removed it, although the final result is slightly different

src/pages/Governance/VotingWallet/index.tsx
gleiser-oliveira: Same as https://github.com/VenusProtocol/venus-protocol-interface/pull/1781#discussion_r1403217341

src/pages/Governance/VotingWallet/index.tsx
therealemjy: Ah I see, makes sense. Let's follow the designs anyway and remove these (as you did).

src/packages/analytics/useAnalytics.ts
gleiser-oliveira, line 1: Updated due to `yarn format`

therealemjy (approved):
Nice. Thanks for hanging in there: the result is very clean. Well done 👌

PR #1804, opened 2023-11-16
feat: enable Prime feature on BSC_MAINNET

gleiser-oliveira:
## Changes

- Enable Prime feature flag on BSC_MAINNET


therealemjy (approved):
You'll just need to update the test of `useIsFeatureEnabled`, otherwise it looks good to me

PR #1820, opened 2023-11-20
feat: add error package

gleiser-oliveira:
## Jira ticket(s)

VEN-2169

## Changes

- Created error package to contain all logic related to error reporting
- We now tag Sentry events with the `accountAddress` in addition to the `chainId`
- Tagging is done in a new component: `SentryEventTagger`


src/packages/errors/index.ts
```diff
@@ -0,0 +1,20 @@
...
+import React, { useEffect } from 'react';
+
+import { useAuth } from 'context/AuthContext';
+
+export const SentryEventTagger: React.FC = () => {
+  const { accountAddress, chainId } = useAuth();
+
+  // Tags Sentry events with the connected address and the chain ID being used
+  useEffect(() => {
+    setTags({ accountAddress, chainId });
```
therealemjy: For the account address, I think we should use the `setUser` method from Sentry instead of `setTags` and use the account address as user ID, as Sentry as specific features around user identity.

src/packages/errors/index.ts
```diff
@@ -0,0 +1,20 @@
+import { setTags } from '@sentry/react';
+import React, { useEffect } from 'react';
+
+import { useAuth } from 'context/AuthContext';
+
+export const SentryEventTagger: React.FC = () => {
```
therealemjy: I like the approach and I'd take it a step further.
Because Sentry requires to wrap the app with an error boundary component ([`<Sentry.ErrorBoundary/>`](src/App/index.tsx)), I'd transform `SentryEventTagger` into a wrapper that wraps its children within the Sentry error boundary (and does what it currently does as well). I'd also move the logic for initializing [Sentry](src/initializeLibraries/sentry.ts) into this wrapper as well, so all the logic for Sentry is contained within the `errors` package. That's the approach I took with the `analytics` [package](src/packages/analytics/index.tsx).

src/App/SentryErrorInfo/index.tsx
gleiser-oliveira, line 1: This component is outside the errors package to avoid a cyclic dependency with `context/AuthContext`

gleiser-oliveira, line 1: This component is outside the errors package to avoid a cyclic dependency with context/AuthContext

therealemjy: How about including it in the `errors` package but not export it from its index file?
```tsx
import { SentryErrorInfo } from 'packages/errors/SentryErrorInfo';
```

The last unaddressed comment on the PR then is the wrapper:
Screen Shot 2023-11-21 at 10 06 57

Sentry is initialized through the logic at `src/initializeLibraries/sentry.ts`, which is the logic I'd move inside the mentioned wrapper (as as well as Sentry error boundary component).

src/App/SentryErrorInfo/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+    if (accountAddress) {
+      setUser({
+        id: accountAddress,
+      });
+    } else {
+      setUser(null);
+    }
+
+    setTag('chainId', chainId);
+  }, [accountAddress, chainId]);
```
therealemjy: I'd separate this effect into two: one for the account address, one for the chain ID. Otherwise both the user and the chainId tag will be updated whenever one of them changes.

therealemjy:
One last thing: in the root README file, we list the different packages and their purpose. I think it would be nice to add a line there for the `errors` package.

therealemjy (approved):
Really nice 💪

PR #1844, opened 2023-11-23
feat: close modal after claiming rewards

therealemjy:
## Changes

- close modal after claiming rewards

### DIFF
--- a/src/containers/Layout/ClaimRewardButton/__tests__/__snapshots__/index.spec.tsx.snap
+++ b/src/containers/Layout/ClaimRewardButton/__tests__/__snapshots__/index.spec.tsx.snap
@@ -42,7 +42,7 @@ exports[`ClaimRewardButton > it claims only selected rewards on submit button cl
 }
 `;
 
-exports[`ClaimRewardButton > it claims reward on submit button click on success 1`] = `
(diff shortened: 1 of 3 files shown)


src/containers/Layout/ClaimRewardButton/__tests__/__snapshots__/index.spec.tsx.snap
```diff
@@ -42,6 +42,59 @@ exports[`ClaimRewardButton > it claims only selected rewards on submit button cl
...
+      "vTokenAddressesWithPendingReward": [
+        "0xcfc8a73f9c888eea9af9ccca24646e84a915510b",
+        "0x12d3a3aa7f4917ea3b8ee34f99a9a7eec521fa61",
+      ],
+    },
+  ],
+}
+`;
+
 exports[`ClaimRewardButton > it claims reward on submit button click on success 1`] = `
```
gleiser-oliveira: It looks like the snapshot for the test that was renamed is still here, when it should have been removed/replaced

therealemjy: Good catch

gleiser-oliveira (approved):

PR #1845, opened 2023-11-23
refactor: remove auth context

therealemjy:
## Jira ticket(s)

VEN-2143

## Changes

- remove auth context in favor of using hooks from `wallet` package
- add mocks for `wallet` package methods
- update `wallet` package methods to make them return objects so they can be easily extended
- remove nesting in options of test render functions
- add eslint rule to detect unused imports. This means unused imports can be automatically removed by running the command `yarn lint:ts --fix` now


src/packages/__mocks__/index.ts
gleiser-oliveira, line 1: File can be removed

gleiser-oliveira (approved):

PR #1869, opened 2023-11-28
refactor: adjust contract config for new Venus packages

gleiser-oliveira:
## Changes

- Bumped all Venus' packages
- Adjusted contract config to use new addresses files


src/packages/contracts/config/index.ts
```diff
@@ -37,6 +38,10 @@ import pancakePairV2Abi from './externalAbis/PancakePairV2.json';
 import vBnbAbi from './externalAbis/VBnb.json';
 import XsequenceMulticall from './externalAbis/XsequenceMulticall.json';
 
+// bscmainnet.json file is too big and TS does not parse its type properly
+const venusProtocolBscMainnetDeployments =
+  venusProtocolBscMainnetDeploymentsJson as typeof venusProtocolBscTestnetDeployments;
```
therealemjy: The issue is that the bigger this document get, the more processing power will be used during the build process and so we might reach a bottleneck.
I think it's a good time to refactor this to use a library that would read the JSON file in chunks (e.g.: https://www.npmjs.com/package/JSONStream). Let me know if you want to collaborate on that.

gleiser-oliveira: Ok, I'll be looking into it!

therealemjy (approved):

PR #1903, opened 2023-12-05
feat: estimate Prime APYs with averages from Prime users

gleiser-oliveira:
## Jira ticket(s)

VEN-2215

## Changes

- Estimate Prime APYs with averages from Prime users

### DIFF
--- a/src/clients/api/queries/getLegacyPool/appendPrimeSimulationDistributions.ts
+++ b/src/clients/api/queries/getLegacyPool/appendPrimeSimulationDistributions.ts
@@ -19,6 +19,20 @@ export interface ResolvePrimeSimulationDistributionsInput {
   accountAddress?: string;
 }
 
+const supplyAveragesForToken: Record<string, BigNumber> = {
+  BTCB: new BigNumber('0.71'),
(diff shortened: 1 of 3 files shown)


src/clients/api/queries/getLegacyPool/appendPrimeSimulationDistributions.ts
```diff
@@ -19,6 +19,27 @@ export interface ResolvePrimeSimulationDistributionsInput {
...
+  USDT: new BigNumber('10009.21'),
+  USDC: new BigNumber('2405.43'),
+} as const;
+
+const xvsStakedAveragesForToken: Record<string, BigNumber> = {
+  BTCB: new BigNumber('4124.59'),
+  ETH: new BigNumber('4788.05'),
+  USDT: new BigNumber('3731.33'),
+  USDC: new BigNumber('3265.30'),
+} as const;
```
therealemjy: You can remove `as const` for each variable: it does nothing here since you're declaring the type explicitly.

therealemjy (approved):

PR #1918, opened 2023-12-07
feat: add Binance wallet using Wagmi

gleiser-oliveira:
## Jira ticket(s)

VEN-2218

## Changes

- Add Binance wallet using Wagmi
- Add plugin to polyfill Buffer on esbuild
- Removed Binance Chain wallet migration warning


vite.config.ts
```diff
@@ -14,6 +15,20 @@ export default defineConfig(({ mode }) => {
 
   return {
     plugins: [react(), viteTsconfigPaths(), svgrPlugin()],
+    optimizeDeps: {
+      esbuildOptions: {
+        // Node.js global to browser globalThis
+        define: {
+          global: 'globalThis'
```
therealemjy, line 22: Why exactly do you need this? It might be a bit risky as other libraries might rely on the `global` object

gleiser-oliveira: This is needed by the polyfill I added on line 26, `NodeGlobalsPolyfillPlugin`. Without defining `global`, the app crashes with this error:
Screenshot 2023-12-11 at 4 06 36

I'm trying other alternatives, but no luck so far

therealemjy: Ok I see. Let's go with it then

therealemjy (approved):

PR #1941, opened 2023-12-12
feat: add ChainID to URL

therealemjy:
## Jira ticket(s)

VEN-2234

## Changes

- update routing to use nested routes
- update `Link` component to automatically set `chainId` parameter in URL search
- add `useFormatTo` hook to format a `to` parameter to automatically set `chainId` parameter in its search field
- add `useNavigate` hook to automatically set `chainId` parameter in target URL search
- update `Web3Wrapper` to handle syncing `chainId` parameter in URL search with connected wallet
- refactor `useUrlPagination` to use `useSearchParams` hook from React Router


src/hooks/useUrlPagination/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    // Add page param to URL if none has been set
+    if (!pageIndex) {
+      // Note: although the pagination starts from 0, we make it start from 1 in the
+      // URL to make it more user-friendly. This is something we need to account for
+      // when updating the page search param
+      setPageIndex(1);
+    }
+
+    // Scroll to the top of the page on search change
+    window.scrollTo(0, 0);
```
coreyar: Search param changes aren't caught by the root scroll to the top [here](https://github.com/VenusProtocol/venus-protocol-interface/pull/1941/files#diff-b226a971b851720f6b199049faf0b035d75c99ff449f9981070491ce7033a76fR43)

coreyar (approved):

src/hooks/useUrlPagination/index.ts
therealemjy: Good point. We can actually remove this logic, as React Router v6 scrolls to the top of the page on param change by default

PR #1953, opened 2023-12-14
fix: multichain-related bugs

therealemjy:
## JIRA ticket(s)

VEN-2249

## Changes

- determine explorer domain from URL
- add Multicall3 Sepolia record
- update chain names
- hide vote CTAs when vote feature is disabled
- remove extra slash at the end of some explorer URLs

### DIFF
--- a/src/packages/contracts/config/index.ts
+++ b/src/packages/contracts/config/index.ts
@@ -193,6 +193,7 @@ export const contracts: ContractConfig[] = [
     address: {
       [ChainId.BSC_TESTNET]: '0xca11bde05977b3631167028862be2a173976ca11',
       [ChainId.BSC_MAINNET]: '0xca11bde05977b3631167028862be2a173976ca11',
+      [ChainId.SEPOLIA]: '0xca11bde05977b3631167028862be2a173976ca11',
     },
   },
   {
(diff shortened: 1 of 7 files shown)


src/packages/contracts/config/index.ts
```diff
@@ -193,6 +193,7 @@ export const contracts: ContractConfig[] = [
     address: {
       [ChainId.BSC_TESTNET]: '0xca11bde05977b3631167028862be2a173976ca11',
       [ChainId.BSC_MAINNET]: '0xca11bde05977b3631167028862be2a173976ca11',
+      [ChainId.SEPOLIA]: '0xca11bde05977b3631167028862be2a173976ca11',
```
chechu, line 196: Would it make sense to add the ETHEREUM entry now, to avoid future unexpected behaviors? AFAIK it's the same address

chechu (approved):

src/packages/contracts/config/index.ts
therealemjy: We have a separate ticket for this: I'd rather add the records for all the contracts at once when we get to it (the idea is that the system needs to be modular enough so that certain contracts can exist on some chains but not others, hence the partitioning).

chechu (approved):

PR #1968, opened 2023-12-18
feat: Bridge page UI

therealemjy:
## Jira ticket(s)

VEN-2246

## Changes

- wrap `Select`,`TextField` and `TokenTextField` components with `forwardRef` function from React so that refs can be forwarded to them
- add UI for Bridge page
- add option to display "NEW" badge on navigation menu items
- update `useGetTokens` and `useGetToken` hooks so they can accept an optional `chainId` parameter
- add "bridge" icon


package.json
```diff
@@ -32,14 +32,15 @@
   },
   "dependencies": {
     "@0xsequence/multicall": "^1.2.9",
-    "@binance/w3w-wagmi-connector": "^1.0.1",
     "@binance-chain/bsc-connector": "^1.0.0",
+    "@binance/w3w-wagmi-connector": "^1.0.1",
```
gleiser-oliveira: Small conflict here due to the package update

src/pages/Bridge/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+    setValue('toChainId', formValues.fromChainId);
+  }, [getValues, setValue]);
+
+  const onSubmit = (values: z.infer<typeof formSchema>) => {
+    // TODO: send transaction (see VEN-2248)
+    console.log(values);
+  };
+
+  // TODO: wire up
+  const walletBalanceTokens = new BigNumber(10);
```
gleiser-oliveira, line 146: @therealemjy I've a question, I'm wondering how this will work if the user changes the "from" chain. Shouldn't "from" be locked to the chain you are connected to?

gleiser-oliveira: Unless we can get the balance the user has on each chain no matter which one they are connected to

therealemjy: Yeah that's a good point. Technically we can fetch the balance on each chain (the app is setup so a provider is given for each chain).
However I think you're right and we should still lock the "from" chain, as users won't be able to send a transaction anyway if the from chain doesn't correspond to the one their wallet is connected to. I'll discuss this with Jesús and Pavel.

therealemjy: I spoke with Jesús and Pavel and the decision taken is to link the "from" chain select to the wallet chain. This means changing the chain will also change the wallet chain, and vice versa.

gleiser-oliveira (approved):

PR #2000, opened 2023-12-28
feat: wire up Prime calculator

gleiser-oliveira:
## Jira ticket(s)

VEN-2222

## Changes

- Bump @venusprotocol/venus-protocol to 7.0.0
- Updated return signature for primeContract.estimateAPR
- Use new info from estimateAPR to wire up the Prime calculator


src/clients/api/queries/getLegacyPool/useGetLegacyPool.ts
```diff
@@ -77,7 +78,7 @@ const useGetLegacyPool = (input?: TrimmedInput, options?: Options) => {
     (options?.enabled === undefined || options?.enabled);
 
   return useQuery(
-    [FunctionKey.GET_LEGACY_POOL, { ...input, chainId }],
+    options?.queryKey ? options.queryKey : [FunctionKey.GET_LEGACY_POOL, { ...input, chainId }],
```
coreyar, line 81: It seems odd to me that the query key would need to change. Why not pass the accountAddress and use that to create they query key?

src/clients/api/queries/useGetPrimeEstimation/index.ts
```diff
@@ -0,0 +1,212 @@
+import BigNumber from 'bignumber.js';
```
coreyar, line 1: This file is wrapping multiple queries like `useGetPrimeDistributionForMarket` `useGetHypotheticalPrimeApys`. It is not making a unique query and contains a lot of extra logic. 

What about organizing this in under hooks?

src/clients/api/queries/getLegacyPool/useGetLegacyPool.ts
gleiser-oliveira: The thing here is that this hook is also used by the ClaimRewards button, which refreshes the key with accountAddress every few seconds. This means that the calculator also gets refreshed values, re-rendering itself. So the intention here was to have a designated key for the calculator alone, so it wouldn't use the same as ClaimRewards.

src/clients/api/queries/useGetPrimeEstimation/index.ts
```diff
@@ -0,0 +1,212 @@
...
+        supplyCapMantissa,
+        supplyCapUsd: supplyCapPriceMantissa,
+        userPrimeRewardsShare,
+      } = hypotheticalPrimeApysData;
+
+      const borrowedTokens = new BigNumber(borrowedAmountTokens);
+
+      const suppliedTokens = new BigNumber(suppliedAmountTokens);
+
+      const supplyCapUsd = convertPriceMantissaToDollars({
```
coreyar: I don't think there should be functions that convert from "calculation" units (units that are used for calculation but not displayed) to doing a calculation and converting to "display" units.
Calculation functions should return "calculation units" and then we can convert calculation to display units. In this way you remove the need for the function on line 161 and have clear separation of responsibilities

coreyar: In this case the value returned is in USD so the order should be clean incoming values to cents in `useGetHypotheticalPrimeApys` first. And then consumers of that query have the values in the correct unit to do calculations with until it is time to display a value

src/pages/PrimeCalculator/Form/index.tsx
```diff
@@ -112,13 +180,94 @@ export const Form: React.FC = () => {
...
+  }, [
+    initialStakedAmountXvsTokens,
+    initialBorrowAmountTokens,
+    initialSupplyAmountTokens,
+    formState.touchedFields.stakedAmountXvsTokens,
+    setValue,
+    reset,
+  ]);
+
+  const { borrowedAmountTokens, stakedAmountXvsTokens, suppliedAmountTokens } = watch();
```
coreyar: I would put in the form santizing inputs and returning them in the correct unit to be consumed by the app. This will clean up downstream logic where these values get converted to different units

src/pages/PrimeCalculator/Form/index.tsx
```diff
@@ -112,13 +180,94 @@ export const Form: React.FC = () => {
...
+        });
+        stakedAmountXvsTokens = defaultValues.initialStakedAmountXvsTokens;
+        borrowAmountTokens = defaultValues.initialBorrowAmountTokens;
+        supplyAmountTokens = defaultValues.initialSupplyAmountTokens;
+      }
+      return {
+        initialStakedAmountXvsTokens: stakedAmountXvsTokens,
+        initialBorrowAmountTokens: borrowAmountTokens,
+        initialSupplyAmountTokens: supplyAmountTokens,
+      };
```
coreyar:
```suggestion
      let initialStakedAmountXvsTokens;
      let initialBorrowAmountTokens;
      let initialSupplyAmountTokens;

      if (selectedAsset && xvs) {
        ({initialStakedAmountXvsTokens,initialBorrowAmountTokens, initialSupplyAmountTokens}  = getInitialValues({
          assetData: selectedAsset,
          primeMinimumStakedXvsMantissa: primeMinimumStakedXvsMantissa || new BigNumber('0'),
          userXvsStakedMantissa,
          xvs,
        }));
      }
      return {
        initialStakedAmountXvsTokens,
        initialBorrowAmountTokens,
        initialSupplyAmountTokens,
      };
```

src/pages/PrimeCalculator/Form/index.tsx
```diff
@@ -14,75 +20,137 @@ import {
...
+      staleTime: Infinity,
+      queryKey: [FunctionKey.GET_LEGACY_POOL, { accountAddress, chainId }, 'PrimeCalculator'],
+    },
+  );
+  const { data: getPrimeStatusData, isLoading: isGetPrimeStatusLoading } = useGetPrimeStatus({
+    accountAddress,
+  });
+  const primeMinimumStakedXvsMantissa = getPrimeStatusData?.primeMinimumStakedXvsMantissa;
+  const primeMaximumStakedXvsMantissa = getPrimeStatusData?.primeMaximumStakedXvsMantissa;
+  const xvsVaultPoolIndex = getPrimeStatusData?.xvsVaultPoolId;
```
coreyar: This query will be easier to work with if it always returns the correct shape of an object or throws an error. If the values can be undefined they should be returned undefined. Then you can clean this up by destructuring instead of manually reassigning to variables
```suggestion
  const { data:{primeMinimumStakedXvsMantissa, primeMaximumStakedXvsMantissa, xvsVaultPoolId}, isLoading: isGetPrimeStatusLoading } = useGetPrimeStatus({
    accountAddress,
  });
  const primeMinimumStakedXvsMantissa = getPrimeStatusData?.primeMinimumStakedXvsMantissa;
  const primeMaximumStakedXvsMantissa = getPrimeStatusData?.primeMaximumStakedXvsMantissa;
  const xvsVaultPoolIndex = getPrimeStatusData?.xvsVaultPoolId;
```

src/pages/PrimeCalculator/Form/index.tsx
```diff
@@ -14,75 +20,137 @@ import {
...
+  const { data: getXvsVaultUserInfoData } = useGetXvsVaultUserInfo(
```
coreyar, line 94: Could be destructured

src/pages/PrimeCalculator/Form/index.tsx
```diff
@@ -14,75 +20,137 @@ import {
...
+import { getInitialValues } from './getInitialValues';
 
 export const QUERY_PARAM_TOKEN_ADDRESS = 'tokenAddress';
 
-const formSchema = z.object({
-  stakedAmountXvs: z.string().min(1),
-  suppliedAmountTokens: z.string().min(1),
-  borrowedAmountTokens: z.string().min(1),
-});
+const validateNumericString = (str: string, minValue = 0) => {
```
coreyar: Should this be organized under utilities?

gleiser-oliveira: Moved to its own file, just not under utilities, since it wouldn't be imported in other components

src/pages/PrimeCalculator/Form/index.tsx
gleiser-oliveira: This pattern comes from the underlying use of `useQuery`, since the query can be disabled, it means running the hook might return undefined.

src/pages/PrimeCalculator/Form/index.tsx
gleiser-oliveira: Same as above

src/pages/PrimeCalculator/Form/validateNumericString.ts
```diff
@@ -0,0 +1,12 @@
+import BigNumber from 'bignumber.js';
+
+export const validateNumericString = (str: string, minValue = new BigNumber(0)) => {
+  if (!str) {
+    return true;
```
coreyar, line 5: This doesn't look right. false, emptyStrings, undefined, and null are not numeric strings

src/clients/api/queries/getHypotheticalPrimeApys/index.ts
```diff
@@ -25,21 +38,57 @@ const getHypotheticalPrimeApys = async ({
...
-  const { borrowAPR, supplyAPR } = await primeContract.estimateAPR(
+  const data = await primeContract.estimateAPR(
     vTokenAddress,
     accountAddress,
     userBorrowBalanceMantissa.toFixed(),
     userSupplyBalanceMantissa.toFixed(),
     userXvsStakedMantissa.toFixed(),
   );
 
+  const {
```
coreyar, line 41: nit: could be done in line also

src/clients/api/queries/getHypotheticalPrimeApys/index.ts
```diff
@@ -25,21 +38,57 @@ const getHypotheticalPrimeApys = async ({
...
+    supplyCapUSD: supplyCapPriceMantissa,
+    totalScore,
+    userScore,
+  } = data;
+
   // Convert APRs to APYs
   const supplyApyPercentage = convertAprToApy({ aprBips: supplyAPR.toString() });
   const borrowApyPercentage = convertAprToApy({ aprBips: borrowAPR.toString() });
 
+  const supplyCapMantissa = new BigNumber(cappedSupply.toString());
```
coreyar, line 64: types should be set correctly in the query so you don't have to do it every time you use the value in the app

src/clients/api/queries/useGetPrimeEstimation/index.ts
```diff
@@ -0,0 +1,173 @@
...
+        value: suppliedAmountMantissa,
+        token: vToken.underlyingToken,
+      });
+
+      const borrowCapTokens = convertMantissaToTokens({
+        value: new BigNumber(borrowCapMantissa.toString()),
+        token: vToken.underlyingToken,
+      });
+
+      const supplyCapTokens = convertMantissaToTokens({
```
coreyar, line 128: Queries should return mantissas for easy calculation downstream. Otherwise it is probable they'll need to be converted back to Mantissa. Converting to tokens should be done in a view before presenting the value, ideally in a component dedicated to this with all the formatting logic

coreyar (approved):

PR #2035, opened 2024-01-10
feat: wire up XVS Bridge

gleiser-oliveira:
## Jira ticket(s)

VEN-2248

## Changes

- Added @venusprotocol/token-bridge dependency
- Added config for the `XVSProxyOFTSrc` and `XVSProxyOFTDest` contracts
- Updated `Bridge` component to use the contracts


src/clients/api/mutations/bridgeXvs/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { NULL_ADDRESS } from 'constants/address';
+import { layerZeroChainIds } from 'constants/layerZero';
+import { XVSProxyOFTDest, XVSProxyOFTSrc } from 'packages/contracts';
+import { ChainId } from 'types';
+
+export interface BridgeXvsInput {
+  tokenBridgeContract: XVSProxyOFTSrc | XVSProxyOFTDest;
+  accountAddress: string;
+  destinationChain: ChainId;
```
therealemjy:
```suggestion
  destinationChainId: ChainId;
```

src/clients/api/mutations/bridgeXvs/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  amountMantissa: BigNumber;
+  gasValueMantissa: BigNumber;
```
therealemjy: I feel this name could be confusing: I first thought this corresponded to the gas fee of the transaction made on the chain of origin (which we let users select in their wallet directly).
Perhaps something like `bridgeFeeMantissa` would be clearer. What do you think?

src/clients/api/queries/getXvsBridgeFeeEstimation/index.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const fakeSendEstimateFeeParams = [
+  layerZeroChainIds[ChainId.BSC_TESTNET],
+  '0x0000000000000000000000003d759121234cd36F8124C21aFe1c6852d2bEd848',
+  fakeAmountMantissa.toFixed(),
+  false,
+  '0x000100000000000000000000000000000000000000000000000000000000000493E0',
+];
+
+describe('api/queries/getXvsBridgeFeeEstimation', () => {
```
therealemjy: I know most tests currently include the path to the file in their name, but I'd actually advise moving away from this as it is unnecessary. It was originally added to make sure the names of the tests were unique, but that's not necessary as vitest groups them by file anyway.
```suggestion
describe('getXvsBridgeFeeEstimation', () => {
```

src/constants/layerZero.ts
```diff
@@ -0,0 +1,11 @@
+import { ChainId } from 'types';
+
+export const layerZeroChainIds: Record<ChainId, number> = {
```
therealemjy:
```suggestion
export const LAYER_ZERO_CHAIN_IDS: Record<ChainId, number> = {
```

src/constants/layerZero.ts
```diff
@@ -0,0 +1,11 @@
...
+  [ChainId.BSC_MAINNET]: 102,
+  [ChainId.BSC_TESTNET]: 10102,
+  [ChainId.ETHEREUM]: 101,
+  [ChainId.SEPOLIA]: 10161,
+};
+
+export const adapterParams =
```
therealemjy:
```suggestion
export const ADAPTER_PARAMS =
```

src/packages/tokens/infos/commonTokens/sepolia.ts
```diff
@@ -44,10 +45,24 @@ export const tokens: Token[] = [
...
   },
+  // TODO: remove duplicate
   {
-    address: '0xdb633c11d3f9e6b8d17ac2c972c9e3b05da59bf9',
+    address: '0x66ebd019E86e0af5f228a0439EBB33f045CBe63E',
     decimals: 18,
     symbol: 'XVS',
     asset: xvsLogo,
   },
+  // TODO: remove duplicate
```
therealemjy: Do we have duplicates because we're using some sort of temporary XVS contract?

src/pages/Bridge/index.tsx
```diff
@@ -132,25 +193,39 @@ const BridgePage: React.FC = () => {
...
-  const walletBalanceTokens = new BigNumber(10);
-  const limitTokens = walletBalanceTokens;
+  const onSubmit = () => {
+    bridgeXvs({
+      accountAddress: accountAddress || '',
+      amountMantissa,
+      destinationChain: toChainId,
+      gasValueMantissa: bridgeEstimatedFeeMantissa,
+    });
+  };
```
therealemjy: To prevent ever calling the bridge with an empty address: 
```suggestion
  const onSubmit = () => {
    if (accountAddress) {
      bridgeXvs({
        accountAddress,
        amountMantissa,
        destinationChain: toChainId,
        gasValueMantissa: bridgeEstimatedFeeMantissa,
      });
    }
  };
```

src/pages/Bridge/index.tsx
```diff
@@ -81,48 +120,70 @@ const BridgePage: React.FC = () => {
...
+      convertMantissaToTokens({
+        value: feeMantissa,
+        token: nativeToken,
+      }),
+    ];
+  }, [getXvsBridgeFeeEstimationData, nativeToken]);
+
   const handleChainFieldChange = useCallback(
-    ({ fromChainId, toChainId }: { fromChainId?: ChainId; toChainId?: ChainId }) => {
+    ({ newFromChainId, newToChainId }: { newFromChainId?: ChainId; newToChainId?: ChainId }) => {
```
therealemjy, line 140: I wouldn't change the names of the properties, as it can reduce reusability (when passing arguments to the function for example).
To prevent conflict with higher-level variables/props, you can instead use the object input directly:
```suggestion
    (input: { fromChainId?: ChainId; toChainId?: ChainId }) => {
```

src/pages/Bridge/__tests__/index.spec.tsx
```diff
@@ -140,6 +142,12 @@ describe('Bridge', () => {
...
   it('updates input value correctly when clicking on max button', async () => {
+    const fakeBalanceMantissa = new BigNumber('10000000000000000000');
+    (useGetBalanceOf as Vi.Mock).mockImplementation(() => ({
+      data: {
+        balanceMantissa: fakeBalanceMantissa,
+      },
+    }));
     const { getByText, getByTestId } = renderComponent(<Bridge />, {
       chainId: ChainId.BSC_TESTNET,
     });
```
therealemjy, line 179: We'll need to add more tests (for example to check that the mutation is called properly when submitting the form)

src/clients/api/mutations/bridgeXvs/useBridgeXvs.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const useBridgeXvs = (options?: Options) => {
+  const { chainId } = useChainId();
+  const xvs = useGetToken({
+    symbol: 'XVS',
+  });
+  const xvsVaultContract = useGetXvsVaultContract({ chainId, passSigner: true });
+  const tokenBridgeContractSrc = useGetXVSProxyOFTSrcContract({ chainId, passSigner: true });
+  const tokenBridgeContractDest = useGetXVSProxyOFTDestContract({ chainId, passSigner: true });
+  const tokenBridgeContract = tokenBridgeContractSrc ?? tokenBridgeContractDest;
```
therealemjy: Looking at the logic in the Bridge UI, it seems the idea is that on Ethereum and Sepolia we use the dest contract, while on other chains we use the src contract right?

To prevent confusion and also potential bugs, I think we should be more explicit about this logic:

```suggestion
  const tokenBridgeContract = chainId === ChainId.SEPOLIA || chainId === ChainId.ETHEREUM ? tokenBridgeContractDest : tokenBridgeContractSrc;
```

src/clients/api/queries/getXvsBridgeFeeEstimation/useGetXvsBridgeFeeEstimation.ts
```diff
@@ -0,0 +1,70 @@
...
+const refetchInterval = generatePseudoRandomRefetchInterval();
+
+const useGetXvsBridgeFeeEstimation = (
+  { accountAddress, amountMantissa, destinationChain }: UseGetTokenUsdPriceInput,
+  options?: Options,
+) => {
+  const { chainId } = useChainId();
+  const tokenBridgeContractSrc = useGetXVSProxyOFTSrcContract({ chainId });
+  const tokenBridgeContractDest = useGetXVSProxyOFTDestContract({ chainId });
+  const tokenBridgeContract = tokenBridgeContractSrc ?? tokenBridgeContractDest;
```
therealemjy: Same as previous comment:
```suggestion
  const tokenBridgeContract = chainId === ChainId.SEPOLIA || chainId === ChainId.ETHEREUM ? 
tokenBridgeContractDest : tokenBridgeContractSrc;
```

src/clients/api/mutations/bridgeXvs/index.ts
gleiser-oliveira: Changed it to `nativeCurrencyFeeMantissa`,  to indicate that the fee will be in BNB, ETH, and so on

src/pages/Bridge/index.tsx
gleiser-oliveira: I actually kept this, not so much because of the name conflict, but because of the callbacks:

```
callback: () => {
  setValue('fromChainId', newFromChainId); -> changing this to input.fromChainId will throw
},
```
Type narrowing does not help inside the callbacks, so you end up having to add even more code to make sure it knows the key inside `input` has a value. Destructuring `input` would work as expected, but then you have clashing names.

therealemjy: I see, that makes sense. Having to assert types would reduce code readability.

src/packages/tokens/infos/commonTokens/sepolia.ts
gleiser-oliveira: Not really, I made this as a temporary fix to Sepolia reporting two XVS addresses as reward tokens, 0x66ebd019E86e0af5f228a0439EBB33f045CBe63E and 0xDb633C11D3F9E6B8D17aC2c972C9e3B05DA59bF9. It will get cleaned up by the contract team, so I removed the obsolete addres (0xDb633C11D3F9E6B8D17aC2c972C9e3B05DA59bF9)

src/types/index.ts
```diff
@@ -14,13 +14,17 @@ export enum ChainId {
   'SEPOLIA' = 11155111,
 }
 
+export type TransactionType = 'Chain' | 'LayerZero';
```
therealemjy: To be consistent with other global types, I'd lowercase the first letter of each type:
```suggestion
export type TransactionType = 'chain' | 'layerZero';
```

src/pages/Bridge/__tests__/index.spec.tsx
```diff
@@ -152,4 +186,118 @@ describe('Bridge', () => {
       expect((getByTestId(TEST_IDS.tokenTextField) as HTMLInputElement).value).toEqual('10'),
     );
   });
+
+  it('lets the user the bridge XVS', async () => {
```
therealemjy:
```suggestion
  it('lets the user bridge XVS', async () => {
```

src/packages/translations/translations/en.json
```diff
@@ -156,11 +156,22 @@
...
+    "errors": {
+      "dailyTransactionLimitExceeded": {
+        "message": "You cannot bridge more than {{readableAmountTokens}} ({{readableAmountUsd}}) on the destination chain due to the 24-hour limit. This limit will be reset on {{ date, dd MMM yyyy h:mm a }}",
+        "submitButton": "Amount entered higher than the daily limit"
+      },
+      "singleTransactionLimitExceeded": {
+        "message": "You cannot bridge more than {{readableAmountTokens}} ({{readableAmountUsd}}) on the destination chain in a single transaction",
+        "submitButton": "Amount entered higher than the limit"
+      },
+      "doesNotHaveEnoughNativeFunds": "Not enough {{token}} for this transaction"
```
therealemjy: I know you used the copy that's in Figma, but I think we should stay consistent with another similar use case we have in the dApp: the supply/repay forms.
So I'd go with the message:
```suggestion
      "doesNotHaveEnoughNativeFunds": "Insufficient {{tokenSymbol}} balance"
```

src/pages/Bridge/index.tsx
```diff
@@ -27,46 +43,108 @@ import {
...
+const BRIDGE_DOC_URL =
+  'https://docs-v4.venus.io/technical-reference/reference-technical-articles/technical-doc-xvs-bridge';
+
 const formSchema = z.object({
-  fromChainId: z.number(),
-  toChainId: z.number(),
-  amountTokens: z.string().min(1),
+  doesNotHaveEnoughNativeFunds: z.boolean().refine((v: boolean) => !v),
+  isDailyTransactionLimitExceeded: z.boolean().refine((v: boolean) => !v),
+  isSingleTransactionLimitExceeded: z.boolean().refine((v: boolean) => !v),
```
therealemjy: The form schema should not contain fields for errors. Instead, you should handle errors and their messages through the validation rules of each field (in this case with the validation rules of `amountTokens` field): https://zod.dev/?id=strings

So for each of these errors you'd define a rule that you attach to `amountTokens` with a corresponding error message when it is not being respected.

src/pages/Bridge/index.tsx
```diff
@@ -81,83 +159,268 @@ const BridgePage: React.FC = () => {
...
+        value: convertDollarsToCents(remainingXvsDailyLimitUsd),
+      });
+      const date = fromUnixTime(dailyLimitResetTimestamp.toNumber());
+      date.setDate(date.getDate() + 1);
+      return t('bridgePage.errors.dailyTransactionLimitExceeded.message', {
+        readableAmountTokens,
+        readableAmountUsd,
+        date,
+      });
+    }
```
therealemjy: Following my previous comment about using Zod's error handling, then you'd be able to get rid of this logic (or rather rewrite it into the validation schema) and get the error to display from the form state directly.

src/utilities/generateChainExplorerUrl.ts
```diff
@@ -1,20 +1,30 @@
 import { CHAIN_METADATA } from 'constants/chainMetadata';
 import { ChainId } from 'types';
 
-export type UrlType = 'address' | 'token' | 'tx';
+export type UrlType = 'address' | 'token' | 'tx' | 'lztx';
```
therealemjy: Just to be more explicit: 
```suggestion
export type UrlType = 'address' | 'token' | 'tx' | 'layerZeroTx';
```

src/pages/Bridge/index.tsx
```diff
@@ -81,83 +159,268 @@ const BridgePage: React.FC = () => {
...
+
+    setValue('isSingleTransactionLimitExceeded', val, { shouldValidate: true });
+  }, [getXvsBridgeStatusData, setValue, xvsAmountUsd]);
+
+  // check if the global daily limit of XVS transferred was reached
+  useEffect(() => {
+    const val = xvsAmountUsd.gt(remainingXvsDailyLimitUsd);
+
+    setValue('isDailyTransactionLimitExceeded', val, { shouldValidate: true });
+  }, [remainingXvsDailyLimitUsd, setValue, xvsAmountUsd]);
```
therealemjy: Following my previous comment: this logic should be written into the validation schema of the form.

src/constants/chainMetadata.ts
```diff
@@ -9,32 +9,40 @@ export const CHAIN_METADATA: {
     name: 'BNB mainnet',
     logoSrc: bnbLogo,
     explorerUrl: 'https://bscscan.com',
+    layerZeroScanUrl: 'https://layerzeroscan.com/',
     blockTimeMs: 3000,
     blocksPerDay: 28800,
     corePoolComptrollerContractAddress: '0xfD36E2c2a6789Db23113685031d7F16329158384',
+    nativeCurrencySymbol: 'BNB',
```
therealemjy: To expend the potential use cases for this new field, I'd store the native token directly.
To do that, I'd recommend adding a new `getToken` utility function to the `tokens` package, which could look like this:

```typescript
import { ChainId } from 'types';

import { tokens } from '../../infos/commonTokens';

export interface GetTokenInput {
  chainId: ChainId;
  symbol: string;
}

export const getToken = ({ chainId, symbol }: GetTokenInput) => {
  const chainTokens = tokens[chainId];
  return chainTokens.find(token => token.symbol === symbol);
};
```

You could then use it in the chain metadata:
```suggestion
    nativeToken: getToken({ chainId: ChainId.BSC_MAINNET, symbol: 'BNB' }),
```

With this function built, you could also update the `useGetToken` hook so it uses it in order to reduce repetition.

src/constants/usdPriceToken.ts
```diff
@@ -0,0 +1,9 @@
+import { Token } from 'types';
+
+// by convention, USD prices are informed with 18 decimals
+export const USD_PRICE_TOKEN: Token = {
+  decimals: 18,
+  address: '',
+  asset: '',
+  symbol: '',
+};
```
therealemjy: This feels risky: someone in the future could use it as a reference for a "USD" token.
I understand this is to be able to use `convertPriceMantissaToDollars` in order to convert USD prices expressed with 18 decimals. I'd recommend updating `convertPriceMantissaToDollars` instead, so that instead of taking a `token` field it takes a `decimals` field. This function doesn't need to receive the full token definition since we only use the decimals for the calculation.

therealemjy (commented):
When updating the input value, the "Approve token" step appears while the bridge's quote is being fetched, making the UI jump a bit:

https://github.com/VenusProtocol/venus-protocol-interface/assets/44675210/9edfd1a3-2c49-4cfc-8443-809119522bd7

Also, on Sepolia it tells me I don't have enough ETH to send this transaction but the ETH balance in wallet should be sufficient to cover the fee:
Screen Shot 2024-01-15 at 18 33 40
Screen Shot 2024-01-15 at 18 34 57

src/pages/Bridge/index.tsx
```diff
@@ -81,48 +97,84 @@ const BridgePage: React.FC = () => {
...
+      accountAddress: accountAddress || '',
+      destinationChain: toChainId,
+      amountMantissa,
+    },
+    {
+      enabled: !!accountAddress && amountMantissa.gt(0),
+    },
+  );
+
+  const [bridgeEstimatedFeeMantissa, brigeEstimatedFeeTokens] = useMemo(() => {
```
therealemjy:
```suggestion
  const [bridgeEstimatedFeeMantissa, bridgeEstimatedFeeTokens] = useMemo(() => {
```

src/pages/Bridge/index.tsx
```diff
@@ -132,32 +184,71 @@ const BridgePage: React.FC = () => {
...
+
+  const onSubmit = async () => {
+    if (accountAddress) {
+      try {
+        await bridgeXvs({
+          accountAddress,
+          amountMantissa,
+          destinationChainId: toChainId,
+          nativeCurrencyFeeMantissa: bridgeEstimatedFeeMantissa,
+        });
```
therealemjy, line 204: I think we should reset the form when the transaction is successfully sent, to be consistent with the other forms' behavior.

therealemjy (commented):
The form is now requesting for XVS approval and disables the input before any amount is entered. To be consistent with other forms, the approve step should only be shown when an amount has been entered in the input.

Screen Shot 2024-01-17 at 09 44 19

src/pages/Bridge/useBridgeForm.ts
```diff
@@ -0,0 +1,226 @@
...
+    toChainId: z.nativeEnum(ChainId),
+    amountTokens: z.string().min(1).superRefine(validateAmountMantissa),
+    bridgeEstimatedFeeMantissa: z.custom<BigNumber | undefined>(validateBridgeFeeMantissa),
+  });
+
+  const defaultValues = useMemo(
+    () => ({
+      fromChainId: chainId,
+      toChainId: chains.find(chain => chain.id !== chainId)?.id as ChainId,
+      amountTokens: undefined,
```
therealemjy:
```suggestion
      amountTokens: '',
```

You need to set this valut to an empty string, otherwise you'll get the following error.
Screen Shot 2024-01-18 at 12 34 24

therealemjy (approved):
.

therealemjy (commented):
.

therealemjy (commented):
.

therealemjy (changes requested):
The issue with the approval being asked for at start is gone, but now the forms remains disabled after giving approval:
Screen Shot 2024-01-18 at 12 35 49

src/pages/Bridge/index.tsx
```diff
@@ -132,32 +191,97 @@ const BridgePage: React.FC = () => {
...
+    nativeToken,
+    formState.errors.amountTokens,
+    formState.errors.bridgeEstimatedFeeMantissa,
+    formState.isValid,
+    xvs,
+  ]);
+
+  // filter out which options to present when selecting chains
+  // either BSC mainnet or BSC testnet must be the origin or the destination
+  const [fromChainIdOptons, toChainIdOptions] = useMemo(() => {
```
therealemjy:
```suggestion
  const [fromChainIdOptions, toChainIdOptions] = useMemo(() => {
```

A small tip, this VSCode extension can check spelling errors in your code: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

src/pages/Bridge/index.tsx
```diff
@@ -132,32 +191,97 @@ const BridgePage: React.FC = () => {
...
+      chains.filter(c => c.id === ChainId.BSC_MAINNET || c.id === ChainId.BSC_TESTNET),
+    );
+    const otherChainsListOptions = getOptionsFromChainsList(
+      chains.filter(c => c.id !== ChainId.BSC_MAINNET && c.id !== ChainId.BSC_TESTNET),
+    );
+    if (chainId === ChainId.BSC_MAINNET || chainId === ChainId.BSC_TESTNET) {
+      return [bscChainListOptions, otherChainsListOptions];
+    }
+    return [otherChainsListOptions, bscChainListOptions];
+  }, [chainId]);
```
therealemjy, line 271: Unfortunately this logic is causing some trouble. If you connect your wallet to Sepolia, go to the Bridge page then change your wallet network to BSC testnet you end up with the following UI:
Screen Shot 2024-01-19 at 10 10 33

Moreover, if you do not list all the available chains, then users won't be able to perform all the actions they should be able to.
Here's a screenshot that illustrates what I mean:
Screen Shot 2024-01-19 at 09 59 19

In that situation, my wallet is connected to the opBNB testnet network when I open the bridge. With this, I cannot bridge XVS from opBNB testnet to Sepolia, or from BNB testnet to opBNB testnet without changing the chain from the header or my wallet directly.

I recommend keeping the original UX:
- all chains are listed in both selects (the logic inside `handleChainFieldChange` will take care of conflicts when changing chains)
- when opening the bridge, the "from" chain corresponds to the current chain of the dApp and the "to" chain corresponds to the next available chain in the list

gleiser-oliveira: Do let me know if this works for you, here's how I tackled this issue:
- I'm still filtering the options based on which chain is the selected source or destiny. In other words, it's not possible to select Sepolia as "from" and opBNB as "to"
- If the chain is changed directly in the header dropdown, the form reacts to it by making it be the "from" chain.

therealemjy (approved):

PR #2101, opened 2024-01-23
refactor: create app workspace

therealemjy:
## Jira ticket(s)

VEN-1765

## Changes

- move app to an `apps` directory and rename it to `evm`
- move config packages into a `configs` directory
- set up workspaces: apps, configs and packages
- replace semantic releases with [changeset](https://github.com/changesets/changesets). With this change, a new `changeset` script has been added so that version changes brought to packages can be recorded when opening a PR by running `yarn changeset`. I will also set up a bot that will warn about missing changesets on PRs, and another bot that will scan the `main` branch and automatically bump package versions based on new changesets

This PR effectively changes the repository into a monorepo containing multiple packages and one app. I invite anyone contributing to this project to have a quick look at [Turbo's documentation](https://turbo.build/) if you're unfamiliar with this build system. In a nutshell, Turbo uses caching to speed up processes and also simplifies managing a monorepo by using a system of pipelines, which correspond to processes that individual packages react to. This PR doesn't do anything groundbreaking here: I've set up pipelines that correspond to our existing scripts (`test`, `build`, `lint` etc.). Each pipeline can be run from the root of the project as a yarn script. e.g.: `yarn test`.

Note: the Vercel pipelines are currently failing because their config needs to be updated to support the monorepo. Once the PR merged with `main`, I'll be able to make these updates.


apps/evm/src/__mocks__/models/tokens.ts
```diff
@@ -1,7 +1,7 @@
 import bnbLogo from 'packages/tokens/img/bnb.svg';
 import busdLogo from 'packages/tokens/img/busd.svg';
 import ethLogo from 'packages/tokens/img/eth.svg';
-import lisUsdLogo from 'packages/tokens/img/lisUSD.png';
```
gleiser-oliveira: This change got reverted, probably due to a conflict

apps/evm/src/clients/api/__mocks__/index.ts
```diff
@@ -127,7 +127,7 @@ export const useGetXvsVaultPendingWithdrawalsFromBeforeUpgrade = () =>
     getXvsVaultPendingWithdrawalsFromBeforeUpgrade,
   );
 
-export const useGetIsolatedPoolsTreasuryTotals = vi.fn();
```
gleiser-oliveira: Reverted renamed hook

apps/evm/src/clients/api/index.ts
```diff
@@ -208,7 +208,7 @@ export { default as useGetPools } from './queries/useGetPools';
 
 export { default as useGetPool } from './queries/useGetPool';
 
-export { default as useGetIsolatedPoolsTreasuryTotals } from './queries/useGetIsolatedPoolsTreasuryTotals';
```
gleiser-oliveira: Same here

gleiser-oliveira (changes requested):
@therealemjy I've changed Vercel's settings and redeployed this PR to it, this way we get all green flags from CI. I think the only thing missing is fixing the changes that got reverted when rebasing.

apps/evm/src/__mocks__/models/tokens.ts
therealemjy: Very good catch, thanks!
I'll reset the `src` folder of this PR to be exactly like in `main` to make sure all latests changes are ported over.

I can see that the HAY token has been replaced by lisUSD in the token models file. That's the only change I'll make to the dApp in this PR.

gleiser-oliveira (approved):

PR #2113, opened 2024-01-25
feat: integrate Sepolia and Ethereum subgraphs

gleiser-oliveira:
## Jira ticket(s)

VEN-2342
VEN-2343

## Changes

- Activate `marketParticipantCounts` feature flag for Ethereum and Sepolia
- Add Ethereum and Sepolia subgraph URLs
- Fetch participants count based on current `chainId`

### DIFF
--- a/src/clients/subgraph/queries/getIsolatedPoolParticipantsCount/index.ts
+++ b/src/clients/subgraph/queries/getIsolatedPoolParticipantsCount/index.ts
@@ -2,6 +2,18 @@ import { request } from 'graphql-request';
 
 import gql from 'clients/subgraph/gql';
 import config from 'config';
+import { ChainId } from 'types';
 
-export const getIsolatedPoolParticipantsCount = () =>
-  request(config.subgraphUrl, gql.IsolatedPoolParticipantsCountDocument);
+export interface GetIsolatedPoolParticipantsCountInput {
+  chainId: ChainId;
+}
+
+export const getIsolatedPoolParticipantsCount = ({
+  chainId,
+}: GetIsolatedPoolParticipantsCountInput) => {
+  const subgraphUrl = config.subgraphUrls[chainId];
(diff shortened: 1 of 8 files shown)


src/clients/subgraph/queries/getIsolatedPoolParticipantsCount/index.ts
```diff
@@ -2,6 +2,15 @@ import { request } from 'graphql-request';
...
-export const getIsolatedPoolParticipantsCount = () =>
-  request(config.subgraphUrl, gql.IsolatedPoolParticipantsCountDocument);
+export interface GetIsolatedPoolParticipantsCountInput {
+  chainId: ChainId;
+}
+
+export const getIsolatedPoolParticipantsCount = ({
+  chainId,
+}: GetIsolatedPoolParticipantsCountInput) =>
+  config.subgraphUrls[chainId]
```
therealemjy: Given the type of `config.subgraphUrls`, it will always be defined for any of the supported chain IDs so you don't need the check.

gleiser-oliveira: This is related to chains that have no subgraph yet, like opBNB. Since it's empty, it will avoid trying to make a request (and logging an error doing so). Do you prefer to disable it somewhere else?

therealemjy: I'd recommend either using the feature flag you added for the participant counts, or updating the type of the config prop for the subgraph URLs so it doesn't need to list all chains.
Relying on checking for empty strings in the config is risky in my opinion

gleiser-oliveira: Changed the type and removed the opBNB key.

therealemjy (approved):

PR #2163, opened 2024-02-06
fix: filter core pool from isolated pools totals

gleiser-oliveira:
## Changes

- Fixes the supply, borrow and liquidity totals from isolated pools aggregating values from the core pool in non BNB chains


therealemjy (commented):
This looks good, however to make it clearer what this hook now does I'd rename it to something like `useGetIsolatedPoolsTreasuryTotal`

chechu (approved):

PR #2175, opened 2024-02-07
feat: validate XVS bridge mint cap

gleiser-oliveira:
## Jira ticket(s)

VEN-2368

## Changes

- Add `XvsTokenMultichain` contract to config. The ABI for the XVS token in the Ethereum and opBNB chains is slightly different than the XVS token from BNB mainnet, as it contains the functions needed 
- Validates if the amount the user wants to bridge is lower than what is still mintable plus the amount of tokens informed in the bridge input


src/pages/Bridge/useBridgeForm.ts
```diff
@@ -104,11 +115,44 @@ const useBridgeForm = ({ toChainIdRef, walletBalanceTokens, xvs }: UseBridgeForm
...
+          token: xvs,
+        });
+        ctx.addIssue({
+          code: 'custom',
+          message: t('bridgePage.errors.mintCapReached.message', {
+            readableAmountTokens,
+          }),
+          path: ['mintCapReached'],
+        });
+      }
```
therealemjy: Looks good, this just needs to be covered with tests.

Also, let's space out the code and add comments describing what you're doing here please, this function is getting quite big and compact.

src/clients/api/queries/getXvsBridgeMintStatus/useGetXvsBridgeMintStatus.ts
therealemjy, line 1: Same of the type and function names are inconsistent with the name of the file itself (e.g.: the `useGetXvsBridgeFeeEstimation` hook should be named `useGetXvsBridgeMintStatus`, the `UseGetTokenBalancesQueryKey` and `UseGetXvsBridgeFeeInput` types seem incorrectly named etc.).

src/clients/api/queries/getXvsBridgeMintStatus/useGetXvsBridgeMintStatus.ts
```diff
@@ -0,0 +1,70 @@
...
+import {
+  getXVSProxyOFTDestContractAddress,
+  getXvsTokenMultichainContract,
+} from 'packages/contracts';
+import { useProvider } from 'packages/wallet';
+import { ChainId } from 'types';
+import { generatePseudoRandomRefetchInterval } from 'utilities';
+
+interface UseGetXvsBridgeFeeInput {
+  destinationChain: ChainId;
```
therealemjy:
```suggestion
  destinationChainId: ChainId;
```

src/clients/api/queries/getXvsBridgeMintStatus/useGetXvsBridgeMintStatus.ts
```diff
@@ -0,0 +1,70 @@
...
+  });
+  const getXvsBridgeMintStatusIfAvailable = useCallback(() => {
+    if (xvsTokenMultichainContract && chainXvsProxyOftDestContractAddress) {
+      return getXvsBridgeMintStatus({
+        xvsTokenMultichainContract,
+        chainXvsProxyOftDestContractAddress,
+      });
+    }
+    return undefined;
+  }, [xvsTokenMultichainContract, chainXvsProxyOftDestContractAddress]);
```
therealemjy: I recommend staying consistent with other hooks and use the `callOrThrow` function to wrap the query used inside `useQuery`. You'd provide the parameters `xvsTokenMultichainContract` and `chainXvsProxyOftDestContractAddress` and so the hook would throw if one of these does not exist.
I'd then use the `enabled` option of the hook to disable the query when `destinationChainId` corresponds to BSC mainnet or testnet.

This could prevent confusion in the future if someone tried to use this hook and received `undefined` as data, while no error is being thrown and loading indicators (`isLoading`, `isFetching`) are `false`.

Let me know your thoughts.

therealemjy (approved):

src/clients/api/queries/getXvsBridgeMintStatus/useGetXvsBridgeMintStatus.ts
gleiser-oliveira: Late reply, but thanks for the suggestion! It fits perfectly with the pattern used all around.

PR #2184, opened 2024-02-09
feat: add opBNB mainnet

gleiser-oliveira:
## Changes

- Add opBNB mainnet
- Enable bridge and chain selecting for opBNB mainnet


src/packages/wallet/chains.ts
```diff
@@ -7,7 +7,7 @@ const getSupportedChains = (): Chain[] => {
     return [bscTestnet, opBNBTestnet, sepolia];
   }
 
-  return [bsc, mainnet];
+  return [bsc, mainnet, opBNB];
```
therealemjy: For now I suggest only enabling opBNB on the preview environment, otherwise if we merge this PR then deploy we'll enable it in production

gleiser-oliveira: The goal is to get this to production soon, but we'll keep this as a draft until then. Only the bridge has been activated so far, we'll wait for the rest of the things to get added.

therealemjy: Ok got you, thanks for the info. PR looks good to me then.

therealemjy (approved):

PR #2207, opened 2024-02-15
feat: update operation modals with wrap/unwrap native token feature

gleiser-oliveira:
## Jira ticket(s)

VEN-2388

## Changes

- Added feature flag `wrapUnwrapNativeToken`. Only active in Sepolia for now, it hides the Uniswap info notice and displays our wrap/unwrap feature in the operation modals
- Migrated the `Delimiter` component to Tailwind
- Updated the supply, borrow, repay and withdraw modals with the new UI for wrapping and unwrapping native tokens. Next step is to integrate with the contracts that will handle it


apps/evm/src/hooks/useOperationModal/Modal/RepayForm/index.tsx
```diff
@@ -361,15 +388,19 @@ const RepayForm: React.FC<RepayFormProps> = ({ asset, pool, onCloseModal }) => {
...
     },
     {
-      enabled: isIntegratedSwapEnabled,
+      enabled: !!asset.vToken.underlyingToken.wrapsNative,
     },
   );
 
+  const { data: integratedSwapTokenBalancesData } = useGetSwapTokenUserBalances({
+    accountAddress,
+  });
```
therealemjy, line 421:
```suggestion
  const { data: integratedSwapTokenBalancesData } = useGetSwapTokenUserBalances({
    accountAddress,
  }, {
      enabled: isIntegratedSwapEnabled,
    });
```

apps/evm/src/components/TokenIcon/index.tsx
```diff
@@ -1,11 +1,17 @@
...
+  useNativeTokenAsset?: boolean;
 }
 
-export const TokenIcon: React.FC<TokenIconProps> = ({ className, token }) => (
-  <img src={token.asset} alt={token.symbol} className={cn('h-6 w-6', className)} />
-);
+export const TokenIcon: React.FC<TokenIconProps> = ({ className, token, useNativeTokenAsset }) => {
+  const { nativeToken } = useGetChainMetadata();
+  const { symbol, wrapsNative } = token;
+  const asset = wrapsNative && useNativeTokenAsset ? nativeToken.asset : token.asset;
```
therealemjy: I think we should keep this component side-effect free to maximize its reusability.
I'd move this logic to the places where the token being wrapped should be displayed instead (I'll elaborate more in a following comment).

apps/evm/src/hooks/useOperationModal/Modal/index.tsx
```diff
@@ -83,7 +83,11 @@ const OperationModal: React.FC<OperationModalProps> = ({
     <Modal
       isOpen
       title={
-        <TokenIconWithSymbol token={vToken.underlyingToken} className="text-lg font-semibold" />
+        <TokenIconWithSymbol
+          token={vToken.underlyingToken}
+          useNativeTokenAsset
```
therealemjy: I think we don't need the `useNativeTokenAsset` prop at all. With the current logic, the title of the modal for WETH on Ethereum would write "ETH" which I think is a mistake, unless I'm mistaken?

My understanding is that we just want to show the same icon for WETH as for ETH, which is the change you're doing already in this PR by changing the svg file of WETH.

apps/evm/src/hooks/useOperationModal/Modal/RepayForm/index.tsx
```diff
@@ -361,15 +388,19 @@ const RepayForm: React.FC<RepayFormProps> = ({ asset, pool, onCloseModal }) => {
 
   const isSubmitting = isRepayLoading || isSwapAndRepayLoading;
 
-  const { data: tokenBalances } = useGetSwapTokenUserBalances(
+  const { data: nativeWrappedTokenBalancesData } = useGetNativeWrappedTokenUserBalances(
```
therealemjy: Because only one token can be wrapped/unwrapped on a given blockchain, you can use the existing `useGetBalanceOf` hook to fetch the balance of the user (after getting the native token from `useGetChainMetadata`).

apps/evm/src/hooks/useOperationModal/Modal/WithdrawForm/index.tsx
```diff
@@ -39,6 +41,20 @@ export const WithdrawFormUi: React.FC<WithdrawFormUiProps> = ({
 }) => {
   const { t } = useTranslation();
   const sharedStyles = useSharedStyles();
+  const { nativeToken } = useGetChainMetadata();
+  const isWrapUnwrapNativeTokenEnabled = useIsFeatureEnabled({ name: 'wrapUnwrapNativeToken' });
+
+  const canReceiveNativeToken = useMemo(
```
therealemjy: `canReceiveNativeToken` might indicate the capability for the user to receive native tokens in general on their wallet. This is a minor thing, but I'd recommend something more explicit, perhaps `canUnwrapToNativeToken`.

apps/evm/src/__mocks__/models/tokens.ts
```diff
@@ -90,6 +91,14 @@ export const eth: Token = {
   asset: ethLogo,
 };
 
+export const weth: Token = {
+  address: '0x700868CAbb60e90d77B6588ce072d9859ec8E281',
+  decimals: 18,
+  symbol: 'WETH',
+  asset: wethLogo,
+  wrapsNative: true,
```
therealemjy: Just an idea, but you could update the `wrapsNative` prop to `tokenWrapped`, so that the token being wrapped would be defined here directly. This could make this feature a bit more flexible and prevent having to look for the token being wrapped whenever we want to use it.

apps/evm/src/hooks/useOperationModal/Modal/BorrowForm/__testUtils__/fakeData.ts
```diff
@@ -16,3 +17,11 @@ export const fakeAsset = {
   userWalletBalanceTokens: new BigNumber(10000000),
   tokenPriceCents: new BigNumber(100),
 };
+
+export const fakeWEthAsset: Asset = {
```
therealemjy: Nit - since we write `weth`:

```suggestion
export const fakeWethAsset: Asset = {
```

apps/evm/src/hooks/useOperationModal/Modal/SupplyForm/index.tsx
```diff
@@ -113,18 +116,37 @@ export const SupplyFormUi: React.FC<SupplyFormUiProps> = ({
     [asset.vToken.underlyingToken, asset.vToken.address, chainId, isIntegratedSwapEnabled],
   );
 
+  const canWrapNative = useMemo(
```
therealemjy: Nit - to be in line with other similar variables:
```suggestion
  const canWrapNativeToken = useMemo(
```

therealemjy (changes requested):
There's a small gap in the token dropdown, in place where the common tokens would normally be displayed:
Screen Shot 2024-02-15 at 17 51 55

apps/evm/src/hooks/useOperationModal/Modal/index.tsx
gleiser-oliveira: To give better context about this: when we first launched the Ethereum chain, we had a specific icon for the WETH market (the one I replaced in this PR). However it was requested that for the WETH market displayed the ETH logo instead, you can see it in mainnet.

So my intention here was to keep showing the ETH logo in certain situations, like the markets table, and show the new WETH logo when selecting it from the select token list. I've talked to Jesús about it and for now we can just display the new logo everywhere, to keep keep it consistent, so I'll remove this prop. We can then showcase this to the product team and see if they approve.

apps/evm/src/hooks/useOperationModal/Modal/RepayForm/index.tsx
gleiser-oliveira: This logic is still in place, just that it was moved to be inside `useGetSwapTokenUserBalances`: https://github.com/VenusProtocol/venus-protocol-interface/pull/2207/files#diff-d424f4961a03b8882506e764cc36d28f71cc8a9d8129a6de82ec67db8701a009R41

therealemjy: I see, makes sense

apps/evm/src/hooks/useOperationModal/Modal/RepayForm/index.tsx
```diff
@@ -90,19 +91,39 @@ export const RepayFormUi: React.FC<RepayFormUiProps> = ({
 
   const sharedStyles = useSharedStyles();
   const styles = useStyles();
+  const isWrapUnwrapNativeTokenEnabled = useIsFeatureEnabled({ name: 'wrapUnwrapNativeToken' });
+  const isIntegratedSwapEnabled = useIsFeatureEnabled({ name: 'integratedSwap' });
+
+  const canWrapNativeToken = useMemo(
```
therealemjy, line 97:
```suggestion
  const canUnwrapNativeToken = useMemo(
```

apps/evm/src/hooks/useOperationModal/Modal/index.tsx
therealemjy: I see, thanks for the context.
I think displaying the same icon all the time makes more sense, otherwise we risk confusing users. Thanks for clarifying this with Jesús.

therealemjy (approved):

PR #2313, opened 2024-03-07
feat: unwrap native token in borrow/withdraw operations

gleiser-oliveira:
## Jira ticket(s)

VEN-2391

## Changes

- Wired up BorrowForm with `borrowAndUnwrap`
- Wired up WithdrawForm with `redeemUnderlyingAndUnwrap` and `redeemAndUnwrap`
- Updated addresses for the NativeTokenGateway in all testnet chains


apps/evm/src/hooks/useOperationModal/Modal/RepayForm/index.tsx
```diff
@@ -387,6 +392,27 @@ const RepayForm: React.FC<RepayFormProps> = ({ asset, pool, onCloseModal }) => {
...
+    return nativeTokenBalanceData
+      ? convertMantissaToTokens({
+          token: asset.vToken.underlyingToken.tokenWrapped,
+          value: nativeTokenBalanceData?.balanceMantissa,
+        })
+      : undefined;
+  }, [asset.vToken.underlyingToken.tokenWrapped, nativeTokenBalanceData]);
+
+  const shouldSelectNativeToken =
+    canWrapNativeToken && nativeTokenBalanceTokens?.gte(asset.userWalletBalanceTokens);
```
therealemjy: I don't think we should ever select the token being wraped by default. The UX can be confusing for users this way, as they'd be looking at a certain market (say WETH), but the form would directly propose them to repay their balance with another token (ETH). For someone who isn't familiar with the UI, they might think they can't repay using WETH.

After checking I realized we have this logic on the Supply form too. Let me know what you think, personally I'd remove the logic from both forms.

apps/evm/src/hooks/useOperationModal/Modal/SupplyForm/index.tsx
```diff
@@ -434,6 +445,26 @@ const SupplyForm: React.FC<SupplyFormProps> = ({ asset, pool, onCloseModal }) =>
...
+    return nativeTokenBalanceData
+      ? convertMantissaToTokens({
+          token: asset.vToken.underlyingToken.tokenWrapped,
+          value: nativeTokenBalanceData?.balanceMantissa,
+        })
+      : undefined;
+  }, [asset.vToken.underlyingToken.tokenWrapped, nativeTokenBalanceData]);
+
+  const shouldSelectNativeToken =
+    canWrapNativeToken && nativeTokenBalanceTokens?.gte(asset.userWalletBalanceTokens);
```
therealemjy: Same comment as on the Repay form: I wouldn't change the token selected by default.

apps/evm/src/hooks/useOperationModal/Modal/WithdrawForm/index.tsx
```diff
@@ -294,17 +315,28 @@ const WithdrawForm: React.FC<WithdrawFormProps> = ({ asset, pool, onCloseModal }
...
       });
     }
 
     // Withdraw entire supply
     if (vTokenBalanceMantissa) {
+      if (formValues.receiveNativeToken) {
+        return redeemAndUnwrap({
+          amountMantissa: vTokenBalanceMantissa,
+        });
+      }
```
therealemjy, line 335: Let's not use nested IF statements, it tends to make the logic harder to read.

A nice way to do that could be to merge all the redeem mutation functions into one `withdraw` function under the API client (similarly to how the `repay` function handles multiple use cases), which would determine what contract to call based on parameters passed to it. What do you think?
We can do this in a separate PR and I'm happy to assist. I'd follow this logic for all four actions: supply, borrow, withdraw, repay, this way we might simplify the forms a little bit.

apps/evm/src/utilities/getUniqueTokenBalances/index.ts
```diff
@@ -0,0 +1,19 @@
...
+const getUniqueTokenBalances = (...tokenBalances: TokenBalance[]) => {
+  const uniqueBalancesObj = tokenBalances.reduce<Record<string, TokenBalance>>(
+    (acc, tokenBalance) => {
+      if (tokenBalance.token.address in acc) {
+        return acc;
+      }
+      return {
+        ...acc,
+        [tokenBalance.token.address]: tokenBalance,
+      };
```
therealemjy, line 12: Nit: you could use the `areAddressesEqual` utility function to prevent cases where the same token is returned twice with the same address but using a different casing.

apps/evm/src/hooks/useOperationModal/Modal/RepayForm/index.tsx
```diff
@@ -387,6 +392,27 @@ const RepayForm: React.FC<RepayFormProps> = ({ asset, pool, onCloseModal }) => {
     [isWrapUnwrapNativeTokenEnabled, asset.vToken.underlyingToken.tokenWrapped],
   );
 
+  const nativeTokenBalanceTokens = useMemo(() => {
```
therealemjy: I think we should keep prefixing variables to make it easy to understand what balance we are defining here: `userWalletNativeTokenBalanceTokens`.

I'd follow this logic on other forms as well (Supply/Borrow/Withdraw).

apps/evm/src/hooks/useOperationModal/Modal/RepayForm/index.tsx
```diff
@@ -372,11 +373,15 @@ const RepayForm: React.FC<RepayFormProps> = ({ asset, pool, onCloseModal }) => {
   const isIntegratedSwapFeatureEnabled = useIsFeatureEnabled({ name: 'integratedSwap' });
   const { accountAddress } = useAccountAddress();
 
-  const [formValues, setFormValues] = useState<FormValues>({
-    amountTokens: '',
-    fromToken: asset.vToken.underlyingToken,
-    fixedRepayPercentage: undefined,
-  });
+  const { data: nativeTokenBalanceData } = useGetBalanceOf(
```
therealemjy: nit: I think we should keep prefixing variables to make it easy to understand what balance we are defining here: `userWalletNativeTokenBalanceData`.

therealemjy (commented):
While testing the UX with the new wrap/unwrap features, I realized we currently cannot repay an entire WETH loan using ETH. The way we currently do this for other markets is by passing `MAX_UINT256` to the `repayBorrow` function of the corresponding VToken contract, but as far as I can understand the NativeTokenGateway contract does not allow to do the same.

I've asked the contract team to confirm whether my assumption is correct and if the contract can be updated to permit this behavior ((here)[https://github.com/VenusProtocol/isolated-pools/pull/361#discussion_r1526043355]).

apps/evm/src/hooks/useOperationModal/Modal/RepayForm/index.tsx
gleiser-oliveira: This is actually a request from product, to select either the native token or the wrapped version, based on the balance the user has

apps/evm/src/hooks/useOperationModal/Modal/WithdrawForm/index.tsx
gleiser-oliveira: Sounds good! I agree, we can refactor it in a separate PR, this way the form only has to reach out a single mutation and keep this logic out of it.

apps/evm/src/hooks/useOperationModal/Modal/RepayForm/index.tsx
therealemjy: I see. Then my only suggestion would be to change this line to use `gt` instead `gte`, so that the underlying token is still favored over over the token being wrapped. What do you think?

therealemjy (approved):

PR #2331, opened 2024-03-11
refactor: move chain logos to wallets lib

gleiser-oliveira:
## Changes

- Chain logos are now separate from token logos
- Add dedicated opBNB logo


therealemjy (commented):
Looks good to me, the checks not passing is unrelated to your changes (we need to use Node 20.x).

therealemjy (approved):

PR #2349, opened 2024-03-13
feat: UI for Governance search

therealemjy:
## Jira ticket(s)

VEN-2477

## Changes

- add UI to support search feature on Governance page
- create `getProposalStateLabel` utility
- refactor `ProposalList` and `VotingWallet` to use Tailwind and remove separation between state and UI
- add `noUnusedImports` rule to Biome
- create `governanceSearch` feature flag and make tests for `useIsFeatureFlagEnabled` hook dynamic


apps/evm/src/pages/Governance/ProposalList/index.tsx
```diff
@@ -1,87 +1,177 @@
...
+        <TextField
+          isSmall
+          value={searchValue}
+          onChange={handleSearchInputChange}
+          placeholder={t('vote.searchInput.placeholder')}
+          leftIconSrc="magnifier"
+          variant="secondary"
+          className="sm:max-w-[300px] w-full"
+        />
+      </div>
```
gleiser-oliveira: Should this be behind a a feature flag for now? I'm wondering about UI releases while the search is sorted out by the subgraph

therealemjy: You're right, good catch. Let me make a feature flag for this.

gleiser-oliveira (approved):

PR #2370, opened 2024-03-15
feat: add support for Governance subgraph

therealemjy:
# JIRA ticket(s)

VEN-2477

## Changes

- update Codegen config so that each project corresponds to a different subgraph (currently "isolatedPools" and "governance") rather than to an environment. As a result, the `start` pipeline now depends on the `generate` pipeline, so that GraphQL types and queries are generated before every start (since before we generated them for all environments, now we generate them only according to the `VITE_ENVIRONMENT` environment variable)
- create `SUBGRAPH_URLS` constant
- add support for governance subgraph


apps/evm/src/clients/subgraph/queries/governance/getProposals/index.ts
```diff
@@ -0,0 +1,14 @@
...
+import type { ChainId } from 'types';
+
+export interface GetProposalsInput {
+  chainId: ChainId;
+}
+
+export const getProposals = ({ chainId }: GetProposalsInput) =>
+  SUBGRAPH_URlS.governance[chainId]
+    ? request(SUBGRAPH_URlS.governance[chainId]!, ProposalsDocument)
+    : undefined;
```
gleiser-oliveira, line 14: I've a question: is the caller of this function going to pass either BSC_MAINNET or BSC_TESTNET? If I got it right, if we are in testnet we always reach out for the BSC_TESTNET subgraph, no matter if the app is using another chain

apps/evm/src/constants/subgraphUrls.ts
```diff
@@ -0,0 +1,22 @@
...
+    [ChainId.OPBNB_MAINNET]:
+      'https://open-platform-ap.nodereal.io/5c42a03458b64b33af7cf9ff0c70c088/opbnb-mainnet-graph-query/subgraphs/name/venusprotocol/venus-isolated-pools-opbnb',
+    [ChainId.ETHEREUM]:
+      'https://api.thegraph.com/subgraphs/name/venusprotocol/venus-isolated-pools-ethereum',
+    [ChainId.SEPOLIA]:
+      'https://api.thegraph.com/subgraphs/name/venusprotocol/venus-isolated-pools-sepolia',
+  },
+  governance: {
+    [ChainId.BSC_MAINNET]: 'https://api.thegraph.com/subgraphs/name/venusprotocol/venus-governance',
+  },
```
gleiser-oliveira: We can add BSC_TESTNET now - https://thegraph.com/hosted-service/subgraph/venusprotocol/venus-governance-chapel

apps/evm/src/clients/subgraph/queries/governance/getProposals/index.ts
therealemjy: The caller can pass any of the supported chain IDs, independently from the chain the dApp is currently on (note that we don't use the `useChainId` hook which grabs the current chain ID of the dApp, we just use a free `chainId` parameter).

apps/evm/src/constants/subgraphUrls.ts
therealemjy: Thanks, I'll add it

apps/evm/src/clients/subgraph/queries/governance/getProposals/index.ts
gleiser-oliveira: Understood, I was checking if you wanted to make this function be the one to select either the BSC_TESTNET or the BSC_MAINNET subgraph based on the environment rather than the chain passed

therealemjy: No, this function is chain-agnostic, similarly to how query and mutation functions are chain-agnostic and the callers are responsible for passing the desired chain ID.

gleiser-oliveira (approved):

PR #2381, opened 2024-03-18
refactor: supply mutation function

therealemjy:
## Jira ticket(s)

VEN-2483

## Changes

- remove `wrapTokensAndSupply ` mutation function
- update `supply` mutation function to handle all supply flows


apps/evm/src/hooks/useOperationModal/Modal/SupplyForm/index.tsx
```diff
@@ -569,9 +558,11 @@ const SupplyForm: React.FC<SupplyFormProps> = ({ asset, pool, onCloseModal }) =>
       }
 
       if (isWrappingNativeToken) {
-        return wrapTokensAndSupply({
+        return supply({
           accountAddress: accountAddress || '',
           amountMantissa,
+          wrap: true,
+          poolComptrollerContractAddress: pool.comptrollerAddress,
```
gleiser-oliveira, line 565: Shouldn't this be unified with the call on line 557? Then we'd pass `wrap: isWrappingNativeToken`

therealemjy: The input type of the `supply` function is quite restrictive (so that it can handle all flows), hence why it's not unified with line 567 as otherwise TypeScript will complain that certain properties passed don't exist on the corresponding input type.

gleiser-oliveira (approved):

PR #2407, opened 2024-03-21
refactor: allow empty strings on Create proposal form

therealemjy:
## Changes

- update `FormikTextField` to set default value on mount
- update validation schema of Create proposal form to allow empty strings in call data


apps/evm/src/pages/Governance/ProposalList/CreateProposalModal/index.spec.tsx
```diff
@@ -365,7 +359,7 @@ describe('pages/Proposal/CreateProposalModal', () => {
 
     // fake Signature should add an input with a bool placeholder and on with a string placeholder
```
gleiser-oliveira:
```suggestion
    // fake Signature should add an input with an address placeholder and on with a string placeholder
```

gleiser-oliveira (approved):

PR #2408, opened 2024-03-21
fix: NativeTokenGateway flow fixes

gleiser-oliveira:
## Changes

- Pass `userWalletNativeTokenBalanceData` as a prop for supply and repay. This allow us to first load the native token balance so we can properly preselect the token with higher balance
- Get the NativeTokenGateway contract inside useWithdraw and useBorrow


apps/evm/src/containers/AssetAccessor/index.tsx
```diff
@@ -1,19 +1,24 @@
...
 export interface AssetAccessorProps {
   vToken: VToken;
   poolComptrollerAddress: string;
   connectWalletMessage: string;
   action: TokenAction;
-  children: (props: { asset: Asset; pool: Pool }) => React.ReactNode;
+  children: (props: {
+    asset: Asset;
+    pool: Pool;
+    userWalletNativeTokenBalanceData?: GetBalanceOfOutput;
```
therealemjy: I think this extends the container too much, as it is only meant to fetch a specific asset (the native balance doesn't really fit in that context).

I'd suggest creating a separate container to fetch the native balance then wrap any component that needs this data with it.

apps/evm/src/clients/api/mutations/borrow/useBorrow.ts
```diff
@@ -4,24 +4,17 @@ import { type UseSendTransactionOptions, useSendTransaction } from 'hooks/useSen
...
 import { callOrThrow, convertMantissaToTokens } from 'utilities';
 
 type TrimmedBorrowInput = Omit<BorrowInput, 'vTokenContract' | 'nativeTokenGatewayContract'>;
 type Options = UseSendTransactionOptions<TrimmedBorrowInput>;
 
-const useBorrow = (
-  {
-    vToken,
-    poolName,
-    poolComptrollerAddress,
```
therealemjy: I'd keep these two props rather than requiring a pool to be passed. We only need two properties from the `pool` object, requiring the whole object to be passed makes the hook a bit harder to use imo.

apps/evm/src/clients/api/mutations/withdraw/useWithdraw.ts
```diff
@@ -3,19 +3,20 @@ import withdraw, { type WithdrawInput } from 'clients/api/mutations/withdraw';
...
 import { useChainId } from 'libs/wallet';
-import type { VToken } from 'types';
+import type { Pool, VToken } from 'types';
 import { callOrThrow, convertMantissaToTokens } from 'utilities';
 
 type TrimmedRedeemInput = Omit<WithdrawInput, 'tokenContract' | 'accountAddress'>;
 type Options = UseSendTransactionOptions<TrimmedRedeemInput>;
 
-const useWithdraw = (
-  { vToken, poolName }: { vToken: VToken; poolName: string },
```
therealemjy, line 15: Same as for `useWithdraw`, I'd keep the `poolName` prop instead of requiring to pass the whole `pool` object.

apps/evm/src/clients/api/mutations/borrow/useBorrow.ts
gleiser-oliveira: I went back for the individual props, but made them mandatory so we avoid silently failing like before

therealemjy (approved):

PR #2416, opened 2024-03-22
feat: make Rabby available in testnet

gleiser-oliveira:
## Changes

- I think it would be useful to have Rabby available not only in mainnet, it helps trying out features and debugging


therealemjy (commented):
Nice

therealemjy (approved):

PR #2433, opened 2024-03-26
feat: wire up search on Governance page

therealemjy:
## Jira ticket(s)

VEN-2477

## Changes

- add optional `proposalState` and `search` parameters to `getProposalPreviews` subgraph function, used to handle searching for proposals by text and proposal state
- use `PAGE_PARAM_KEY` constant from `useUrlPagination` hook in `Pagination` component
- create `getProposalMinQuorumVotes` query function
- wire up search on Governance page


apps/evm/src/clients/subgraph/queries/governance/getProposalPreviews/getProposalPreviews.graphql
```diff
@@ -1,38 +1,14 @@
...
       support
     }
   }
 
-  total: proposals(first: 1, orderBy: startBlock, orderDirection: desc) {
+  total: proposals(first: 1000, where: $where) {
     id
   }
-}
+}
```
gleiser-oliveira: nit: EOF new line missing

gleiser-oliveira (commented):
@therealemjy should we activate the corresponding mainnet feature flag in this PR? I tried it on testnet, searching VIPs is looks and feels good!

therealemjy:
True, we can enable it already. I'll make the change.

gleiser-oliveira:
@therealemjy thanks, I tried it now on mainnet, here is a couple of points you might find helpful:

- Defeated proposal shows as canceled
Screen Shot 2024-03-28 at 23 12 08

- Not sure if we should show canceled proposals when filtering by succeeded 🤔
Screen Shot 2024-03-28 at 23 13 22

therealemjy:
@gleiser-oliveira Thanks for testing this. You're right in both cases, I've updated the search queries so that only proposals of the selected state are listed.

gleiser-oliveira (approved):

PR #2435, opened 2024-03-26
fix: label on Repay form when using integrated swap

therealemjy:
## Changes

- fix label on Repay form when using integrated swap

### DIFF
--- a/.changeset/cuddly-bugs-fly.md
+++ b/.changeset/cuddly-bugs-fly.md
@@ -1,0 +1,5 @@
+---
+"@venusprotocol/evm": patch
+---
+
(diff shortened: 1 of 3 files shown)


apps/evm/src/hooks/useDelegateApproval/index.ts
```diff
@@ -44,6 +44,8 @@ const useDelegateApproval = ({
       },
     );
 
+  console.log(poolComptrollerAddress);
```
gleiser-oliveira: Log here

gleiser-oliveira (approved):

PR #2444, opened 2024-03-27
feat: filter deprecated markets in BSC testnet

gleiser-oliveira:
## Changes

- Filter deprecated markets in BSC testnet, these are currently logging errors in the console


therealemjy:
Since the plan is to delist these assets completely from the contracts, I think we should just leave the code as it is for now and wait until the change is made from the contract side. What do you think?

gleiser-oliveira:
> Since the plan is to delist these assets completely from the contracts, I think we should just leave the code as it is for now and wait until the change is made from the contract side. What do you think?

@therealemjy the main reason why I filtered them out is due to being spammed with logs when working on testnet at the moment, it was a mild annoyance while I was working on the ticket related to fetching the markets from the backend. We can either take this or I can manually filter them out while working locally, both alternatives work for me.

therealemjy:
> > Since the plan is to delist these assets completely from the contracts, I think we should just leave the code as it is for now and wait until the change is made from the contract side. What do you think?
> 
> @therealemjy the main reason why I filtered them out is due to being spammed with logs when working on testnet at the moment, it was a mild annoyance while I was working on the ticket related to fetching the markets from the backend. We can either take this or I can manually filter them out while working locally, both alternatives work for me.

To be honest I didn't realize we already had logic in place to prevent spamming Sentry on BSC mainnet, so it makes sense to add this logic for BSC testnet too!

therealemjy (approved):

PR #2465, opened 2024-04-01
refactor: log errors when calls to retrieve user balances fail

therealemjy:
## Changes

- log errors when calls to retrieve user balances fail

### DIFF
--- a/.changeset/pink-fireants-try.md
+++ b/.changeset/pink-fireants-try.md
@@ -1,0 +1,5 @@
+---
+"@venusprotocol/evm": patch
+---
+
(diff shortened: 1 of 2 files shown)


apps/evm/src/pages/Account/AccountBreakdown/index.tsx
```diff
@@ -108,6 +108,8 @@ const Account: React.FC = () => {
     accountAddress,
   });
 
+  console.log(getVaultsData);
```
gleiser-oliveira: Extra log

gleiser-oliveira (approved):

PR #2482, opened 2024-04-02
fix: bridge daily transaction limit logic

gleiser-oliveira:
## Jira ticket(s)

VEN-2517

## Changes

- If the last successful transaction happened 24 hours ago or more, the next possible transaction will belong to a new window, meaning the amount that can be bridged is the maxDailyLimitUsd
- If the last successful transaction happened in the current 24 hour window, use the current logic of calculating the remainingUsdValue when validating how much a user can bridge

### DIFF
--- a/apps/evm/src/pages/Bridge/useBridgeForm.ts
+++ b/apps/evm/src/pages/Bridge/useBridgeForm.ts
@@ -1,6 +1,6 @@
 import { zodResolver } from '@hookform/resolvers/zod';
 import BigNumber from 'bignumber.js';
-import { fromUnixTime } from 'date-fns';
+import { fromUnixTime, isBefore } from 'date-fns';
 import { type MutableRefObject, useCallback, useEffect, useMemo } from 'react';
 import { useForm } from 'react-hook-form';
 import { z } from 'zod';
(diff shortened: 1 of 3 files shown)


apps/evm/src/pages/Bridge/useBridgeForm.ts
```diff
@@ -177,7 +178,16 @@ const useBridgeForm = ({ toChainIdRef, walletBalanceTokens, xvs }: UseBridgeForm
...
       // checks if this bridge transaction is going to exceed the global daily limit in USD
-      const isDailyTransactionLimitExceeded = xvsAmountUsd.gt(remainingXvsDailyLimitUsd);
+      // and if we are inside the daily limit transaction window (24 hours from the informed dailyLimitResetTimestamp)
+      const dailyLimitResetDate = fromUnixTime(dailyLimitResetTimestamp.toNumber());
+      dailyLimitResetDate.setDate(dailyLimitResetDate.getDate() + 1);
+      const nowDate = new Date();
+      const isNowBeforeDailyLimitReset = isBefore(nowDate, dailyLimitResetDate);
+      const dailyBridgeUsdLimit = isNowBeforeDailyLimitReset
+        ? remainingXvsDailyLimitUsd
+        : maxDailyLimitUsd;
```
therealemjy, line 188: I think that someone could bypass the checks, if the current time was after the daily limit reset and they made multiple transactions that separately do not exceed the maximum daily limit, but combined go above it  (in which case we'd always consider their limit to be the maximum daily limit allowed, despite having used some of it from the next 24-hour cycle).

gleiser-oliveira: Not sure if I understood the comment completely 😂, I think your concern comes from this

`const isNowBeforeDailyLimitReset = isBefore(nowDate, dailyLimitResetDate);`

The full idea is:
1) `dailyLimitResetDate` is the timestamp (`dailyLimitResetTimestamp`) informed by the contract, plus 24 hours in advance (meaning if the timestamp is right now, `dailyLimitResetDate`should be in 24 hours from now)
2) `dailyLimitResetTimestamp` only updates in the contract side if there is activity (successful transactions) in the bridge. Meaning it can "stay" in the past until there is new activity.
3) If `dailyLimitResetTimestamp` is in the past over 24 hours from now, it means that the next bridge operation can potentially use all the `maxDailyLimitUsd`.
4) Once the user submits this transaction, the contract updates `dailyLimitResetTimestamp` and we go back to the situation where `remainingXvsDailyLimitUsd` is used, because now `nowDate` is before `dailyLimitResetDate`

therealemjy: Ahh I see, that's clearer, thanks! All good for me then :)

therealemjy (approved):

apps/evm/src/pages/Bridge/useBridgeForm.ts
therealemjy: Ahh I see, that's clearer! All good for me then. Thanks :)

PR #2540, opened 2024-04-12
feat: add network statistics to Dashboard

therealemjy:
## Jira ticket(s)

VEN-2520

## Changes

- create `PoolStats` container component
- remove "Connect Wallet" banner from Dashboard page
- add network statistics to Dashboard page
- make input of `userGetTokenBalances` hook optional


gleiser-oliveira (commented):
@therealemjy I've observed that pages like the Core pool and Lido pool, which use the `Pool`end up not having the treasury info, is it intentional? I'd think that since they contribute to the overall treasury numbers in the dashboard, it would also be nice to have their own treasury numbers

therealemjy:
It is intentional yes, I followed the designs hehe
We anyway can't display a treasury value for each pool, because they all share the same treasury so my understanding is that we can't know what portion of the treasury each pool represents.

gleiser-oliveira (approved):

PR #2564, opened 2024-04-18
refactor: rename Lido pool to Staked ETH

therealemjy:
## Changes

- rename Lido pool to Staked ETH
- add "eth" icon


apps/evm/CHANGELOG.md
gleiser-oliveira, line 1: This is minor, but I think the changelog should keep the name as it was in that moment

therealemjy: You're right. I've reverted this change

gleiser-oliveira (approved):

PR #2622, opened 2024-04-30
feat: remove Convert VRT from the menu

gleiser-oliveira:
## Jira ticket(s)

VEN-2584

## Changes

- Removed "Convert VRT" from the navigation menu. The page route is still available to be accessed through its URL, this way amabassadors are still able to share it if needed.

### DIFF
--- a/apps/evm/src/containers/Layout/useGetMenuItems.tsx
+++ b/apps/evm/src/containers/Layout/useGetMenuItems.tsx
@@ -12,7 +12,6 @@ const useGetMenuItems = () => {
   const { stakedEthPoolComptrollerContractAddress } = useGetChainMetadata();
   const swapRouteEnabled = useIsFeatureEnabled({ name: 'swapRoute' });
   const historyRouteEnabled = useIsFeatureEnabled({ name: 'historyRoute' });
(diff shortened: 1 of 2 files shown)


apps/evm/src/containers/Layout/useGetMenuItems.tsx
```diff
@@ -12,7 +12,6 @@ const useGetMenuItems = () => {
   const { stakedEthPoolComptrollerContractAddress } = useGetChainMetadata();
   const swapRouteEnabled = useIsFeatureEnabled({ name: 'swapRoute' });
   const historyRouteEnabled = useIsFeatureEnabled({ name: 'historyRoute' });
-  const convertVrtRouteEnabled = useIsFeatureEnabled({ name: 'convertVrtRoute' });
```
therealemjy, line 15: We can remove the "convertVrtRoute" feature flag too then

gleiser-oliveira: We actually can't, because it is still used for the route itself, which is only valid in BSC mainnet/testnet

therealemjy: Ahh right, makes sense

therealemjy (approved):

PR #2630, opened 2024-05-01
feat: add support for Arbitrum Sepolia network

therealemjy:
## Jira ticket(s)

VEN-2543

## Changes

- add env variables for Arbitrum Sepolia
- update Venus packages
- update config of the "contracts" lib to add Arbitrum Sepolia contract records
- update feature flags to enable relevant features on Arbitrum Sepolia
- rename `convertAprToApy` to `convertAprBipsToApy`
- remove `supplyPercentageRatePerBlock` and `borrowPercentageRatePerBlock` properties from `Asset` type. These were only used on the Market page to calculate daily interests, that are now calculated using the supply and borrow APYs. As a result, the interests shown will be slightly higher than the ones currently displayed in production, which is actually a more accurate representation for them
- update `ChainMetadata` type to make `blockTimeMs` and `blocksPerDay` optional. With this PR, chains can now either have block or time based rates. If a given chain defines `blockTimeMs` and `blocksPerDay`, then its rates will be calculated based on blocks. Otherwise, they will be calculated based on time
- rename `multiplyMantissaDaily` to `calculateDailyTokenRate` and update the function so it can now handle time based rates
- rename `calculateApy` to `calculateYearlyPercentageRate` and update the function so it now returns either an APY or an APR, depending on the value of the `compound` parameter passed
- rename `getXvsVaultRewardPerBlock` to `getXvsVaultsTotalDailyDistributedXvs` and update the function so it can handle XVS vaults with time based rates


apps/evm/src/hooks/useIsFeatureEnabled/index.tsx
```diff
@@ -48,8 +49,9 @@ export const featureFlags = {
     ChainId.ETHEREUM,
     ChainId.OPBNB_TESTNET,
     ChainId.OPBNB_MAINNET,
+    ChainId.ARBITRUM_SEPOLIA,
   ],
-  newMarketPage: [ChainId.BSC_TESTNET, ChainId.OPBNB_TESTNET, ChainId.SEPOLIA],
+  newMarketPage: [ChainId.BSC_TESTNET, ChainId.SEPOLIA, ChainId.OPBNB_TESTNET],
```
gleiser-oliveira: Could add ChainId.ARBITRUM_SEPOLIA again here

apps/evm/src/libs/contracts/config/index.ts
```diff
@@ -426,6 +443,10 @@ export const contracts: ContractConfig[] = [
         [isolatedPoolsSepoliaDeployments.addresses['Comptroller_Liquid Staked ETH'].toLowerCase()]:
           isolatedPoolsSepoliaDeployments.addresses.NativeTokenGateway_vWETH_LiquidStakedETH,
       },
+      [ChainId.ARBITRUM_SEPOLIA]: {
+        [isolatedPoolsArbitrumSepoliaDeployments.addresses.Comptroller_Core.toLowerCase()]:
+          '0xa9D6d0dc76a9f0B87E52df1326F2f0E4e422BC52', // TODO: replace with the package address once it gets updated
```
gleiser-oliveira: I think the Native Token Gateway has also been added

apps/evm/src/libs/contracts/config/index.ts
```diff
@@ -261,6 +274,7 @@ export const contracts: ContractConfig[] = [
       [ChainId.BSC_MAINNET]: venusProtocolBscMainnetDeployments.addresses.Prime,
       [ChainId.ETHEREUM]: venusProtocolEthereumDeployments.addresses.Prime,
       [ChainId.SEPOLIA]: venusProtocolSepoliaDeployments.addresses.Prime,
+      [ChainId.ARBITRUM_SEPOLIA]: '0xAdB04AC4942683bc41E27d18234C8DC884786E89', // TODO: replace with the package address once it gets updated
```
gleiser-oliveira: Prime has been added to the package

apps/evm/src/hooks/useIsFeatureEnabled/index.tsx
therealemjy: Good point

gleiser-oliveira (approved):

PR #2642, opened 2024-05-03
feat: add FRAX and sFRAX to Sepolia

gleiser-oliveira:
## Jira ticket(s)

VEN-2582

## Changes

- Added FRAX and sFRAX tokens to Sepolia


chechu (approved):
The new markets were added on Sepolia, and it seems they are working as expected
Screenshot 2024-05-06 at 17 51 10

PR #2647, opened 2024-05-04
chore: fix constant naming

ipiranhaa:
## Changes

- Rename the oderByColumn to orderByColumn


gleiser-oliveira (approved):
@ipiranhaa thank you for the fix!

PR #2737, opened 2024-05-28
fix: Prime loss warning

therealemjy:
## Changes

- remove pending withdrawals when determining amount user can withdraw from XVS vault before losing their Prime token

### DIFF
--- a/apps/evm/src/pages/Vault/modals/WithdrawFromVestingVaultModal/RequestWithdrawal/index.tsx
+++ b/apps/evm/src/pages/Vault/modals/WithdrawFromVestingVaultModal/RequestWithdrawal/index.tsx
@@ -196,11 +196,12 @@ const RequestWithdrawal: React.FC<RequestWithdrawalProps> = ({
       return undefined;
     }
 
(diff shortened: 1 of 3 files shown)


apps/evm/src/pages/Vault/modals/WithdrawFromVestingVaultModal/RequestWithdrawal/index.tsx
```diff
@@ -196,9 +196,10 @@ const RequestWithdrawal: React.FC<RequestWithdrawalProps> = ({
       return undefined;
     }
 
-    const primeLossDeltaWei = xvsVaultUserInfo.stakedAmountMantissa.minus(
-      getPrimeStatusData.primeMinimumStakedXvsMantissa,
-    );
+    const primeLossDeltaWei = xvsVaultUserInfo.stakedAmountMantissa
```
gleiser-oliveira: nit: this varable should probably be mantissa

therealemjy: Good point. It's updated

gleiser-oliveira (approved):

PR #2774, opened 2024-06-04
feat: enable the XVS bridge between all chains

gleiser-oliveira:
## Jira ticket(s)

VEN-2576

## Changes

- Removed the logic that enforced one of the bridge options to be BSC, now any chain can be selected as "from" or "to"


therealemjy (approved):
Nice one

therealemjy (approved):

PR #2846, opened 2024-06-19
refactor: small UI updates

therealemjy:
## Jira ticket(s)

VEN-2636

## Changes

- replace instances of the `Paper` component with `Card`
- update alignments and border-radius of blocks using `Card` component
- update Account page placeholder
- update `Select` component: add `size` prop and update variants


apps/evm/src/libs/translations/translations/en.json
```diff
@@ -697,8 +692,7 @@
     },
     "promotionalTitle": "Boost your earnings by up to <GreenText>{{percentage}}</GreenText>",
     "stakeButtonLabel": "Stake XVS",
-    "tokensLeft_one": "<WhiteText>{{count}}</WhiteText> prime token left",
-    "tokensLeft_other": "<WhiteText>{{count}}</WhiteText> prime tokens left",
+    "tokensLeft": "TRANSLATION NEEDED",
```
gleiser-oliveira: I think this is unintended

therealemjy: Good catch. It seems `i18next-parser` is ignoring keys using pluralization. I've added them as comments to make sure they don't get overriden when calling the `extract-translation-keys` script.

gleiser-oliveira (approved):

PR #2867, opened 2024-06-24
feat: filter out errors

therealemjy:
## Changes

- filter out errors from third-party frames
- stop recording errors about missing token records. These are clogging Sentry due to some users keeping the dApp open for extended periods of time and don't highlight actual JS errors.


apps/evm/src/clients/api/queries/getIsolatedPools/formatOutput/index.ts
```diff
@@ -70,7 +69,6 @@ const formatToPools = ({
       });
 
       if (!underlyingToken) {
-        logError(`Record missing for underlying token: ${vTokenMetaData.underlyingAssetAddress}`);
```
gleiser-oliveira, line 73: I see the rationale behind removing these logs due to Sentry, but wouldn't it make debugging these cases a bit harder? Right now the message makes it clear for us what we're looking for.

One suggestion would be to keep logging in a function "logInfo", that still shows the token record is missing, but does not send the report to Sentry.

therealemjy: My reasoning for this was that we've actually never encountered this issue in production. We could keep logging into the console, but I don't think we'd gain much from it (we'd only realize a token record is missing while developing).

It's also worth noting that once we switch to fetching markets from the backend, these issues won't be happening at all since the API will be returning the token records too.

gleiser-oliveira: Alright, sounds good!

gleiser-oliveira (approved):

PR #2925, opened 2024-07-08
refactor: use correct address for native token

therealemjy:
## Changes

- add `NATIVE_TOKEN_ADDRESS` constant
- add logic to detect market with no underlying token and set native token instead

### DIFF
--- a/apps/evm/src/constants/address.ts
+++ b/apps/evm/src/constants/address.ts
@@ -1,4 +1,5 @@
 export const NULL_ADDRESS = '0x0000000000000000000000000000000000000000';
+export const NATIVE_TOKEN_ADDRESS = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE';
 
 // We do not list the following tokens in the tokens package because they are not meant to be listed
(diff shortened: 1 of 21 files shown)


apps/evm/src/constants/address.ts
```diff
@@ -1,4 +1,5 @@
 export const NULL_ADDRESS = '0x0000000000000000000000000000000000000000';
+export const NATIVE_TOKEN_ADDRESS = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE';
```
gleiser-oliveira, line 2: Isn't `0x0000000000000000000000000000000000000000` a more accurate value to an address that doesn't exist? Unless we from now on want to adopt the E address as the representation

therealemjy: `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` is a representation of the address of the native currency of an EVM compatible chain. A lot of companies use it, such as swap for example, so I figured it would be more accurate for us to do the same.

gleiser-oliveira (approved):

PR #2952, opened 2024-07-15
feat: add landing app

therealemjy:
This PR adds the landing page repository (https://github.com/VenusProtocol/landing) as an app called "landing".

## Other changes
- remove the `--` characters from the root package scripts (except `test` as otherwise we'd break the CI pipeline) as these prevented us from being able to pass values, for example to filter what workspace a certain turbo action should run for
- use React Query v5 to fetch data
- fix minor design issues after migrating to Vite

Once merged, the following steps will consist in relying on Fleek to deploy the landing page (we'll need to point the https://venus.io DNS to Fleek's endpoint) then deprecating the landing page repository.


apps/landing/public/coins/vbusd.png
gleiser-oliveira, line 1: Are the VToken icons still used?

therealemjy: Not by any of our apps, so I'll just remove them. Good catch

gleiser-oliveira (approved):

PR #2965, opened 2024-07-17
feat: add median user averages for Prime estimations in Ethereum

gleiser-oliveira:
## Changes

- Added specific amounts for the median user XVS staked, supply and borrow in Prime markets for the Ethereum network
- Refactored them to be found by the market's address instead of underlying symbol


apps/evm/src/clients/api/queries/getIsolatedPools/__tests__/__snapshots__/indexPrime.spec.ts.snap
gleiser-oliveira, line 1: Since I refactored the reference values to use VToken addresses, some mock USDT markets were updated

therealemjy (approved):

PR #2999, opened 2024-07-25
ci: enable automatic sorting on VSCode

therealemjy:
## Changes

- enable automatic sorting on VSCode

I won't add a changeset for this change, as it affects the entire repository rather than specific packages.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+  "editor.codeActionsOnSave": {
+    "source.organizeImports.biome": "explicit"
+  }
+}


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "editor.codeActionsOnSave": {
+    "source.organizeImports.biome": "explicit"
```
gleiser-oliveira, line 3: Just to be sure, this is covered by CI and the format command? So I can use neovim 😆

therealemjy: It is yes don't worry :) Feel free to use whatever IDE you want, this file simply won't do anything if you're not on VSCode but it's not mandatory.

gleiser-oliveira (approved):

PR #3000, opened 2024-07-25
feat: show commands progress on governance proposal

therealemjy:
## Jira ticket(s)

VEN-2700
VEN-2698

## Changes

- remove stepper on Proposal page when `multichainGovernance` feature is enabled
- create basic `Commands` component under Proposal page
- create `Progress` component under Proposal page to display successful commands count
- add `size` and `fillColor` properties to `ProgressCircle` component


apps/evm/src/pages/Proposal/Commands/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Progress } from './Progress';
+
+export interface CommandsProps extends CardProps {}
+
+export const Commands: React.FC<CommandsProps> = ({ ...otherProps }) => {
+  const { t } = useTranslation();
+
+  // TODO: fetch (see VEN-2701)
+  const successfulPayloadsCount = 4;
+  const totalPayloadsCount = 5;
```
gleiser-oliveira, line 12: Wouldn't these be props passed to this component?

therealemjy: I'm not sure yet. The idea I had was to fetch them from the component directly, since we use React Query it wouldn't generate an additional query if it was fetched from the Proposal page too.

gleiser-oliveira (approved):

PR #3032, opened 2024-08-01
feat: update commands layout + add switch chain CTA

therealemjy:
## Jira ticket(s)

VEN-2699

## Changes

- update UI of `Commands` component following design updates
- handle scenario where a command is executable and dApp is connected to the wrong chain
- add missing top padding to `AccordionAnimatedContent` component
- rewrite `ReadableActionSignature` component with Tailwind
- add mock for `useVote` hook
- replace the term "network" with "chain" in the dApp to be consistent


Frayst2:
IMG_5350

Only 2 commands have been executed in the test (2/7). Nevertheless, pie chart seems to show 5/7 (= green color). Is it normal? @therealemjy

therealemjy:
@Frayst2 That's interesting, the progress circle seems to be reversed in your case. Would you mind sharing some information on your device and client please (windows/mac, browser and its version)?

This is what I see on a Mac, tested on Chrome and Firefox:
Screen Shot 2024-08-08 at 10 33 58

Thanks for your contribution!

Frayst2:
@therealemjy: I have the issue with my Mac on Safari Version 17.6, and also on my mobile iOS (Safari) 16.7.8.

It works with Brave Browser and Opera with my Mac though

therealemjy:
Thank you for the extra information. Let me check if I can replicate the issue and find a fix.

therealemjy:
@Frayst2 I've identified the issue, Safari does not support negative `stroke-dashoffset` values. I updated the `ProgressCircle` component so it uses a positive value (and reverses the circle so it fills up in the same direction).

Thanks again for your contribution on this!

Frayst2:
It works properly now! Thanks

gleiser-oliveira (approved):

PR #3053, opened 2024-08-06
feat: fetch specific chart periods from API

gleiser-oliveira:
## Changes

- This PR is meant to be used together with the new API, allowing us to select the desired period from the market chart endpoint


apps/evm/src/components/charts/ApyChart/formatToReadableDate.ts
```diff
@@ -1,8 +1,15 @@
+import type { MarketHistoryPeriodType } from 'clients/api';
 import { format as formatDate } from 'date-fns/format';
 
-const READABLE_DATE_FORMAT = 'MM.dd.yy';
+const DateFormatPerPeriod: Record<MarketHistoryPeriodType, string> = {
+  year: 'MM.dd.yy',
+  halfyear: 'MM.dd.yy',
+  month: 'MM.dd.yy hh aa',
```
therealemjy: Minor thing, but I'd use a format we widely use in the dApp (although we have a few):
```suggestion
  month: 'MM.dd.yy h:mm a',
```

apps/evm/src/pages/Market/Market/MarketHistory/index.tsx
```diff
@@ -13,8 +15,23 @@ export const MarketHistory: React.FC<MarketHistoryProps> = ({
   asset,
   poolComptrollerContractAddress,
 }) => {
-  const { data: chartData, isLoading: isChartDataLoading } = useGetChartData({
+  const [selectedSupplyPeriod, setSelectedSupplyPeriod] = useState<MarketHistoryPeriodType>('year');
+  const [selectedBorrowPeriod, setSelectedBorrowPeriod] = useState<MarketHistoryPeriodType>('year');
```
therealemjy: UX-wise, do you think it would make sense to share the selected period between both graphs? I Imagine a user would want to compare them side by side.

gleiser-oliveira: I'm fine with both ways, one gives you independent controls and the other saves you an extra click. Let's try sharing the periods and see if everyone likes it.

Another thing here is that I should probably restore having 30d as the default option.

therealemjy: Sounds good to me.

Yeah, 30 days might be better for most use cases.

therealemjy (approved):

PR #3054, opened 2024-08-06
refactor: market snapshot types

gleiser-oliveira:
## Changes

- Refactor market snapshot types into possible strings as we are migrating these types to strings in the API

### DIFF
--- a/apps/evm/src/types/index.ts
+++ b/apps/evm/src/types/index.ts
@@ -298,8 +298,9 @@ export interface Market {
 }
 
 export interface MarketSnapshot {
-  blockNumber: number;
-  blockTimestamp: number;
+  // we are migrating these values to strings, as they are bigints instead of numbers
+  blockNumber: string | number;
+  blockTimestamp: string | number;
   borrowApy: string;
   supplyApy: string;
   totalBorrowCents: string;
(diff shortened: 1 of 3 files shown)


apps/evm/src/types/index.ts
```diff
@@ -298,8 +298,9 @@ export interface Market {
 }
 
 export interface MarketSnapshot {
-  blockNumber: number;
-  blockTimestamp: number;
+  // we are migrating these values to strings, as they are bigints instead of numbers
```
therealemjy, line 301: Wouldn't it make more sense for the backend to return numbers directly? These values shouldn't need the range of `BigInt`.

therealemjy (approved):

apps/evm/src/types/index.ts
gleiser-oliveira: The range of number is big enough is most cases, that's true. But potentially you can have those running in the bigint range, in fact once we adopted viem on the backend they return these values as bigints, so I decided to adopt them instead of casting back to a number

therealemjy: Ok I see. Thanks for the explanations: that works for me.

PR #3137, opened 2024-08-23
feat: enable isolated pools and Lido market on Arbitrum Sepolia

therealemjy:
## Jira ticket(s)

VEN-2779

## Changes

- enable isolated pools and Lido market on Arbitrum Sepolia
- add weETH and wstETH token records on Arbitrum Sepolia

### DIFF
--- a/.changeset/strange-steaks-attack.md
+++ b/.changeset/strange-steaks-attack.md
@@ -1,0 +1,5 @@
+---
+"@venusprotocol/evm": minor
+---
+
+enable isolated pools and Lido market on Arbitrum Sepolia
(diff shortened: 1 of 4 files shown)


gleiser-oliveira (approved):

gleiser-oliveira:
@therealemjy let's hold on merging this, since we will deploy the weETHs market on Monday.

therealemjy:
We can probably merge this already as it only concerns Arbitrum Sepolia, but if you prefer me to update this PR on Monday with support for Arbitrum One that works too.

gleiser-oliveira:
> We can probably merge this already as it only concerns Arbitrum Sepolia, but if you prefer me to update this PR on Monday with support for Arbitrum One that works too.

Oh, true, I was thinking about mainnet but we didn't change it yet, all good then!

PR #3144, opened 2024-08-25
feat: fetch Proposal page data from subgraph

therealemjy:
## Jira ticket(s)

VEN-2701

## Changes

- update `useGetProposal` hook to utilize cache from `GET_PROPOSALS` query and fetch data from subgraph
- remove VSCode settings to enable automatic sorting with Biome. This is still an experimental feature and it sometimes messes up with the code upon saving changes
- fix small UI issue with vote summaries on Proposal page
- update `formatToProposal` subgraph utility function to order votes by voting weight in descending order
- create `Proposal` GraphQL fragment
- create `getProposal` subgraph query function
- move `formatToProposal` utility function into `getVoterHistory` query folder since it is specific to fetching proposals from the API


apps/evm/src/clients/subgraph/queries/governanceBsc/getBscProposals/getBscProposals.graphql
```diff
@@ -0,0 +1,9 @@
+query Proposals($limit: Int = 10, $skip: Int = 0, $where: Proposal_filter) {
+  proposals(first: $limit, skip: $skip, where: $where, orderBy: startBlock, orderDirection: desc) {
+    ...ProposalFragment
+  }
+
+  total: proposals(first: 1000, where: $where) {
+    id
+  }
```
gleiser-oliveira: This is related to what I mentioned in a DSU, wouldn't we have to revisit this number in a not so distant future? Ideally we'd receive the total count directly from the response

therealemjy: Yeah that's a good point, especially since this number cannot be set to a higher value than 1000.
Let me ask Corey if we can find an alternative solution.

therealemjy: Corey pointed out we can simply remove the `first` parameter to fetch all proposals.

gleiser-oliveira (approved):

PR #3201, opened 2024-09-05
refactor: remove unlimited borrow supply caps

gleiser-oliveira:
## Changes

- Removed concept of unlimited borrow and supply caps


therealemjy (approved):
Code looks good. You'll just need to run `yarn extract-translation-keys` so that the translations that aren't used anymore get removed from the codebase.

PR #3295, opened 2024-09-25
feat: use TVL endpoint from API

gleiser-oliveira:
## Jira ticket(s)

VEN-2782

## Changes

- Displays TVL and number of markets across all chains


apps/landing/src/api/utils.ts
```diff
@@ -92,3 +92,18 @@ export const nFormatter = (num: number, digits = 2) => {
 
   return item ? formatValue(num / item.value) + item.symbol : formatValue(num);
 };
+
+export function formatTvlData(apiData: TvlResponseData) {
+  const totalSupplyCents = Number(apiData.suppliedSumCents.split('.', 1));
```
therealemjy, line 97: Just a side note, but the sums should never have decimals since they are expressed in cents which is the smallest unit possible for dollars (as an example, we round down values in cents on the frontend when a calculation returns a value with decimals).

therealemjy (approved):

PR #3309, opened 2024-09-27
feat: Zyfi integration

gleiser-oliveira:
## Jira ticket(s)

VEN-2844

## Changes

- integrate with Zyfi to add gas free transactions to zkSync
- add `getPayMasterInfo` query function
- add top-level `GasslessChecker` and `ResendPayingGasModal` containers
- refactor mutation functions to return transaction data that can be passed to `sendContractTransaction` instead of directly sending a transaction
- refactor `useSendTransaction` to handle gasless flow, with fallback to regular flow if conditions are not met
- update font weight of "text" variant of `Button` component
- add "gas", "gasSlashed" and "gasSad" icons
- create new `gaslessTransactions` feature flag and enable it on zkSync networks
- add `ZyFiVault` contract record to contracts package
- create `useSendContractTransaction` in charge of sending a contract transaction (with support for gasless transactions) without any UI side-effect


apps/evm/src/utilities/requestGaslessTransaction/index.ts
```diff
@@ -0,0 +1,121 @@
...
+
+      return {
+        hash: txHash,
+        wait: async (confirmations?: number) =>
+          await publicClient.waitForTransactionReceipt({ hash: txHash, confirmations }),
+      };
+    } catch (error) {
+      logError(error);
+      // return a normal transaction
+      return contract.functions[methodName](...args);
```
therealemjy: We can't send a transaction the user would need to pay gas for if the gas-free transaction failed, we need to display a modal asking for their approval on that first. We don't have designs for this yet, but I think you could prepare the feature already.

You could update the `useSendTransaction` hook directly, so that instead of taking an `fn` parameter (the function in charge of sending the transaction), it would take the same parameters as your current `requestGaslessTransaction` and be in charge of sending the transaction (gas-less or not) using the logic you wrote here (in fact you could just move this function to the folder of the hook and call it from there). I would rename this function to `sendTransaction` and give it an additional boolean parameter (e..g: `gasLess`) that determines whether the transaction is sent through the gas-less route or not. If for whatever reason the gas-less option fails then the function could throw an error indicating that, in which case you would be able to display the modal with a button calling this same function again with the boolean parameter set to `false`.

The advantage of this approach is that all the mutation hooks are already using the `useSendTransaction` hook, so we'd only depend on one piece of logic to send (gas-less and paid) transactions and track them.

apps/evm/src/utilities/requestGaslessTransaction/index.ts
```diff
@@ -0,0 +1,121 @@
...
+  txData: {
+    chainId: number;
+    from: `0x${string}`;
+    to: `0x${string}`;
+    data: `0x${string}`;
+    value: number;
+    customData: {
+      paymasterParams: {
+        paymaster: `0x${string}`;
+        paymasterInput: `0x${string}`;
```
therealemjy:
```suggestion
    chainId: number;
    from: Address;
    to: Address;
    data: Address;
    value: number;
    customData: {
      paymasterParams: {
        paymaster: Address;
        paymasterInput: Address;
```

apps/evm/src/utilities/requestGaslessTransaction/index.ts
```diff
@@ -0,0 +1,121 @@
...
+      }).extend(eip712WalletActions());
+
+      const walletClient = createWalletClient({
+        chain,
+        transport: custom(window.ethereum),
+      }).extend(eip712WalletActions());
+
+      const nonce = await publicClient.getTransactionCount({
+        address: accountAddress,
+      });
```
therealemjy: You can group three calls together: `await contract.signer.getAddress()`, `await fetch(..)` and `await publicClient.getTransactionCount()` in a `Promise.all` call.

Also, I'd throw a generic `VError` instead of an `Error`, so that whatever is calling this function can display it in a toast without extra logic.

apps/evm/src/utilities/requestGaslessTransaction/index.ts
gleiser-oliveira: Do you think that the wallet confirmation is not enough then? Because right now, if for whatever reason we can't send a gas-less transaction, you'd get a usual confirmation popup from the wallet UI, so if I understood your idea correctly, we'd have to ask it before even generating the transaction

therealemjy: I don't think it is, because the UI would be misleading: we would have told users that transactions are gas-free when they send a transaction (from the "Gas-free" badge in the header), yet when their wallet UI shows up we'd propose to send a paid transaction. We need to be explicit about why that is the case and to make sure not to catch anyone off-guard.

And correct yes, we'll have to ask before generating the transaction; only if the "Gas-free" badge is still showing in the header and after sending a gas-free transaction failed.

apps/evm/src/containers/Layout/Header/TopBar/GaslessStatus/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  // @TODO: add link to docs
+  return (
+    <Tooltip
+      title={
+        <Trans
+          i18nKey="gaslessTransactions.tooltip"
+          components={{
+            Link: <Link href={''} />,
```
gleiser-oliveira: @therealemjy let's replace this with the actual link once the docs are ready

apps/evm/src/clients/api/mutations/approveToken/index.ts
```diff
@@ -1,21 +1,25 @@
...
+
+export type ApproveTokenOutput = ContractTxData<ContractsWithApprove, 'approve'>;
 
 const approveToken = async ({
   tokenContract,
   spenderAddress,
   allowance = MAX_UINT256.toFixed(),
-}: ApproveTokenInput): Promise<ApproveTokenOutput> =>
-  tokenContract.approve(spenderAddress, allowance);
+}: ApproveTokenInput): Promise<ApproveTokenOutput> => ({
```
coreyar: This is returning an object of arguments. Should it be making a call?

apps/evm/src/clients/api/mutations/borrow/index.ts
```diff
@@ -39,7 +47,11 @@ const borrow = async ({
   }
 
   // Handle borrow flow
-  return vTokenContract.borrow(amountMantissa.toFixed());
+  return {
```
coreyar, line 50: Same here, it changed from calling a function to returning arguments

apps/evm/src/libs/wallet/hooks/useSendContractTransaction/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+      };
+
+      const payload = {
+        chainId,
+        sponsorshipRatio: 100,
+        txData: txDataPayload,
+      };
+
+      const [response, nonce] = await Promise.all([
+        fetch(ZYFI_SPONSORED_PAYMASTER_ENDPOINT, {
```
coreyar, line 81: Alternatively this could use react-query

coreyar (approved):

apps/evm/src/clients/api/mutations/approveToken/index.ts
coreyar: I see that it is just reshaping arguments which get passed down to `sendTransaction` Does it need to be async?

apps/evm/src/libs/wallet/hooks/useSendContractTransaction/index.tsx
therealemjy: I see your point and I even started refactoring this function with react-query after your comment, but it actually complicates thing. If we were to fetch the nonce using react-query for example, then we'd also have to refetch it every time a transaction has been sent otherwise we'd risk sending two transactions in a row with the same nonce. 

Fetching the response from ZyFi using react-query is also challenging here, because it depends on parameters passed to the `send` function so using the `useQuery` hook gets a bit tricky since we have to play with the `enabled` option and here too make sure it gets updated on time.

However, I think the `send` function itself can be wrapped with the `useMutation` to benefit from react-query's features.

apps/evm/src/clients/api/mutations/approveToken/index.ts
therealemjy: You're right yeah, let me update the type of definition of `sendTransaction` and make these synchronous.

coreyar (approved):

PR #3399, opened 2024-10-18
feat: display warning on test environments

therealemjy:
## Jira ticket(s)

VEN-2889

## Changes

- display warning on test environments

### DIFF
--- a/apps/evm/src/constants/production.ts
+++ b/apps/evm/src/constants/production.ts
@@ -1,3 +1,3 @@
 export const MAIN_PRODUCTION_HOST = 'app.venus.io';
-
-export const productionHosts = [MAIN_PRODUCTION_HOST, 'app-alt.venus.io'];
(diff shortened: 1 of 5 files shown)


apps/evm/src/constants/production.ts
```diff
@@ -1 +1,3 @@
-export const productionHosts = ['app.venus.io', 'app-alt.venus.io'];
+export const MAIN_PRODUCTION_HOST = 'app.venus.io';
+export const ALT_PRODUCTION_HOST = 'app.venus.io';
```
gleiser-oliveira:
```suggestion
export const ALT_PRODUCTION_HOST = 'app-alt.venus.io';
```

therealemjy: Nice catch, thanks!

gleiser-oliveira (approved):

PR #3448, opened 2024-10-29
feat: fetch domain names for addresses

gleiser-oliveira:
## Jira ticket(s)

VEN-2892

## Changes

- Fetch web3 domain names from the API


apps/evm/src/components/DomainNameOrEllipseAddress/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { cn, truncateAddress } from 'utilities';
+import type { Address } from 'viem';
+
+type DomainNameAddressProps = {
+  showProvider?: boolean;
+  showTooltip?: boolean;
+  shouldEllipseAddress?: boolean;
+} & EllipseAddressProps;
+
+export const DomainNameOrEllipseAddress: React.FC<DomainNameAddressProps> = ({
```
therealemjy: nit: how about just "Username"?

apps/evm/src/containers/Layout/ConnectButton/PrimeButton/index.tsx
```diff
@@ -25,7 +27,7 @@ export const PrimeButton: React.FC<PrimeButtonProps> = ({
       <>
         <img className="mr-2 w-5" src={primeLogoSrc} alt={t('PrimeButton.primeLogoAlt')} />
 
-        {truncateAddress(accountAddress)}
+        {addressDomainName ?? truncateAddress(accountAddress)}
```
therealemjy: You can use the `DomainNameOrEllipseAddress` component, this way you don't need to fetch and pass the domain name or truncate the address (you will need to pass `chainId` to make sure the request to fetch the domain name isn't batched, which means the `DomainNameOrEllipseAddress` component will need to be updated to add an optional `chainId` prop):
```tsx
<DomainNameOrEllipseAddress
  showProvider={false}
  showTooltip={false}
  address={accountAddress}
  shouldEllipseAddress={false}
  chainId={chainId}
/>
```

apps/evm/src/pages/VoterLeaderboard/LeaderboardTable/index.tsx
```diff
@@ -43,7 +43,7 @@ export const LeaderboardTable: React.FC<LeaderboardTableProps> = ({
               to={routes.governanceVoter.path.replace(':address', voter.address)}
               css={styles.address}
             >
-              <EllipseAddress address={voter.address} ellipseBreakpoint="lg" />
+              <DomainNameOrEllipseAddress address={voter.address} />
```
therealemjy:
```suggestion
              <DomainNameOrEllipseAddress address={voter.address} ellipseBreakpoint="lg" />
```

apps/evm/src/containers/Layout/Header/TopBar/Breadcrumbs/index.tsx
```diff
@@ -100,7 +100,12 @@ export const Breadcrumbs: React.FC = () => {
...
             <div className="inline-flex items-center gap-x-2">
-              <EllipseAddress address={params.address || ''} ellipseBreakpoint="xxl" />
+              <DomainNameOrEllipseAddress
+                address={params.address || ''}
+                showProvider={false}
+                showTooltip={false}
+                ellipseBreakpoint="xxl"
+              />
 
               {!!params.address && <CopyAddressButton address={params.address} />}
```
therealemjy:
```suggestion
	      {!!params.address &&
                <>
                  <DomainNameOrEllipseAddress
                    address={params.address}
                    showProvider={false}
                    showTooltip={false}
                    ellipseBreakpoint="xxl"
                  />

                  <CopyAddressButton address={params.address} />
                </>
              }
```

apps/evm/src/components/DomainNameOrEllipseAddress/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+            >
+              <Icon className="ml-1" name="info" />
+            </Tooltip>
+          )}
+        </div>
+      ) : shouldEllipseAddress ? (
+        <EllipseAddress className={className} address={address} {...ellipseAddressProps} />
+      ) : (
+        <span className={className}>{address}</span>
+      )}
```
therealemjy: As a rule of thumb, I think we should avoid chaining ternary conditions as it makes the logic harder to understand.
You could use IF statements that return the corresponding DOM for each option instead (that's the approach taken in other components with a similar behavior).

apps/evm/src/clients/api/queries/getAddressDomainName/useGetAddressDomainName.ts
```diff
@@ -0,0 +1,47 @@
...
+      accountAddress,
+      chainId,
+    },
+  ];
+
+  return useQuery({
+    queryKey: queryKey,
+
+    queryFn: () =>
+      callOrThrow({ accountAddress }, params => getAddressDomainName({ ...params, chainId })),
```
therealemjy: To be consistent with other query hooks: `accountAddress` should be required and the components that use the hook should make sure it is defined (using the `enabled` option in case it is not).

apps/evm/src/components/DomainNameOrEllipseAddress/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  showProvider = true,
+  showTooltip = true,
+  shouldEllipseAddress = true,
+  ...ellipseAddressProps
+}) => {
+  const { t } = useTranslation();
+  const { chainId } = useChainId();
+  const { data: domainNames } = useGetAddressDomainName({
+    accountAddress: address as Address,
+  });
```
therealemjy: `address` can be `undefined` here:
```suggestion
  const { data: domainNames } = useGetAddressDomainName({
    accountAddress: address as Address,
  }, {
    enabled: !!address
  });
```

apps/evm/src/components/DomainNameOrEllipseAddress/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  const chainDomainName = domainNames?.[chainId];
```
therealemjy: It feels unnecessary to fetch all domain names of the address if we know in advance we're only going to use the one on the current chain.

apps/evm/src/containers/Layout/ConnectButton/index.tsx
```diff
@@ -52,15 +59,43 @@ export const ConnectButton: React.FC<ConnectButtonProps> = ({
...
+  if (chainDomainName && accountAddress) {
+    content = (
+      <DomainNameOrEllipseAddress
+        showProvider={false}
+        showTooltip={false}
+        address={accountAddress}
+        shouldEllipseAddress={false}
+      />
+    );
+  }
```
therealemjy: In the current implementation, a connected user with a domain name will trigger two requests to the `/web3-domain-name` endpoint: one with `chainId` defined and one without.

You can avoid this situation by updating the `Prime` button as suggested in my previous comment, then removing the call to the `useGetAddressDomainName` hook in this file and updating the `content` variable:

```suggestion
  const content = accountAddress
    ? (
      <DomainNameOrEllipseAddress
        showProvider={false}
        showTooltip={false}
        address={accountAddress}
        shouldEllipseAddress={false}
        chainId={chainId}
      />
    )
    : t('connectButton.connect');
```

The added bonus is that you centralize the logic: the `DomainNameOrEllipseAddress` component is in charge of determining how it fetches and renders the address/domain name in every case.

apps/evm/src/components/DomainNameOrEllipseAddress/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+      ? t('web3DomainNames.providers.ens')
+      : t('web3DomainNames.providers.spaceId');
+
+  return (
+    <>
+      {chainDomainName ? (
+        <div className={cn('flex flex-row items-center', className)}>
+          {showProvider && (
+            <Tooltip title={providerText}>
+              <Icon className="mr-1" name={providerIcon} />
```
therealemjy:
```suggestion
              <Icon className="mr-1 cursor-pointer" name={providerIcon} />
```

apps/evm/src/components/DomainNameOrEllipseAddress/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+                    {t('web3DomainNames.tooltip.address', { address: truncateAddress(address) })}
+                  </span>
+                  <span className="hidden md:block">
+                    {t('web3DomainNames.tooltip.address', { address })}
+                  </span>
+                </div>
+              }
+              className="inline-flex"
+            >
+              <Icon className="ml-1" name="info" />
```
therealemjy:
```suggestion
              <Icon className="ml-1 cursor-pointer" name="info" />
```

apps/evm/src/components/DomainNameOrEllipseAddress/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const providerText =
+    chainId === ChainId.ETHEREUM || chainId === ChainId.SEPOLIA
+      ? t('web3DomainNames.providers.ens')
+      : t('web3DomainNames.providers.spaceId');
+
+  return (
+    <>
+      {chainDomainName ? (
+        <div className={cn('flex flex-row items-center', className)}>
+          {showProvider && (
```
therealemjy: nit: this one is more of a trick than anything, but you can add the `space-x-1` class so that all the children are separated by 1 unit of spacing (and you can remove the `mr-` classes from the children).

therealemjy (commented):
Some small discrepancies between the designs and the UI:
Design:
Screenshot 2024-12-09 at 12 03 27

UI:
Screenshot 2024-12-09 at 11 53 25

apps/evm/src/containers/Layout/ConnectButton/PrimeButton/index.tsx
gleiser-oliveira: Made it slightly different: added a `useBatching` prop and made the `chainId` required. So now it always batches by default unless we tell it not to, and separates `chainId` from this responsability.

apps/evm/src/components/DomainNameOrEllipseAddress/index.tsx
gleiser-oliveira: Due to https://github.com/VenusProtocol/venus-protocol-interface/pull/3448#discussion_r1876704105, `chainId` is always passed to the request, thus requesting only what is informed.

However this doesn't change, the response is always going to be:

```
{[address]: { [chainId]: domainName}}
```

Even if there is a single chain being requested.

apps/evm/src/containers/Layout/ConnectButton/PrimeButton/index.tsx
therealemjy: Nice, that's a great idea. It makes the logic much easier to understand too I think.

therealemjy (approved):

therealemjy (commented):
Nice, this looks really good

PR #3487, opened 2024-11-06
feat: show borrow and supply cap thresholds on Market page

therealemjy:
## Jira ticket(s)

VEN-2894

## Changes

- fix viewbox issues on `ProgressCircle` component
- update `ProgressCircle` props to offer more modularity with sizes and colors
- show borrow and supply cap thresholds on Market page

### DIFF
--- a/apps/evm/src/pages/Market/Market/MarketHistory/Card/CapThreshold/index.tsx
+++ b/apps/evm/src/pages/Market/Market/MarketHistory/Card/CapThreshold/index.tsx
@@ -1,0 +1,150 @@
+import BigNumber from 'bignumber.js';
+import { ProgressCircle, Tooltip } from 'components';
+import { useTranslation } from 'libs/translations';
+import { useMemo } from 'react';
+import { theme } from 'theme';
+import type { Token } from 'types';
+
+import {
+  formatCentsToReadableValue,
+  formatPercentageToReadableValue,
+  formatTokensToReadableValue,
+} from 'utilities';
+
+const THRESHOLD_GRADIENT_ID = 'cap-threshold-gradient';
+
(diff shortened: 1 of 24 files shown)


apps/evm/src/pages/Market/Market/MarketHistory/Card/CapThreshold/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+            ? t('market.supplyCapThreshold.title')
+            : t('market.borrowCapThreshold.title')}
+        </p>
+
+        <p className="text-sm font-semibold sm:text-lg">
+          {readableBalanceDollars} / {readableCapDollars}
+        </p>
+
+        <p className="text-grey text-xs">
+          {readableBalanceTokens} / {readableCapTokens}
```
gleiser-oliveira, line 145: Shouldn't we add the token symbol here? I feel the numbers alone don't say what they mean, they end looking like this: 0.052792 / 10K

therealemjy: Yeah, it could be an issue. Let me change that

gleiser-oliveira: Sure thing! And one additional question here: should this string be moved to the en.json file and then accessed through the translation layer?

therealemjy: In this I don't think so, because it would always be displayed the same regardless of the language used on the UI.

gleiser-oliveira (approved):

PR #3522, opened 2024-11-14
chore: upgrade Vite, Vitest and other Vite plugins

therealemjy:
## Changes

- upgrade Vite, Vitest and other Vite plugins to latest available versions


apps/evm/package.json
```diff
@@ -123,8 +123,8 @@
     "@venusprotocol/token-bridge": "2.3.0",
     "@venusprotocol/typescript-config": "*",
     "@venusprotocol/venus-protocol": "9.3.0",
-    "@vitejs/plugin-react": "^4.0.4",
-    "@vitest/coverage-v8": "^0.34.4",
+    "@vitejs/plugin-react": "^4.3.3",
+    "@vitest/coverage-v8": "^2.1.5",
```
gleiser-oliveira, line 127: Will we settle on v8 and not istanbul?

therealemjy: For now yes, because I wasn't able to make Istanbul work. It seems our `yarn.lock` file has some issues that need to be resolved for it to work (removing it and recreating it with `yarn install` fixed the issue, but that's a not feasible solution since it upgrades all the dependencies of the repo).
I'll switch to using Istanbul through another PR.

gleiser-oliveira (approved):

PR #3562, opened 2024-11-25
feat: create reusable chains package

therealemjy:
## Changes

- create reusable chains package
- update environment variables used for subgraph URLs so that only keys for third-party services are included
- move codegen config to `src` directory and rename it `codegen.config.ts`


apps/evm/src/types/index.ts
```diff
@@ -1,8 +1,12 @@
 import type { Token as PSToken } from '@pancakeswap/sdk';
+import type { ChainId } from '@venusprotocol/chains';
 import type BigNumber from 'bignumber.js';
 import type { BaseContract, ContractReceipt } from 'ethers';
 import type { TransactionReceipt } from 'viem';
 
+// TODO: import from package in places where it's used in the codebase
+export { ChainId, type ChainMetadata } from '@venusprotocol/chains/types';
```
therealemjy, line 8: This code is here temporarily to prevent polluting this PR with hundreds of import changes. As more reusable packages will be created, I think it would be a good practice to focus on keeping PRs creating them as small as possible by re-exporting types used across the codebase. Once the most significant reusable packages created we can open new PRs to remove these re-exports.

therealemjy:
Note for later: before merging this PR, the BSC governance subgraph URL needs to be updated to the latest deployment (once published).

apps/evm/src/pages/Account/Settings/index.tsx
```diff
@@ -1,15 +1,13 @@
...
 import { useUserChainSettings } from 'hooks/useUserChainSettings';
 import { useTranslation } from 'libs/translations';
-import { useChainId } from 'libs/wallet';
 import Section from '../Section';
 
 export const Settings: React.FC = () => {
   const { t } = useTranslation();
-  const { chainId } = useChainId();
-  const { name: chainName } = CHAIN_METADATA[chainId];
+  const chainMetadata = useGetChainMetadata();
```
gleiser-oliveira: Shouldn't this be destructured like before?

therealemjy: Sure

gleiser-oliveira (approved):

PR #3643, opened 2024-12-17
fix: avoid trying unecessary domain resolutions

gleiser-oliveira:
## Changes

- No need to try resolving names in chains other than BSC_MAINNET, ETHEREUM and ARBITRUM_ONE, it will just fail to validate


therealemjy (commented):
Nice optimisation

therealemjy (approved):

PR #3669, opened 2024-12-24
feat: add copy addresses and contract links to market page

gleiser-oliveira:
## Jira ticket(s)

VEN-2938

## Changes

- Most of the changes are related to snapshots, because I moved the token assets to a new folder, they are now separated as `img/underlyingTokens` and `img/vTokens`
- Decoupled the `Dropdown` component from the `Select` component, allowing it to be used in the new copy/add to wallet and contract link options
- Added VToken assets
- Added the new dropdown components to the Market page


apps/evm/src/libs/tokens/infos/vTokens/index.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+export const getVTokenAsset = ({
+  vTokenAddress,
+  chainId,
+}: { vTokenAddress: string; chainId: ChainId }): VTokenLogo | undefined => {
+  const chainVTokenAssets = vTokenAssetsPerChainId[chainId];
+
+  return chainVTokenAssets[vTokenAddress as Address];
+};
```
therealemjy: To be consistent with how the `tokens` lib is built, this should be a separate utility function (see `libs/utilities/getTokens.ts` for example) that is exported from the lib's barrel file (`libs/tokens/index.ts`).

apps/evm/src/libs/tokens/img/underlyingTokens/aave.svg
therealemjy, line 1: How about storing icons of vTokens and underlying tokens under the current `tokens` directory? My thinking is that this way the directory can contain all relevant token icons, including those that are neither vTokens nor underlying tokens (for example tokens distributed through a campaign and that aren't listed in the protocol).

apps/evm/src/containers/AddTokenToWalletDropdown/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { cn } from 'utilities';
+import type { ButtonProps } from '../../components/Button';
+import { Icon } from '../../components/Icon';
+
+export interface AddTokenToWalletDropdownProps extends Omit<ButtonProps, 'onClick' | 'variant'> {
+  isUserConnected: boolean;
+  vToken: VToken;
+}
+
+export const AddTokenToWalletDropdown: React.FC<AddTokenToWalletDropdownProps> = ({
```
therealemjy: Since this component is only used by the `MarketInfo` component, then it should be stored in its directory.

apps/evm/src/containers/AddTokenToWalletDropdown/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+          }
+        >
+          {vToken.asset && (
+            <img className="w-5 max-w-none flex-none" src={vToken.asset} alt={vToken.symbol} />
+          )}
+          {vToken.symbol}
+        </span>
+      ),
+      value: vToken.symbol,
+    },
```
therealemjy: Since the structure is the same for both option types, you can create a component to render both and make the code a bit more DRY.

apps/evm/src/containers/GoToTokenContractDropdown/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  return (
+    <Dropdown
+      className={className}
+      options={options}
+      buttonClassName="flex justify-center items-center bg-background/40 p-1 h-8 w-8 border-none"
+      menuTitle={t('interactive.goTo.modalTitle')}
+    >
+      {({ isDropdownOpened }) => (
+        <Icon name="link" className={cn('h-6 w-6 border-none', isDropdownOpened && 'text-blue')} />
```
therealemjy: According to do the designs:
```suggestion
        <Icon name="link" className={cn('h-5 w-5 border-none', isDropdownOpened && 'text-blue')} />
```

apps/evm/src/containers/AddTokenToWalletDropdown/index.tsx
therealemjy: `GoToTokenContractDropdown` and `GoToTokenContractDropdown` share the same UI: you can create a component reused by both.

apps/evm/src/components/Dropdown/index.tsx
```diff
@@ -0,0 +1,208 @@
...
+      onChange,
+      onBlur,
+      label,
+      placeLabelToLeft = false,
+      size = 'medium',
+      variant = 'primary',
+      menuTitle,
+      menuPosition = 'left',
+      ...otherProps
+    }: DropdownProps<TValue>,
```
therealemjy: This component feels like a select rather than a dropdown. I think this component should only be in charge of rendering the children passed to it and pass them props to handle opening/showing the menu and track whether it is opened or closed.

The `onChange` and `value` props should be part of the `Select` component definition instead, and the children should be rendered without wrapping them in a button (that part would be handled by the children directly).

apps/evm/src/libs/tokens/infos/vTokens/arbitrumOne.ts
```diff
@@ -0,0 +1,27 @@
...
+import vWethLsEthLogo from 'libs/tokens/img/vTokens/vWethLsEth.svg';
+import vWstEthLsEthLogo from 'libs/tokens/img/vTokens/vWstEthLsEth.svg';
+
+import isolatedPoolsArbitrumOneDeployments from '@venusprotocol/isolated-pools/deployments/arbitrumone_addresses.json';
+
+import type { VTokenAssets } from 'libs/tokens/types';
+
+export const vTokenAssets: VTokenAssets = {
+  // Core pool
+  [isolatedPoolsArbitrumOneDeployments.addresses.VToken_vARB_Core]: vArbCoreLogo,
```
therealemjy: The addresses should be lowercased to prevent any issues with casing mismatches. The `vsfrxETH` market of the core pool on Sepolia highlights this issue ([link](https://dapp-testnet-git-feat-copy-addresses-and-8c0beb-venus-protocol.vercel.app/#/core-pool/market/0x7C4890D673985CE22A4D38761473f190e434c956?chainId=11155111)):
Screenshot 2024-12-30 at 11 45 19

apps/evm/src/libs/tokens/infos/vTokens/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  [ChainId.BASE_SEPOLIA]: baseSepoliaVTokenAssets,
+};
+
+export const getVTokenAsset = ({
+  vTokenAddress,
+  chainId,
+}: { vTokenAddress: string; chainId: ChainId }): VTokenLogo | undefined => {
+  const chainVTokenAssets = vTokenAssetsPerChainId[chainId];
+
+  return chainVTokenAssets[vTokenAddress as Address];
```
therealemjy: Following my previous comment about lowercasing vToken addresses:
```suggestion
  return chainVTokenAssets[vTokenAddress.toLowerCase()];
```

apps/evm/src/libs/tokens/types.ts
```diff
@@ -12,3 +13,7 @@ export interface DisabledTokenAction {
   address: string;
   disabledActions: TokenAction[];
 }
+
+export type VTokenLogo = string;
```
therealemjy: I'd suggest using the `string` type directly. Otherwise why would we use the `VTokenLogo` type with vTokens, but not a `TokenLogo` type with tokens?

apps/evm/src/libs/tokens/types.ts
```diff
@@ -12,3 +13,7 @@ export interface DisabledTokenAction {
...
+
+export type VTokenAssets = Record<Address, VTokenLogo>;
```
therealemjy: I'd suggest using the `string` type instead of `Address`, to be consistent with the rest of the codebase. If you feel the `Address` type is more accurate, then we need to use it all the way (for example `getVTokenAsset`, and all other similar functions, use the `string` type and not the `Address` type) as otherwise we'll have a lot of `as Address` in the codebase.

apps/evm/src/components/Dropdown/index.tsx
```diff
@@ -0,0 +1,208 @@
...
+              className={cn(
+                'relative w-full',
+                getVariantClasses({ variant, isDropdownOpened }),
+                buttonSizeClasses,
+                buttonClassName,
+              )}
+              contentClassName={cn('w-full justify-between text-sm font-semibold')}
+            >
+              {children({ isDropdownOpened })}
+            </Button>
```
therealemjy: Because the button applies horizontal padding, then it creates a gap on which we can click and close the dropdown without firing the `onClick` handler of the children:

https://github.com/user-attachments/assets/6ba921d0-b299-4dc6-abf3-4135c4103b2a

This issue would anyway be fixed by letting the children render their own button and handle closing/opening the dropdown menu.

therealemjy (commented):
There's a race condition happening with the dropdown indexes, where you can open both at the same time by opening the one on the left then the one on the right:
Screenshot 2024-12-30 at 11 10 34

apps/evm/src/libs/tokens/img/underlyingTokens/aave.svg
gleiser-oliveira: I tried it, but I'm honestly not a fan of having them all in the same folder, I feel it's easier to understand what asset I want to import by looking at the path, instead of looking at the name of the file itself. Although it's not a case right now, if we had non-VTokens starting with a lowercase `v`, it would be kind of difficult to set them apart (eg: vSymbol/vVSymbol/vvSymbol)

gleiser-oliveira: For the possible case of unlisted tokens, I'd create a `others` folder or rename the `underlyingTokens`. In any case suggest keeping the separation for clarity

therealemjy: Ok, that works for me

apps/evm/src/libs/tokens/infos/vTokens/bscTestnet.ts
```diff
@@ -0,0 +1,158 @@
...
+import venusProtocolBscTestnetDeployments from '@venusprotocol/venus-protocol/deployments/bsctestnet_addresses.json';
+
+import type { VTokenAssets } from 'libs/tokens/types';
+
+export const vTokenAssets: VTokenAssets = {
+  // Core pool
+  // [venusProtocolBscTestnetDeployments.addresses.vAAVE.toLowerCase()]: vAaveCoreLogo,
+  // [venusProtocolBscTestnetDeployments.addresses.vADA.toLowerCase()]: vAdaCoreLogo,
+  // [venusProtocolBscTestnetDeployments.addresses.vBCH.toLowerCase()]: vBchCoreLogo,
+  // [venusProtocolBscTestnetDeployments.addresses.vBETH.toLowerCase()]: vBethCoreLogo,
```
therealemjy, line 82: Do we need to keep these?

apps/evm/src/containers/Layout/Header/MarketInfo/AddTokenToWalletDropdown/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  token: VToken | Token,
+  label: string,
+  onClick: () => void,
+) => (
+  <span
+    className="flex items-center justify-start py-3 px-4 text-left text-sm font-semibold flex-row gap-2 grow min-w-[180px] cursor-pointer"
+    onClick={onClick}
+  >
+    <TokenDropdownOption token={token} label={label} />
+  </span>
```
therealemjy: The `span` tag can also be reused by the `AddTokenToWalletDropdown` and `GoToTokenContractDropdown` components, because although you currently use different classes for each they actually share the same UI so using either for both would give the same result.

This means the whole dropdown component itself can be transformed into a reusable component used by both `AddTokenToWalletDropdown` and `GoToTokenContractDropdown`, by just passing them props for the label and the options with their on-click actions.

apps/evm/src/components/Dropdown/index.tsx
therealemjy: I think you missed this comment

therealemjy (commented):
We've lost the hover effect on dropdown options on the Market page:
Screenshot 2024-12-31 at 09 55 19

Because the padding was removed from the `Dropdown` content tag, you'll need to add it to the options of existing dropdowns within the dApp:
Screenshot 2024-12-31 at 09 54 40

Screenshot 2024-12-31 at 10 03 02

On the Proposal list, the dropdown to filter by state now shows an unstyled input:
Screenshot 2024-12-31 at 10 03 02

apps/evm/src/libs/tokens/infos/vTokens/bscTestnet.ts
gleiser-oliveira: The testnet deployment has many VTokens missing, but I reported this and it's getting added here https://github.com/VenusProtocol/venus-protocol/pull/544

therealemjy: I see, thanks for getting this updated.

apps/evm/src/components/Dropdown/types.ts
```diff
@@ -0,0 +1,23 @@
+import type { CardProps } from 'components/Card';
+
+export interface DropdownProps extends Omit<CardProps, 'onChange' | 'children'> {
```
therealemjy: I think the `CardProps` type can be removed here. Otherwise the `Dropdown` component takes an `asChild` prop that has no effect.

apps/evm/src/components/DropdownToggleButton/index.tsx
therealemjy, line 1: Since this component is only shared by the `AddTokenToWalletDropdown` and `GoToTokenContractDropdown` components, it should be contained in their parent directory.

therealemjy (approved):

PR #3753, opened 2025-01-16
feat: fetch user Merkl rewards amounts

gleiser-oliveira:
## Jira ticket(s)

VEN-2993

## Changes

- Allow the user to see the available amount of rewards accrued from Merkl campaigns
- Reworked claim reward modal to display external rewards
- Reworked mobile version of the claim reward modal to be an overlay

### DIFF
--- a/apps/evm/src/__mocks__/models/asset.ts
+++ b/apps/evm/src/__mocks__/models/asset.ts
@@ -41,6 +41,7 @@ export const assetData: Asset[] = [
         token: xvs,
         apyPercentage: new BigNumber('0.11720675342484096'),
         dailyDistributedTokens: new BigNumber('9999999.5'),
+        isActive: true,
       },
     ],
     borrowTokenDistributions: [
@@ -49,6 +50,7 @@ export const assetData: Asset[] = [
         token: xvs,
         apyPercentage: new BigNumber('4.17469243006608279'),
         dailyDistributedTokens: new BigNumber('9999999.5'),
+        isActive: true,
       },
     ],
     supplyPointDistributions: [
@@ -104,11 +106,13 @@ export const assetData: Asset[] = [
         token: xvs,
         apyPercentage: new BigNumber('1.353105649796123742'),
         dailyDistributedTokens: new BigNumber('9999999.5'),
+        isActive: true,
       },
       {
         type: 'primeSimulation',
         token: usdc,
         apyPercentage: new BigNumber('0.953105649796123742'),
+        isActive: true,
         referenceValues: {
           userBorrowBalanceTokens: new BigNumber('1000'),
           userSupplyBalanceTokens: new BigNumber('1000'),
@@ -119,6 +123,7 @@ export const assetData: Asset[] = [
         type: 'prime',
         token: usdc,
         apyPercentage: new BigNumber('0.753105649796123742'),
+        isActive: true,
       },
     ],
     borrowTokenDistributions: [
@@ -127,11 +132,13 @@ export const assetData: Asset[] = [
         token: xvs,
         apyPercentage: new BigNumber('1.670327607690572731'),
         dailyDistributedTokens: new BigNumber('9999999.5'),
+        isActive: true,
       },
       {
         type: 'primeSimulation',
         token: usdc,
         apyPercentage: new BigNumber('1.233105649796123742'),
+        isActive: true,
         referenceValues: {
           userBorrowBalanceTokens: new BigNumber('1000'),
           userSupplyBalanceTokens: new BigNumber('1000'),
@@ -142,6 +149,7 @@ export const assetData: Asset[] = [
         type: 'prime',
         token: usdc,
         apyPercentage: new BigNumber('0.913105649796123742'),
+        isActive: true,
       },
     ],
     supplyPointDistributions: [],
@@ -181,11 +189,13 @@ export const assetData: Asset[] = [
         token: xvs,
         apyPercentage: new BigNumber('0.421719501189155143'),
         dailyDistributedTokens: new BigNumber('9999999.5'),
+        isActive: true,
       },
       {
         type: 'primeSimulation',
         token: usdt,
         apyPercentage: new BigNumber('1.753105649796123742'),
+        isActive: true,
         referenceValues: {
           userBorrowBalanceTokens: new BigNumber('1000'),
           userSupplyBalanceTokens: new BigNumber('1000'),
@@ -199,11 +209,13 @@ export const assetData: Asset[] = [
         token: xvs,
         apyPercentage: new BigNumber('0.522209972682294832'),
         dailyDistributedTokens: new BigNumber('9999999.5'),
+        isActive: true,
       },
       {
         type: 'primeSimulation',
         token: usdt,
         apyPercentage: new BigNumber('1.015310564979612374'),
+        isActive: true,
         referenceValues: {
           userBorrowBalanceTokens: new BigNumber('1000'),
           userSupplyBalanceTokens: new BigNumber('1000'),
@@ -256,11 +268,13 @@ export const assetData: Asset[] = [
         token: xvs,
         apyPercentage: new BigNumber('0.678420831753642169'),
         dailyDistributedTokens: new BigNumber('9999999.5'),
+        isActive: true,
       },
       {
         type: 'primeSimulation',
         token: busd,
         apyPercentage: new BigNumber('1.753105649796123742'),
+        isActive: true,
         referenceValues: {
           userBorrowBalanceTokens: new BigNumber('1000'),
           userSupplyBalanceTokens: new BigNumber('1000'),
@@ -271,6 +285,7 @@ export const assetData: Asset[] = [
         type: 'prime',
         token: xvs,
         apyPercentage: new BigNumber(0),
+        isActive: true,
       },
     ],
     borrowTokenDistributions: [
@@ -279,11 +294,13 @@ export const assetData: Asset[] = [
         token: xvs,
         apyPercentage: new BigNumber('0.852697602175970714'),
         dailyDistributedTokens: new BigNumber('9999999.5'),
+        isActive: true,
       },
       {
         type: 'primeSimulation',
         token: busd,
         apyPercentage: new BigNumber('1.753105649796123742'),
+        isActive: true,
         referenceValues: {
           userBorrowBalanceTokens: new BigNumber('1000'),
           userSupplyBalanceTokens: new BigNumber('1000'),
@@ -294,6 +311,7 @@ export const assetData: Asset[] = [
         type: 'prime',
         token: busd,
         apyPercentage: new BigNumber('0.913105649796123742'),
+        isActive: true,
       },
     ],
     supplyPointDistributions: [
--- a/apps/evm/src/clients/api/queries/getPendingRewards/__tests__/index.spec.ts
+++ b/apps/evm/src/clients/api/queries/getPendingRewards/__tests__/index.spec.ts
@@ -61,6 +61,8 @@ describe('getPendingRewards', () => {
       venusLensContract: fakeVenusLensContract,
       vaiVaultContract: fakeVaiVaultContract,
       xvsVaultContract: fakeXvsVaultContract,
+      chainId: ChainId.BSC_TESTNET,
+      merklCampaigns: {},
     });
 
     expect(res).toMatchSnapshot();
@@ -79,6 +81,8 @@ describe('getPendingRewards', () => {
       vaiVaultContract: fakeVaiVaultContract,
       xvsVaultContract: fakeXvsVaultContract,
       primeContract: fakePrimeContract,
+      chainId: ChainId.BSC_TESTNET,
+      merklCampaigns: {},
     });
 
     expect(res).toMatchSnapshot();
--- a/apps/evm/src/clients/api/queries/getPendingRewards/__tests__/index.spec.ts
+++ b/apps/evm/src/clients/api/queries/getPendingRewards/__tests__/index.spec.ts
@@ -4,6 +4,8 @@ import tokens from '__mocks__/models/tokens';
 import type { PoolLens, Prime, VaiVault, VenusLens, XvsVault } from 'libs/contracts';
 
 import { ChainId } from 'types';
+import { restService } from 'utilities';
+import type { Mock } from 'vitest';
 import getPendingRewards from '..';
 import {
   fakeGetIsolatedPoolPendingRewardsOutput,
@@ -14,7 +16,12 @@ import {
   fakeGetXvsVaultPendingRewardOutput,
   fakeGetXvsVaultPendingWithdrawalsBeforeUpgradeOutput,
   fakeGetXvsVaultPoolInfosOutput,
+  fakeMerklCampaigns,
+  fakeMerklRewardsResponse,
 } from '../__testUtils__/fakeData';
+import { BASE_MERKL_API_URL } from '../getMerklRewards';
+
+vi.mock('utilities/restService');
 
 const fakeLegacyPoolComptrollerAddress = '0x94d1820b2D1c7c7452A163983Dc888CEC546b77D';
 const fakeIsolatedPoolComptrollerAddress = '0x1291820b2D1c7c7452A163983Dc888CEC546b78k';
@@ -49,9 +56,8 @@ const fakePrimeContract = {
 } as unknown as Prime;
 
 describe('getPendingRewards', () => {
-  test('returns pool rewards of the user in the correct format on success', async () => {
+  it('returns pool rewards of the user in the correct format on success', async () => {
     const res = await getPendingRewards({
-      chainId: ChainId.BSC_TESTNET,
       legacyPoolComptrollerContractAddress: fakeLegacyPoolComptrollerAddress,
       isolatedPoolComptrollerAddresses: [fakeIsolatedPoolComptrollerAddress],
       tokens,
@@ -68,9 +74,8 @@ describe('getPendingRewards', () => {
     expect(res).toMatchSnapshot();
   });
 
-  test('returns pool rewards of the user, including Prime rewards, in the correct format on success', async () => {
+  it('returns pool rewards of the user, including Prime rewards, in the correct format on success', async () => {
     const res = await getPendingRewards({
-      chainId: ChainId.BSC_TESTNET,
       legacyPoolComptrollerContractAddress: fakeLegacyPoolComptrollerAddress,
       isolatedPoolComptrollerAddresses: [fakeIsolatedPoolComptrollerAddress],
       tokens,
@@ -83,6 +88,37 @@ describe('getPendingRewards', () => {
       primeContract: fakePrimeContract,
       chainId: ChainId.BSC_TESTNET,
       merklCampaigns: {},
+    });
+
+    expect(res).toMatchSnapshot();
+  });
+
+  it('returns pool rewards of the user, including Merkl rewards, in the correct format on success', async () => {
+    (restService as Mock).mockImplementation(async () => ({
+      data: [fakeMerklRewardsResponse],
+    }));
+    const res = await getPendingRewards({
+      legacyPoolComptrollerContractAddress: fakeLegacyPoolComptrollerAddress,
+      isolatedPoolComptrollerAddresses: [fakeIsolatedPoolComptrollerAddress],
+      tokens,
+      xvsVestingVaultPoolCount: 1,
+      accountAddress: fakeAddress,
+      poolLensContract: fakePoolLensContract,
+      venusLensContract: fakeVenusLensContract,
+      vaiVaultContract: fakeVaiVaultContract,
+      xvsVaultContract: fakeXvsVaultContract,
+      chainId: ChainId.BSC_TESTNET,
+      merklCampaigns: fakeMerklCampaigns,
+    });
+
+    expect(restService).toHaveBeenCalledTimes(1);
+    expect(restService).toHaveBeenCalledWith({
+      baseUrl: BASE_MERKL_API_URL,
+      endpoint: `users/${fakeAddress}/rewards`,
+      method: 'GET',
+      params: {
+        chainId: ChainId.BSC_TESTNET,
+      },
     });
 
     expect(res).toMatchSnapshot();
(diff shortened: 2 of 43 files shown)


apps/evm/src/containers/Layout/ClaimRewardButton/InternalRewardGroup/index.tsx
therealemjy, line 1: This component's layout is very similar to the one of `ExternalRewardGroup`. I think you could create a reusable component to render the frame for each of them (similarly to how you render external and internal reward groups with `RewardGroupFrame`).

apps/evm/src/containers/Layout/ClaimRewardButton/InternalRewardGroup/index.tsx
```diff
@@ -3,16 +3,16 @@ import { useMemo } from 'react';
...
-  onCheckChange: (newIsChecked: boolean) => void;
+  group: InternalRewardsGroup;
+  onCheckChange?: (newIsChecked: boolean) => void;
 }
 
-export const RewardGroup: React.FC<RewardGroupProps> = ({ group, onCheckChange }) => {
+export const InternalRewardGroup: React.FC<RewardGroupProps> = ({ group, onCheckChange }) => {
   const handleOnCheckChange: CheckboxProps['onChange'] = event =>
-    onCheckChange(event.currentTarget.checked);
+    !!onCheckChange && onCheckChange(event.currentTarget.checked);
```
therealemjy: nit: you can use a cool little trick (this is just FYI):
```suggestion
    onCheckChange?.(event.currentTarget.checked);
```

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -251,8 +251,10 @@
         "disabledLabel": "Select groups to collect reward from",
         "enabledLabel": "Claim"
       },
-      "selectAll": "Select all",
-      "title": "Claim reward ({{chainName}})"
+      "externalRewards": "External rewards",
+      "rewards": "Rewards",
+      "title": "Claim reward ({{chainName}})",
+      "claimableAt": "Claimable at {{appName}}"
```
therealemjy: I think this sounds a bit awkward. How about "Claim on {{appName}}"?
Using a verb at the start indicates an action and so that this is a call to action, while "on" seems more fitting for an app than "at".

apps/evm/src/__mocks__/models/asset.ts
```diff
@@ -41,6 +41,7 @@ export const assetData: Asset[] = [
         token: xvs,
         apyPercentage: new BigNumber('0.11720675342484096'),
         dailyDistributedTokens: new BigNumber('9999999.5'),
+        isActive: true,
```
therealemjy, line 44: Seeing that you updated this file, does that mean the API now returns the `isActive` field for all distributions?
If that's the case, then perhaps we can remove the `lastRewardingSupplyBlockOrTimestamp` and `lastRewardingBorrowBlockOrTimestamp` properties from the reward distributors and remove the logic to check if a reward distributor returned from the API is still active.

apps/evm/src/clients/api/mutations/claimRewards/types.ts
```diff
@@ -28,12 +29,19 @@ export interface PrimeClaim {
...
+}
+
 export type Claim =
   | VaiVaultClaim
   | XvsVestingVaultClaim
   | LegacyPoolComptrollerClaim
   | RewardsDistributorClaim
-  | PrimeClaim;
+  | PrimeClaim
+  | MerklClaim;
```
therealemjy: You shouldn't need to add the `MerklClaim` type, as this type of claims will never be passed as input to the `claimRewards` function since it's not claimable from our dApp. Looking at the PR I don't think it's used anywhere.

apps/evm/src/containers/Layout/ClaimRewardButton/useGetGroups.ts
```diff
@@ -29,149 +29,172 @@ const useGetGroups = ({ uncheckedGroupIds }: { uncheckedGroupIds: string[] }) =>
...
+          appName: pendingRewardGroup.appName,
+          name: pendingRewardGroup.campaignName,
+          pendingRewards: pendingRewardGroup.pendingRewards.map(pendingReward => ({
+            rewardToken: pendingReward.rewardToken,
+            rewardAmountMantissa: pendingReward.rewardAmountMantissa,
+            rewardAmountCents: pendingReward.rewardAmountCents,
+          })),
+          claimUrl: pendingRewardGroup.claimUrl,
+          campaignName: pendingRewardGroup.campaignName,
+        };
```
therealemjy:
```suggestion
        const group: ExternalRewardsGroup = {
          id,
          name: pendingRewardGroup.campaignName,
          ...pendingRewardGroup,
        };
```

apps/evm/src/containers/Layout/ClaimRewardButton/useGetGroups.tsx
therealemjy, line 1: This file needs to be removed I believe: it's a duplicate of `useGetGroups.ts` in the same directory.

apps/evm/src/clients/api/queries/getPendingRewards/types.ts
```diff
@@ -23,20 +23,63 @@ export interface GetPendingRewardsInput {
...
+    amount: BigNumber;
+  }[];
+};
+
 export interface GetPendingRewardsOutput {
   pendingRewardGroups: PendingRewardGroup[];
 }
 
 export interface IsolatedPoolPendingReward {
+  rewardType: 'venus';
```
therealemjy: I don't think we need this property, it isn't used anywhere. Same thing for `ExternalPendingReward`.

Since these are used inside other types that have a reward `type` property themselves, then we can assume that the nested pending rewards are of the same type.

apps/evm/src/clients/api/queries/getPendingRewards/index.ts
```diff
@@ -88,6 +101,9 @@ const getPendingRewards = async ({
   const xvsVestingVaultPendingWithdrawalsBeforeUpgradeResults =
     await xvsVestingVaultPendingWithdrawalsBeforeUpgradeSettledPromises;
   const [isPrimeContractPausedResult, primePendingRewardsResults] = await primePromises;
+  const merklPendingRewards = merklClaimableRewardsPerPoolPromise
+    ? await merklClaimableRewardsPerPoolPromise
+    : [];
```
therealemjy: Because this promise is to make a call to a third-party server, it won't be batched with the other promises above (these make calls to our RPC provider, for which we have automatic batching in place).

For that reason, you don't need to initialize the promise before awaiting it: you can await it directly (and ideally batch it with other similar requests that aren't automatically batched, but in this case I don't believe we have the opportunity).

Also, with the current setup if Merkl's server throws an error then our dApp won't be able to load other pending rewards of the user (since the whole request will be dropped). So we should wrap this call in a `try ... catch` or `Promise.allSettled` block.

apps/evm/src/clients/api/queries/getPendingRewards/types.ts
```diff
@@ -23,20 +23,63 @@ export interface GetPendingRewardsInput {
...
+export type PendingInternalRewardSummary = {
+  type: 'legacyPool' | 'isolatedPool';
+  poolComptrollerAddress: string;
+  distributorAddress: string;
+  rewardTokenAddress: string;
+  totalRewards: BigNumber;
+  pendingRewards: {
+    vTokenAddress: string;
+    amount: BigNumber;
+  }[];
```
therealemjy: A reusable type can be created for this since it's also used by `PendingExternalRewardSummary` below.
Also, the `amount` property should be `amountMantissa` to follow the codebase's practices.

apps/evm/src/clients/api/queries/getMerklRewards/index.ts
therealemjy, line 1: This is only used by the `getPendingRewards` query function, so it should be located in its directory.

apps/evm/src/clients/api/queries/getMerklRewards/index.ts
```diff
@@ -0,0 +1,73 @@
...
+  accountAddress: string;
+  merklCampaignsPerPool: Record<string, Record<string, MerklDistribution[]>>;
+}
+
+export type GetMerklUserRewardsOutput = PendingExternalRewardSummary[];
+
+const getMerklUserRewards = async ({
+  chainId,
+  accountAddress,
+  merklCampaignsPerPool,
```
therealemjy: To simplify the logic and potentially the tests, I'd suggest not to pass `merklCampaignsPerPool` but instead only use this function to fetch the Merkl rewards from Merkl's API and return them in the wanted format (which is the approach followed in [`getApiPools`](apps/evm/src/clients/api/queries/useGetPools/getPools/getApiPools/index.ts)).

Mapping those to Merkl campaigns would then be done inside the `formatOutput` function used in the `getPendingRewards` query function.

apps/evm/src/clients/api/queries/getPendingRewards/formatOutput/formatToMerklPendingRewardGroup.ts
```diff
@@ -0,0 +1,51 @@
...
+      pendingRewards: [],
+    };
+    campaignPendingRewards.pendingRewards.push(pendingReward);
+    return {
+      ...acc,
+      [pendingReward.campaignName]: campaignPendingRewards,
+    };
+  }, {});
+
+  return [...Object.values(pendingRewardPerCampaign)];
```
therealemjy: You can do all that in one loop:

```suggestion
  const pendingRewardPerCampaign = merklRewardsSummaries.reduce<{
    [campaignName: string]: ExternalPendingRewardGroup;
  }>((acc, rewardSummary) => {
    const formattedReward = formatRewardSummaryData({
      rewardSummary,
      tokenPriceMapping,
      tokens,
    });

    if (!formattedReward?.rewardToken) {
      return acc;
    }

    const { appName, campaignName, claimUrl } = rewardSummary;

    let campaign = acc[campaignName] || {
      type: 'external',
      appName,
      campaignName,
      claimUrl,
      pendingRewards: [],
    };

    campaign = {
      ...campaign,
      pendingRewards: [...campaign.pendingRewards, formattedReward],
    };

    return {
      ...acc,
      [campaignName]: campaign,
    };
  }, {});

  return Object.values(pendingRewardPerCampaign);
```

apps/evm/src/clients/api/queries/getPendingRewards/types.ts
```diff
@@ -23,20 +23,63 @@ export interface GetPendingRewardsInput {
...
+  rewardType: 'external';
+  campaignName: string;
+  appName: string;
+  claimUrl: string;
+  rewardToken: Token;
+  rewardAmountMantissa: BigNumber;
+  rewardAmountCents: BigNumber | undefined;
+  rewardsDistributorAddress: string;
+  totalRewards: BigNumber;
+  vTokenAddressesWithPendingReward: string[];
```
therealemjy: Some of these fields aren't used:
```suggestion
  rewardToken: Token;
  rewardAmountMantissa: BigNumber;
  rewardAmountCents: BigNumber | undefined;
```

apps/evm/src/clients/api/queries/getPendingRewards/__tests__/index.spec.ts
therealemjy, line 1: It would be great to add a case to test Merkl rewards

apps/evm/src/clients/api/queries/getPendingRewards/formatOutput/formatRewardSummaryData.ts
```diff
@@ -1,17 +1,19 @@
...
 type FormatRewardSummaryDataOutput =
   | {
+      rewardType: 'venus' | 'external';
       rewardToken: Token;
       rewardAmountMantissa: BigNumber;
       rewardAmountCents: BigNumber | undefined;
       vTokenAddressesWithPendingReward: string[];
       rewardsDistributorAddress: string;
+      proofs?: string[];
+      totalRewards?: BigNumber;
```
therealemjy: These aren't used and can be removed

apps/evm/src/clients/api/queries/getPendingRewards/formatOutput/formatRewardSummaryData.ts
```diff
@@ -1,17 +1,19 @@
...
-
-import type { VenusLens } from 'libs/contracts';
 import type { Token } from 'types';
 import { convertDollarsToCents, convertMantissaToTokens } from 'utilities';
 import findTokenByAddress from 'utilities/findTokenByAddress';
+import type { PendingExternalRewardSummary, PendingInternalRewardSummary } from '../types';
 
 type FormatRewardSummaryDataOutput =
   | {
+      rewardType: 'venus' | 'external';
```
therealemjy: This isn't used and can be removed

apps/evm/src/clients/api/queries/getPendingRewards/formatOutput/formatToMerklPendingRewardGroup.ts
```diff
@@ -0,0 +1,51 @@
...
+import type { Token } from 'types';
+
+import type {
+  ExternalPendingReward,
+  ExternalPendingRewardGroup,
+  PendingExternalRewardSummary,
+} from '../types';
+import formatRewardSummaryData from './formatRewardSummaryData';
+
+function formatToMerklPendingRewardGroup({
```
therealemjy: nit: seeing that the returned pending rewards are called "external pending rewards", then perhaps we could follow the same logic to name this function

apps/evm/src/clients/api/queries/getMerklRewards/formatMerklRewardsResponse.ts
```diff
@@ -0,0 +1,139 @@
+import BigNumber from 'bignumber.js';
+import type { MerklDistribution, Token } from 'types';
+import { areAddressesEqual } from 'utilities';
+import type { GetMerklUserRewardsResponse, MerklRewardBreakdown } from '.';
+import type { PendingExternalRewardSummary } from '../getPendingRewards/types';
+
+type MerklCampaign = MerklDistribution & { poolComptrollerAddress: string };
+
+export const formatMerklRewardsPayload = (
+  apiPayload: GetMerklUserRewardsResponse,
```
therealemjy: The type of `apiPayload` could be stricter and not allow to be passed as `undefined`, since you check for that in the `getMerklRewards` query function already.

apps/evm/src/clients/api/queries/getMerklRewards/formatMerklRewardsResponse.ts
```diff
@@ -0,0 +1,139 @@
...
+    (acc, d) => ({
+      ...acc,
+      [d.rewardDetails.merklCampaignIdentifier.toLowerCase()]: d.token,
+    }),
+    {},
+  );
+
+  // from the API response, filter the reward breakdowns from campaigns relevant to Venus
+  // it will only contain one entry at most, since we filtered by chainId
+  const apiRewards = apiPayload?.[0].rewards || [];
```
therealemjy: This is causing the query function to throw an error if there aren't any external rewards to fetch, because if `apiPayload` is passed as an empty array then `apiPayload?.[0]` is `undefined`, and so we are trying to access the `rewards` property of an `undefined` value.

```suggestion
  const apiRewards = apiPayload?.[0]?.rewards || [];
```

apps/evm/src/clients/api/queries/getPendingRewards/useGetPendingRewards.ts
```diff
@@ -74,14 +76,62 @@ const useGetPendingRewards = (input: TrimmedGetPendingRewardsInput, options?: Pa
...
+
+        // and then classify them under the pool (if there are any)
+        const merklRewards =
+          Object.keys(merklRewardsPerAsset).length > 0
+            ? {
+                ...acc.merklCampaignsPerPool,
+                [pool.comptrollerAddress]: {
+                  ...merklRewardsPerAsset,
+                },
+              }
```
therealemjy: I think you can remove the top layer of that object, since the pool comptroller address key isn't used afterwards (other pieces of logic only use the rewards listed by vToken address).

apps/evm/src/clients/api/queries/getMerklRewards/formatMerklRewardsResponse.ts
```diff
@@ -0,0 +1,139 @@
+import BigNumber from 'bignumber.js';
+import type { MerklDistribution, Token } from 'types';
+import { areAddressesEqual } from 'utilities';
+import type { GetMerklUserRewardsResponse, MerklRewardBreakdown } from '.';
+import type { PendingExternalRewardSummary } from '../getPendingRewards/types';
+
+type MerklCampaign = MerklDistribution & { poolComptrollerAddress: string };
```
therealemjy: `poolComptrollerAddress` isn't used, so I believe it can be removed (and you can use the `MerklDistribution` type directly then).

apps/evm/src/clients/api/queries/getMerklRewards/formatMerklRewardsResponse.ts
```diff
@@ -0,0 +1,139 @@
...
+            ...d,
+            poolComptrollerAddress,
+          })),
+        ],
+        [],
+      );
+      return [...acc, ...assetDistributionArray];
+    },
+    [],
+  );
```
therealemjy: If you don't need the `poolComptroller` prop, then this logic can be simplified:
```suggestion
  const allDistributions = Object.values(merklCampaignsPerPool).flatMap(
    merklDistributionsPerAsset =>
      Object.values(merklDistributionsPerAsset)
        .flat()
```

apps/evm/src/clients/api/queries/getMerklRewards/formatMerklRewardsResponse.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  const campaignIdToBreakdownMapping = breakdownsFromVenusCampaigns.reduce<
+    Record<string, MerklRewardBreakdown | undefined>
+  >(
+    (acc, b) => ({
+      ...acc,
+      [b.campaignId.toLowerCase()]: b,
+    }),
+    {},
+  );
```
therealemjy: You can simplify this logic:
```suggestion
  const campaignIdToBreakdownMapping: Record<string, MerklRewardBreakdown> = {};

  apiRewards.forEach(reward => {
    reward.breakdowns.forEach(breakdown => {
      const campaignId = breakdown.campaignId.toLowerCase();
      if (campaignIdRewardTokenMapping[campaignId]) {
        campaignIdToBreakdownMapping[campaignId] = breakdown;
      }
    });
  });
```

therealemjy (commented):
The last external reward block has extra bottom margin:

Screenshot 2025-02-26 at 21 12 42

apps/evm/src/__mocks__/models/asset.ts
gleiser-oliveira: `isActive` is present for all of them, but it doesn't really substitute the logic involving `lastRewardingSupplyBlockOrTimestamp` and `lastRewardingBorrowBlockOrTimestamp`, because we need to determine if the given distribution, apart from being active, is also a borrow/supply reward.

There is also a backend PR open that is relevant for this.

apps/evm/src/clients/api/queries/getPendingRewards/index.ts
gleiser-oliveira: Wouldn't https://github.com/VenusProtocol/venus-protocol-interface/pull/3753/files#diff-7586f97a1955056ea4f70a5269f25c7cf6b689bdba30d4eeb38ed0698a4d37acR59 cover the error?

apps/evm/src/__mocks__/models/asset.ts
therealemjy: Got you, we can keep this for a future PR then once the backend has been updated.

apps/evm/src/clients/api/queries/getPendingRewards/index.ts
therealemjy: You're right, it would. But I'd recommend following the codebase's practice and make this function throw a `VError` when an error occur, then catch the error in this function (e.g.: [getApiPools.ts](https://github.com/VenusProtocol/venus-protocol-interface/blob/main/apps/evm/src/clients/api/queries/useGetPools/getPools/getApiPools/index.ts#L103-L111)).

apps/evm/src/containers/Layout/ClaimRewardButton/ClaimRewardsContent/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+              <SwitchChain>
+                <PrimaryButton
+                  className="w-full"
+                  onClick={handleClaimReward}
+                  disabled={isSubmitDisabled}
+                  data-testid={TEST_IDS.claimRewardSubmitButton}
+                  loading={isClaimingRewards}
+                >
+                  {isSubmitDisabled
+                    ? t('claimReward.modal.claimButton.disabledLabel')
```
therealemjy, line 83: We have two translations for the same component: `claimReward.modal.[...]` and `claimRewardModal.[...]`. What do you think about merging them?

apps/evm/src/clients/api/mutations/claimRewards/types.ts
```diff
@@ -28,6 +29,12 @@ export interface PrimeClaim {
   vTokenAddressesWithPendingReward: string[];
 }
 
+export interface MerklClaim {
+  contract: 'merkl';
+  totalRewards: BigNumber;
+  token: string;
+}
+
```
therealemjy: This isn't used anywhere so I believe it can be removed.

apps/evm/src/clients/api/queries/getPendingRewards/useGetPendingRewards.ts
```diff
@@ -74,14 +76,60 @@ const useGetPendingRewards = (input: TrimmedGetPendingRewardsInput, options?: Pa
...
+                    [a.vToken.address]: assetMerklRewards,
+                  }
+                : accMerklDistributionsForAsset;
+
+            return entries;
+          },
+          {},
+        );
+
+        // and then classify them under the pool (if there are any)
```
therealemjy: Since we no longer map them by pool:
```suggestion
        // and then map them by market
```

therealemjy (approved):

PR #3785, opened 2025-01-23
refactor: query function to fetch pools

therealemjy:
## Jira ticket(s)

VEN-3025

## Changes

- lowercase variable names of ABIs exported from the contracts lib
- create `usePublicClient` hook, exported from the contracts lib
- refactor `getApiPools` query function to use viem
- remove unused `getVaiCalculateRepayAmount` query function
- rename `getVaiRepayAmountWithInterests` query function to `getUserVaiBorrowBalance`


apps/evm/src/clients/api/queries/useGetPools/getPools/index.ts
```diff
@@ -30,87 +28,101 @@ const safeGetIsolatedPoolParticipantCount = async (
...
-        legacyPoolComptrollerContract,
+        accountAddress: accountAddress as Address,
+        legacyPoolComptrollerContractAddress,
         apiPools,
-        provider,
+        publicClient,
       }),
       getUserTokenBalances({
-        accountAddress,
+        accountAddress: accountAddress as Address,
```
gleiser-oliveira: Wouldn't it be better to cast to `Address` in the hook and have the account address be of type `Address | undefined` here? Would avoid overriding the type in multiple lines

apps/evm/src/clients/api/queries/useGetPools/getPools/getUserTokenBalances/index.ts
```diff
@@ -73,17 +75,39 @@ export const getUserTokenBalances = async ({
...
+        address: poolLensContractAddress as Address,
+        functionName: 'vTokenBalancesAll',
+        args: [isolatedPoolsVTokenAddresses as Address[], accountAddress as Address],
+      }),
+      venusLensContractAddress
+        ? publicClient.simulateContract({
+            abi: venusLensAbi,
+            address: venusLensContractAddress as Address,
+            functionName: 'vTokenBalancesAll',
+            args: [legacyPoolVTokenAddresses as Address[], accountAddress as Address],
```
gleiser-oliveira: `accountAddress` was updated to the `Address` type, so it looks like it doesn't need the type override

apps/evm/src/clients/api/queries/useGetPools/getPools/getUserPrimeApys/index.ts
```diff
@@ -1,15 +1,27 @@
...
+  primeVTokenAddresses: readonly Address[];
+}) => {
   const primeAprs = await Promise.all(
     primeVTokenAddresses.map(primeVTokenAddress =>
-      primeContract.calculateAPR(primeVTokenAddress, accountAddress),
+      publicClient.readContract({
+        address: primeContractAddress,
+        abi: primeAbi,
+        functionName: 'calculateAPR',
+        args: [primeVTokenAddress, accountAddress as Address],
```
gleiser-oliveira: Would also say this can be `Address`

apps/evm/src/clients/api/queries/useGetPools/getPools/appendPrimeSimulationDistributions/index.ts
```diff
@@ -71,21 +73,26 @@ export const appendPrimeSimulationDistributions = async ({
...
-          averageSupplyBalanceMantissa.toFixed(),
-          averageXvsStakedMantissa.toFixed(),
-        );
+        const simulatedPrimeAprs = await publicClient.readContract({
+          abi: primeAbi,
+          address: primeContractAddress,
+          functionName: 'estimateAPR',
+          args: [
+            primeVTokenAddress,
+            (accountAddress as Address) || NULL_ADDRESS,
```
gleiser-oliveira: What about this one?

apps/evm/src/clients/api/queries/useGetPools/__tests__/__snapshots__/index.prime.spec.ts.snap
```diff
@@ -2876,7 +2876,7 @@ exports[`useGetPools > filters out Prime simulations that are 0 1`] = `
                 "decimals": 6,
                 "symbol": "USDT",
               },
-              "type": undefined,
+              "type": "merkl",
```
gleiser-oliveira, line 2879: I'm in doubt about some of the mocks set as a `merkl` distribution type 🤔, might be worth double checking if the condition is right

apps/evm/src/clients/api/queries/getTokenBalances/__tests__/__snapshots__/index.spec.ts.snap
```diff
@@ -47,7 +47,7 @@ exports[`api/queries/getTokenBalances > returns token balances, including BNB, i
       },
     },
     {
-      "balanceMantissa": "1000000000000000000",
+      "balanceMantissa": "10000",
```
gleiser-oliveira: Not a big deal, but this doesn't need to change if...

apps/evm/src/clients/api/queries/getTokenBalances/__tests__/index.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    });
+
+    expect(res).toMatchSnapshot();
+  });
+
+  it('returns token balances, including native token, in the right format on success', async () => {
+    const fakePublicClient = {
+      readContract: () => 10000n,
+      getBalance: () => 10000n,
+    } as unknown as PublicClient;
```
gleiser-oliveira: ... the mock keeps the previous value `1000000000000000000n`

apps/evm/src/clients/api/queries/getUserVaiBorrowBalance/index.ts
```diff
@@ -0,0 +1,50 @@
+import BigNumber from 'bignumber.js';
+import { vaiControllerAbi } from 'libs/contracts';
+import { VError } from 'libs/errors';
+import type { Address, PublicClient } from 'viem';
+
+export interface GetUserVaiBorrowBalanceInput {
+  accountAddress: string;
+  vaiControllerContractAddress: Address;
```
gleiser-oliveira: Similar to the other comments, I'm in doubt about when `accountAddress`should be `Address` or `string` 🤔

apps/evm/src/clients/api/queries/useGetPools/__tests__/__snapshots__/index.prime.spec.ts.snap
therealemjy: Yeah these ones are correct, I manually updated the mock data to make sure all reward distributors had a type, so that it reflects better what the API would respond (since technically it would never respond `undefined` for the type).

apps/evm/src/clients/api/queries/useGetPools/getPools/index.ts
therealemjy: Yeah makes sense. I originally kept the implementation as it is because it requires a lot of changes to the codebase, but I think we should make the changes at some point anyway so why not do it in this PR.

apps/evm/src/clients/api/queries/getUserVaiBorrowBalance/index.ts
therealemjy: Let's use `Address` everywhere, so we're consistent with `viem`.

apps/evm/src/clients/api/queries/getTokenBalances/__tests__/index.spec.ts
therealemjy: Good point, I made the update.

gleiser-oliveira (approved):

PR #3786, opened 2025-01-23
fix: preventDefault when clicking a Tooltip

gleiser-oliveira:
## Changes

- This allows the `InfoIcon` to show it's `Tooltip` without triggering the surrounding navigation on mobile

### DIFF
--- a/.changeset/fuzzy-suits-collect.md
+++ b/.changeset/fuzzy-suits-collect.md
@@ -1,0 +1,5 @@
+---
+"@venusprotocol/evm": minor
+---
+
+preventDefault when clicking a Tooltip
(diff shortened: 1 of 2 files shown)


apps/evm/src/components/InfoIcon/index.tsx
```diff
@@ -17,7 +17,11 @@ export const InfoIcon = ({
   iconName = 'info',
   iconClassName,
 }: InfoIconProps) => (
-  <Tooltip className={cn('inline-flex', className)} title={tooltip}>
+  <Tooltip
+    className={cn('inline-flex', className)}
+    onClick={e => e.preventDefault()}
```
therealemjy: How about going one level higher and updating the Tooltip component directly so that the children DOM it renders does not propagate click actions (event.stopPropagation())? This way we won't need to add this fix everywhere we render a Tooltip and clicking on a link in its content  will still work without triggering any unwanted behavior.

PR #3799, opened 2025-01-27
feat: highlight markets with boosted APYs

therealemjy:
## Changes

- update `getCombinedDistributionApys` utility function to return total supply and borrow APYs. Also make sure all returned properties are defined (defaulted to 0)
- remove "LTV" info from `MarketTable` container
- update `MarketTable` container to highlight markets with boosted APYs
- rename `rewardDescription` property of `MerklDistribution` type to `description`
- add optional `claimUrl` property to `MerklDistribution` type, coming from the API


apps/evm/src/containers/MarketTable/Apy/BoostBadge/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+        name: d.description || t('marketTable.apy.primeBadge.tooltip.defaultMerklRewardName'),
+        value: formatPercentageToReadableValue(d.apyPercentage),
+        description: (
+          <Trans
+            i18nKey="marketTable.apy.primeBadge.tooltip.externalRewardDescription"
+            components={{
+              AppLink: (
+                <Link
+                  target="_blank"
+                  href={d.claimUrl || 'https://merkl.angle.money/'}
```
gleiser-oliveira: About the claimUrl, if the type is merkl there is always a claimUrl (could be either Ignite or Merkl currently). If we keep the fallback, I'd use https://app.merkl.xyz/

gleiser-oliveira: I'd also think how this would look like if there were multiple distributions with the same URL, I'd think this would start looking repetitive

therealemjy: Ok I see, good to know. No need for a fallback in this case.

The URL won't be visible, as it will be hidden under an "official app" text link. It is true that the same message would be shown multiple times if there were multiple Merkl rewards, but I don't think this is necessarily a bad thing as this way users will be used to it. An alternative would be to have a different UI for when there are multiple Merkl rewards, but to be honest I don't know how often this will be the case and so whether it is worth it (we'd also have to think about cases like the Ignite campaign where although it is a Merkl campaign, rewards are claimed through a separate app).

gleiser-oliveira (approved):

Frayst2:
IMG_0148

Render of stars logo looks weird on mobile. I assume It should be on the left or right of APY value

therealemjy:
@Frayst2 Agreed. After speaking with our designer, we will use an alternative UI and show everything in one line.

PR #3818, opened 2025-01-29
refactor: rework tooltip component

gleiser-oliveira:
## Changes

- add @radix-ui/react-tooltip
- Rework Tooltip: replaced MUITooltip, fixed mobile tooltips by opening a modal with its contents, replaced emotion styles with tailwind


apps/evm/src/components/Tooltip/index.tsx
```diff
@@ -1,22 +1,87 @@
...
+>(({ className, sideOffset = 4, ...props }, ref) => (
+  <Portal>
+    <Content
+      ref={ref}
+      onClick={e => e.stopPropagation()}
+      sideOffset={sideOffset}
+      className={cn(
+        'flex z-50 overflow-hidden rounded-xl bg-lightGrey p-2 max-w-75 text-sm text-primary-foreground animate-in fade-in-0 zoom-in-95',
+        'data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[side=bottom]:slide-in-from-top-2',
+        'data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2',
```
therealemjy: Though I can see the classes to animate the tooltip, the animation doesn't seem to work:

https://github.com/user-attachments/assets/6e192f4a-cd20-47ee-8276-a827248a79af

apps/evm/src/components/Tooltip/index.tsx
```diff
@@ -1,22 +1,87 @@
...
-  children: React.ReactNode;
+export interface TooltipProps extends TooltipPrimitiveProps {
+  className?: string;
+  content: string | React.ReactElement;
+  side?: TooltipContentProps['side'];
 }
 
-export const Tooltip = ({ children, placement = 'top', ...rest }: TooltipProps) => {
-  const styles = useStyles();
+const TooltipContent = React.forwardRef<
```
therealemjy: To be consistent with how other components are built, I'd recommend moving this into its own subdirectory (inside the Tooltip directory).

apps/evm/src/containers/MarketTable/Apy/ApyWithPrimeSimulationBoost/index.tsx
therealemjy, line 1: This file has been removed from the repository.

apps/evm/src/containers/MarketTable/Apy/ApyWithPrimeBoost/index.tsx
therealemjy, line 1: This file has been removed from the repository.

apps/evm/src/containers/MarketTable/Apy/PrimeBadge/index.tsx
```diff
@@ -38,7 +38,7 @@ export const PrimeBadge: React.FC<PrimeBadgeProps> = ({
   return (
     <Tooltip
```
therealemjy, line 39: Clicking on this tooltip on mobile activates the onClick event of the parent, redirecting the user to the Market page.

apps/evm/src/components/Tooltip/index.tsx
```diff
@@ -1,22 +1,87 @@
...
+  </Portal>
+));
+TooltipContent.displayName = Content.displayName;
+
+export const Tooltip = ({
+  className,
+  content,
+  children,
+  delayDuration = 200,
+  side,
```
therealemjy: I don't think either of these properties are needed: both should be constant across the codebase to make the UI look consistent.

apps/evm/src/components/Tooltip/index.tsx
```diff
@@ -1,22 +1,87 @@
...
+              setIsTooltipOpened(true);
+              e.preventDefault();
+            }
+          }}
+        >
+          <div>{children}</div>
+        </Trigger>
+        <TooltipContent side={side} className={cn('block p-3', !isMdOrUp && 'hidden')}>
+          {content}
+          <Arrow className="fill-lightGrey" width={14} height={7} />
```
therealemjy: nit: could you define the width and height using classes instead?
```suggestion
          <Arrow className="fill-lightGrey w-[14px] h-[7px]" />
```

therealemjy (commented):
Clicking on the children of a tooltip, on tablet and up, hides the tooltip itself.

https://github.com/user-attachments/assets/440a7db6-941d-4b27-a9cd-9ed8ba1fdc6c

apps/evm/src/components/Tooltip/index.tsx
gleiser-oliveira: `side` is replacing the `placement` prop that tooltip used to have ([defaults to top](https://github.com/VenusProtocol/venus-protocol-interface/pull/3818/files#diff-0f120b659412e76d2f4b4cd0c78e5b574db4f90a4b8e08e52749aa900695fe55L11)). `delayDuration` is the time that it takes for the tooltip to pop up, removing it would mean using the default from `TooltipPrimitive` which I feel is too long. What do you think?

therealemjy: I suggest using constants for both. So instead of using the default delay from `TooltipPrimitive`, pass 0 (or whatever number you feel works best) to it. Same thing for `side`, I'd pass `top` all the time.

apps/evm/src/components/Tooltip/index.tsx
```diff
@@ -1,22 +1,60 @@
...
+        </TooltipContent>
+      </TooltipPrimitive>
+
+      <Modal
+        buttonClassName="right-3"
+        onClick={e => e.stopPropagation()}
+        isOpen={isTooltipOpened && !isMdOrUp}
+        handleClose={handleToggleDropdown}
+      >
+        <div className="flex justify-center pt-3">{content}</div>
```
therealemjy:
```suggestion
        <div className="pt-3">{content}</div>
```

This is to prevent messing up with the content passed. e.g.:

Screenshot 2025-02-07 at 16 01 53

therealemjy (approved):

PR #3819, opened 2025-01-29
refactor: improve APY boost feature

therealemjy:
## Changes

- update UI of APY boost feature
- remove "helper" cursors


gleiser-oliveira:
Honestly I don't think that grouping the supply/borrow APY under "token distributions" really fits, because it's not a distribution/reward that boosts your earnings.

distrib

If we want to have it inside the tooltip, I'd put it above the "token distributions" section.

therealemjy:
@gleiser-oliveira I get your point, though I think separating the base APY into its own group would be cumbersome. I've proposed Pavel to just remove the "Token distributions" title, let's see what he says.

therealemjy:
@gleiser-oliveira Pavel accepted to remove the title altogether:
Screenshot 2025-01-31 at 14 20 26

gleiser-oliveira (approved):

PR #3922, opened 2025-02-20
feat: update behavior to switch chains

therealemjy:
## Changes

- refactor `ConnectWallet` container to remove logic handling chain switching
- create `SwitchChain` container to prompt user to connect to the wanted chain if they aren't already
- update `AuthHandler` container so that users don't get automatically disconnected when they are connected to the wrong chain
- update transaction forms to prompt users to connect to the correct chain before submitting
- create `useAccountChainId` hook
- refactor `Notice` component to use nested divs instead of `::before` pseudo-element to apply background color
- refactor logic to automatically scroll to the top of the page on navigation, so that it only triggers when URL changes and before browser repaints the screen
- apply 50% opacity to `Toggle` component when it's disabled


gleiser-oliveira:
@therealemjy what about transactions that don't have a confirmation button, like the collateral switch? If I change the chain in the wallet UI and try to click on the collateral switch, it now generates an error.

therealemjy:
@gleiser-oliveira That's a very good point. Let me speak with Pavel to see what solution we can come up with.

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -810,6 +803,13 @@
     },
     "walletBalance": "Wallet balance"
   },
+  "switchChain": {
+    "switchButton": "Switch to {{chainName}}"
+  },
+  "switchChainNotice": {
+    "buttonLabel": "Switch to {{chainName}}",
+    "description": "You will need to switch to {{chainName}} in order to enable or disable collaterals"
```
gleiser-oliveira: Should we be clearer about this being the chain selected on the wallet's UI?

gleiser-oliveira (approved):

PR #3942, opened 2025-02-26
feat: integrate BSC Core pool subgraph

gleiser-oliveira:
## Jira ticket(s)

VEN-3078

## Changes

- Added the BSC Core Pool subraph, it is now used to unify fetching the counts of borrowers and suppliers along with the isolated pools


apps/evm/src/clients/api/queries/useGetPools/__tests__/__snapshots__/index.spec.ts.snap
```diff
@@ -1090,7 +1090,7 @@ exports[`useGetPools > returns pools with time based reward rates in the correct
           "liquidityCents": "1.000006200805040439464388948970138750743214608e+23",
           "reserveFactor": 1,
           "reserveTokens": "0",
-          "supplierCount": 10,
+          "supplierCount": 0,
```
therealemjy, line 1093: It looks like this isn't getting mocked

apps/evm/src/clients/api/queries/useGetPools/getPools/getParticipantCounts/index.ts
```diff
@@ -0,0 +1,31 @@
...
+      pool.markets.forEach(market => {
+        participantsCountMap.set(market.id.toLowerCase(), {
+          borrowerCount: +market.borrowerCount,
+          supplierCount: +market.supplierCount,
+        });
+      }),
+    );
+
+    return { participantsCountMap };
+  }
```
therealemjy: The IF condition means participants counts will only be fetched on BNB Chain mainnet and testnet, while I believe what you want is to call `getBscCorePoolParticipantsCount` only on those chains instead. You can also batch the two requests in a `Promise.allSettled` (and not a `Promise.all`, so that all calls are completed even if some fail) call and make the code a bit more DRY:

```suggestion
  const participantsCountMap = new Map<string, MarketParticipantsCounts>();

  const [isolatedPoolParticipantsCountsResult, bscCorePoolParticipantsCountsResult] =
    await Promise.allSettled([
      getIsolatedPoolParticipantsCount({ chainId }),
      chainId === ChainId.BSC_MAINNET || chainId === ChainId.BSC_TESTNET
        ? getBscCorePoolParticipantsCount({ chainId })
        : undefined,
    ]);

  if (isolatedPoolParticipantsCountsResult.status === 'rejected') {
    logError(isolatedPoolParticipantsCountsResult.reason);
  }

  if (bscCorePoolParticipantsCountsResult.status === 'rejected') {
    logError(bscCorePoolParticipantsCountsResult.reason);
  }

  const isolatedPoolParticipantsCounts = (
    extractSettledPromiseValue(isolatedPoolParticipantsCountsResult)?.pools || []
  ).flatMap(p => p.markets);

  const bscCorePoolParticipantsCounts =
    extractSettledPromiseValue(bscCorePoolParticipantsCountsResult)?.markets || [];

  [...isolatedPoolParticipantsCounts, ...bscCorePoolParticipantsCounts].forEach(market => {
    participantsCountMap.set(market.id.toLowerCase(), {
      borrowerCount: +market.borrowerCount,
      supplierCount: +market.supplierCount,
    });
  });

  return { participantsCountMap };
```

Since this code uses `Promise.allSettled`, you would be able to remove the `safeGetParticipantCounts` function.

apps/evm/src/clients/api/queries/useGetPools/getPools/safeGetParticipantCounts/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  } catch (error) {
+    // Safari throws a "TypeError: Load failed" error if the fetch is canceled
+    // e.g., if the user navigates away from the page before the request is finished
+    // we can safely filter them out from being logged
+    if (error instanceof Error && error.name === 'TypeError' && error.message === 'Load failed') {
+      return undefined;
+    }
+    // Log error without throwing to prevent the entire query from failing, since this relies on a
+    // third-party service that could be down and doesn't constitute a critical failure
+    logError(error);
```
therealemjy: I was thinking about this more; how about adding the IF condition directly inside the `logError` function? This way we'll apply this filter to all errors sent from the dApp.

apps/evm/src/clients/subgraph/__mocks__/index.ts
```diff
@@ -13,4 +15,6 @@ export const getBscProposal = vi.fn(async () => ({
 
 export const getNonBscProposals = vi.fn(async () => nonBscProposalsResponse);
 
+export const getBscCorePoolParticipantsCount = vi.fn(async () => bscCorePoolParticipantsCount);
+
 export const getIsolatedPoolParticipantsCount = vi.fn();
```
therealemjy: I'd suggest following the same approach to mock each call (here we have two different approaches: `getIsolatedPoolParticipantsCount` is mocked to returned `undefined` and its returned value is later mocked in the tests directly, while `getBscCorePoolParticipantsCount` is mocked to return values).

It might be useful to dynamically mock the returned values based on the `__mocks__/api/pools.json` file content, since it's what we use as mock data throughout tests.

apps/evm/src/clients/api/queries/useGetPools/__tests__/__snapshots__/index.spec.ts.snap
gleiser-oliveira: After reviewing this, I'd actually say this is right.

Or maybe not right per se, but it's more precise: this test mocks the chain to be Arbitrum, becase we want to test time based reward rates. But the other depending mocks are all BSC based: tokens, pools, participant counts, etc. I'd say that all those would have to be mocked differently so that the resulting snapshot makes better sense.

Even better would be not snapshotting the whole pools array, just the time based reward rates. I feel both are kind of out of scope from this PR, but let me know if I should address this here or in a new one.

therealemjy: Ok I see, that makes sense.
I'm ok with leaving the PR like this for now, I agree it's out of scope.

therealemjy (approved):

PR #4012, opened 2025-03-13
feat: use price endpoint in getPendingRewards

gleiser-oliveira:
## Jira ticket(s)

VEN-3136

## Changes

- Use the price endpoint to fetch reward token prices


apps/evm/src/clients/api/queries/getPendingRewards/index.ts
```diff
@@ -114,18 +114,15 @@ const getPendingRewards = async ({
...
     ...primeRewardTokenAddresses,
   ]);
-  const rewardTokenPricesPromises = Promise.allSettled(
-    rewardTokenAddresses.map(rewardTokenAddress =>
-      resilientOracleContract.getPrice(rewardTokenAddress),
-    ),
-  );
-
-  const rewardTokenPricesResults = await rewardTokenPricesPromises;
+  const tokenPriceMantissaMapping = await getApiTokenPrice(rewardTokenAddresses);
```
therealemjy: Rather than passing `getApiTokenPrice` through props, it should be imported (so that we follow the codebase's practices). You can then remove the `useGetApiTokenPrice` hook, and mock the function in tests.

apps/evm/src/clients/api/queries/getPendingRewards/index.ts
```diff
@@ -114,18 +114,15 @@ const getPendingRewards = async ({
...
+  const tokenPriceMapping: Record<string, BigNumber> = Object.entries(
+    tokenPriceMantissaMapping,
+  ).reduce<{
     [address: string]: BigNumber;
-  }>((acc, rewardTokenPricesResult, index) => {
-    const rewardTokenAddress = rewardTokenAddresses[index];
+  }>((acc, tokenPriceMantissaTuple) => {
+    const rewardTokenAddress = tokenPriceMantissaTuple[0];
+    const rewardTokenPriceMantissa = tokenPriceMantissaTuple[1];
     const rewardToken = findTokenByAddress({
```
therealemjy, line 130: We can remove this check actually. We are sending the list of tokens we want to fetch the prices of, so by default the list is verified. Even if an unlisted token was added to the mapping, it wouldn't have any impact since we don't use the mapping as a list of tokens but just a reference for prices.

apps/evm/src/clients/api/queries/getApiTokenPrice/index.ts
therealemjy, line 1: nit: since you're fetching multiple prices, the function should probably be named `getApiTokenPrices` (same change for the type names).

apps/evm/src/clients/api/queries/getPendingRewards/index.ts
gleiser-oliveira: You mean findTokenByAddress? I need to use it to find the token record and then use its decimals to calculate the USD value.

apps/evm/src/clients/api/queries/getApiTokenPrice/index.ts
gleiser-oliveira: Hmm, I actually see it as "get each token price" instead of "get the prices for the list of tokens" 😆

apps/evm/src/clients/api/queries/getPendingRewards/index.ts
therealemjy: Ah yes, my bad

therealemjy (approved):

PR #4026, opened 2025-03-17
feat: update zkSync name to ZKsync

therealemjy:
## Changes

- update zkSync name to ZKsync


apps/evm/CHANGELOG.md
gleiser-oliveira, line 1: It's a nitpick, but this file should probably not have manual changes

therealemjy: Good point

gleiser-oliveira (approved):

PR #4033, opened 2025-03-18
fix: proposal action parsing

therealemjy:
## Changes

- fix how proposal actions are parsed, so that all primitives are supported


apps/evm/src/pages/Governance/ProposalList/CreateProposalModal/proposalSchema.ts
```diff
@@ -56,10 +56,12 @@ const proposalSchema = yup.object({
                       // @ts-expect-error The yup type doesn't show this value exists but it does
                       parseFunctionSignature(this.options.from[0].value.signature)?.inputs || [];
 
+                    console.log(value);
```
chechu: Do we want to print this?

chechu (approved):

PR #4039, opened 2025-03-19
refactor: use viem in some query functions

therealemjy:
## Jira ticket(s)

VEN-3025

## Changes

- refactor some query functions to use viem instead of ethers.js


apps/evm/src/clients/api/queries/getPoolLiquidationIncentive/index.ts
```diff
@@ -1,23 +1,30 @@
...
-  poolComptrollerContract,
+  publicClient,
+  poolComptrollerContractAddress,
 }: GetPoolLiquidationIncentiveInput): Promise<GetPoolLiquidationIncentiveOutput> => {
-  const res = await poolComptrollerContract.liquidationIncentiveMantissa();
+  const liquidationIncentiveMantissa = await publicClient.readContract({
+    address: poolComptrollerContractAddress,
+    abi: legacyPoolComptrollerAbi,
+    functionName: 'liquidationIncentiveMantissa',
+  });
```
gleiser-oliveira: This was compatible with both legacy and isolated pool, isn't it necessary anymore? Wouldn't we prefer to pass the isolated pools ABI?

apps/evm/src/clients/api/queries/getPrimeToken/useGetPrimeToken.ts
```diff
@@ -30,14 +31,20 @@ type Options = QueryObserverOptions<
...
   return useQuery({
     queryKey: [FunctionKey.GET_PRIME_TOKEN, { ...input, chainId }],
 
     queryFn: () =>
-      callOrThrow({ primeContract, accountAddress: input.accountAddress }, params =>
-        getPrimeToken(params),
+      callOrThrow(
+        {
+          primeContractAddress: primeContractAddress as Address,
+          accountAddress: input.accountAddress as Address,
```
gleiser-oliveira: Aren't those already typed as `Address`?

apps/evm/src/clients/api/queries/getProposalEta/index.ts
```diff
@@ -10,13 +12,19 @@ export type GetProposalEtaOutput = {
...
+  const [_id, _proposerAddress, etaSeconds] = await publicClient.readContract({
+    address: governorBravoDelegateContractAddress,
+    abi: governorBravoDelegateAbi,
+    functionName: 'proposals',
+    args: [BigInt(proposalId)],
+  });
 
   // Convert ETA expressed in seconds to milliseconds
-  const eta = new Date(+resp.eta * 1000);
+  const eta = new Date(Number(etaSeconds) * 1000);
```
gleiser-oliveira, line 27: I like this change, `+string` to turn a string into a number is just JS wizardry

apps/evm/src/clients/api/queries/getVoterHistory/types.ts
```diff
@@ -3,7 +3,7 @@ import type { Address, ByteArray } from 'viem';
 
 export interface ProposalActionApiResponse {
   actionIndex: number;
-  calldata: `0x${string}` | ByteArray;
+  calldata: Address | ByteArray;
```
gleiser-oliveira: Hmm, I feel this is better typed as `Hex`. Both are `0x${string}` at the end of the day, but this isn't really an `Address`.

apps/evm/src/clients/subgraph/utilities/formatToProposal/formatToProposalActions/index.ts
```diff
@@ -1,13 +1,13 @@
...
+import type { Address, ByteArray } from 'viem';
 
 export const formatToProposalActions = ({
   callDatas,
   signatures,
   targets,
   values,
 }: {
-  callDatas: (`0x${string}` | ByteArray)[];
+  callDatas: (Address | ByteArray)[];
```
gleiser-oliveira: Same

apps/evm/src/clients/subgraph/utilities/formatToProposal/formatToRemoteProposal/index.ts
```diff
@@ -28,7 +28,7 @@ export const formatToRemoteProposal = ({
   proposalState: ProposalState;
   layerZeroChainId: number;
   proposalId: number;
-  callDatas: (`0x${string}` | ByteArray)[];
+  callDatas: (Address | ByteArray)[];
```
gleiser-oliveira: Same

apps/evm/src/clients/api/queries/getPrimeToken/useGetPrimeToken.ts
therealemjy: They are, that's an oversight on my end

apps/evm/src/clients/api/queries/getProposalEta/index.ts
therealemjy: Hehe sometimes I remember that some primitives can be used as functions.

apps/evm/src/clients/api/queries/getVoterHistory/types.ts
therealemjy: Good point.

apps/evm/src/clients/api/queries/getPoolLiquidationIncentive/index.ts
therealemjy: Because this part is shared between the legacy and the isolated `PoolComptroller` ABIs, then this function still supports legacy and isolated pools.
I've updated the code anyway to pass the isolated `PoolComptroller` ABI, since in similar situations we've tended to lean towards using the isolated pool implementation.

gleiser-oliveira (approved):

PR #4040, opened 2025-03-19
fix: alignment issue on pool tables on Account page

therealemjy:
## Changes

- fix alignment issue on pool tables on Account page


gleiser-oliveira:
@therealemjy I think this can have a fix changeset

gleiser-oliveira (approved):

PR #4063, opened 2025-03-25
feat: enable dApp to be embedded in Safe

therealemjy:
## Jira ticket(s)

VEN-3093

### evm
- add `manifest.json` file
- add `network`, defaulted to BNB Chain testnet, to mock provider
- update `borrow` and `withdraw` mutation functions to add an access list when it concerns a native token. This is a workaround for the "transfer" and "send" actions being limited to 2300 gas, which is insufficient post-Berlin fork ([EIP-2929](https://eips.ethereum.org/EIPS/eip-2929))
- make `description` prop of `Notice` component optional
- add `disabled` optional prop to `Select` component
- disable chain selector and remove chevron icon when running in Safe
- disable "from chain" field on Bridge page when running in Safe
- remove "Switch to BNB" button on Governance and Proposal pages when running in Safe
- remove option to disconnect wallet when running in Safe
- update `useChainId` hook so the wallet chain ID becomes the only source of truth when running in Safe
- hide notification toasts when running in Safe
- add `isSafeApp` prop to config
- fix typo (accoutChainId -> accountChainId)
- update `useSendTransaction` hook to support Safe transactions

### How to test the embedded app in Safe

In Safe, you can add a custom app using the URL of one of the test environments set up for this branch (e.g.: https://appvenusio-git-feat-safe-venus-protocol.vercel.app). A tutorial explaining this process can be found [here](https://help.safe.global/en/articles/40859-add-a-custom-safe-app).


packages/chains/src/chainMetadata.ts
```diff
@@ -87,6 +89,7 @@ export const chainMetadata: {
     logoSrc: ethLogo,
     explorerUrl: 'https://sepolia.etherscan.io',
     layerZeroScanUrl: 'https://testnet.layerzeroscan.com',
+    safeWalletApiUrl: 'https://safe-transaction-base-sepolia.safe.global',
```
gleiser-oliveira:
```suggestion
    safeWalletApiUrl: 'https://safe-transaction-sepolia.safe.global',
```

therealemjy: Good catch, thanks

gleiser-oliveira (approved):

PR #4094, opened 2025-03-31
refactor: some query functions to use viem 

therealemjy:
## Jira ticket(s)

VEN-3025
## Changes

- refactor some query functions to use viem


apps/evm/src/clients/api/queries/getVTokenApySimulations/__tests__/__snapshots__/index.spec.ts.snap
gleiser-oliveira, line 1: Since the mocks for the borrow and supply rates being the same now, the snapshot now has the same value for both. Wouldn't it potentially hide issues? Say that by accident the variables in `getVTokenApySimulations/formatToApySnapshots.ts` were swapped or repeated.

therealemjy: That's a good point. I've updated this test so that the mocked borrow rate is different from the mocked supply rate.

gleiser-oliveira (approved):

PR #4288, opened 2025-05-16
refactor: mutation functions to use viem

therealemjy:
## Jira ticket(s)

VEN-3025

- refactor mutation functions to use viem


apps/evm/src/pages/Swap/__tests__/index.spec.tsx
```diff
@@ -672,10 +676,39 @@ describe('Swap', () => {
     fireEvent.click(submitButton!);
 
     // Check swap was executed
-    await waitFor(() => expect(swapTokens).toHaveBeenCalledTimes(1));
-    expect(swapTokens).toHaveBeenCalledWith({
-      swap: fakeExactAmountInSwap,
-    });
+    await waitFor(() => expect(mockSwapTokens).toHaveBeenCalledTimes(1));
+    expect(mockSwapTokens.mock.calls[0]).toMatchInlineSnapshot(`
```
gleiser-oliveira: nit: wouldn't it be better to just snapshot these instead of having them inline?

therealemjy: Yeah you're right. When the content is small I tend to use inline snapshots as I feel it makes the code more eligible, but in this case there's too much content. I'll update the test.

apps/evm/src/clients/api/mutations/useBorrow/__tests__/index.spec.ts
gleiser-oliveira, line 1: Long inline snapshots

apps/evm/src/clients/api/mutations/useSwapTokens/__tests__/index.spec.ts
gleiser-oliveira, line 1: Long inline snapshots

apps/evm/src/clients/api/mutations/useWithdraw/__tests__/index.spec.ts
gleiser-oliveira, line 1: Here too

apps/evm/src/pages/Swap/__tests__/index.spec.tsx
gleiser-oliveira: I pointed out some other files that have them, other than that PR looks good!

therealemjy: Thanks, let me update those too

apps/evm/src/clients/api/mutations/useSwapTokens/__tests__/index.spec.ts
therealemjy: I'd keep these as inline snapshots actually. I feel they're reasonable in size and it's much easier to check the test is actually looking at the right thing.

therealemjy:
@gleiser-oliveira I updated the tests to use snapshots when inline snapshots would be too big. I've kept inline snapshots for the checks I feel are the most important and that don't take too much space.

gleiser-oliveira (approved):

PR #4314, opened 2025-05-22
fix: find addresses in getContractAddress

gleiser-oliveira:


apps/evm/src/clients/api/mutations/useRepay/index.ts
```diff
@@ -118,6 +118,10 @@ export const useRepay = (options?: Partial<Options>) => {
         input.wrap &&
         (!nativeTokenGatewayContractAddress || !input.vToken.underlyingToken.tokenWrapped)
       ) {
+        console.log({
+          nativeTokenGatewayContractAddress,
+          input,
+        });
```
therealemjy: Let's remove this

therealemjy (approved):

PR #4315, opened 2025-05-22
refactor: better types for getContractAddress

gleiser-oliveira:
- Added a bit extra type safety for `getContractAddress`

### DIFF
--- a/apps/evm/src/containers/ReadableActionSignature/getContractName.ts
+++ b/apps/evm/src/containers/ReadableActionSignature/getContractName.ts
@@ -34,11 +34,7 @@ const getContractName = ({ target, vTokens, tokens, chainId }: GetContractNameIn
   // Search within contracts
   const matchingUniqueContractInfo = Object.entries(addresses.uniques).find(
     ([_uniqueContractName, address]) => {
-      let contractAddress: string | Record<number, string> = '';
-
-      if (Object.prototype.hasOwnProperty.call(address, chainId)) {
-        contractAddress = address[chainId as keyof typeof address];
-      }
+      const contractAddress = address[chainId];
 
       if (!contractAddress) {
         return false;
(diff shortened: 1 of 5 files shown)


apps/evm/src/libs/contracts/utilities/getContractAddress/index.ts
```diff
@@ -13,14 +13,14 @@ export type GetUniqueContractAddressInput = {
 export type GetUniquePerPoolContractAddressInput = {
   name: UniquePerPoolContractName;
   chainId: ChainId;
-  poolComptrollerContractAddress: string;
+  poolComptrollerContractAddress: Address;
```
therealemjy, line 16: That's a nice touch.

apps/evm/src/libs/contracts/utilities/getContractAddress/index.ts
```diff
@@ -13,14 +13,14 @@ export type GetUniqueContractAddressInput = {
...
 };
 
 export type GetContractAddressInput =
   | GetUniqueContractAddressInput
   | GetUniquePerPoolContractAddressInput;
 
-export const getContractAddress = (input: GetContractAddressInput) => {
+export const getContractAddress = (input: GetContractAddressInput): Address | undefined => {
```
therealemjy: Doing this is riskier because you cast the return type of the whole function. Ideally we won't to cast as less as possible of the type, hence my original implementation.

gleiser-oliveira: Removed this and the `keyof typeof` casts in favor of generating strict types. The only exception now is the `toLowerCase()` call, which makes the key get typed to `string` instead of `Address`.

apps/evm/src/containers/ReadableActionSignature/getContractName.ts
```diff
@@ -48,11 +44,6 @@ const getContractName = ({ target, vTokens, tokens, chainId }: GetContractNameIn
         // Handle unique contracts
         return areAddressesEqual(contractAddress, target);
       }
-
-      // Handle SwapRouter contract
-      return Object.values(contractAddress).some(swapRouterAddress =>
-        areAddressesEqual(swapRouterAddress, target),
-      );
```
gleiser-oliveira, line 55: AFAIU this wasn't even possible before. On line 35 we only consider the entries from `uniques` and `SwapRouter` is not there.

therealemjy (approved):
Really nice! I like this implementation a lot, and on top of that it highlighted a bug in the codebase. Well done 👍

PR #4389, opened 2025-06-11
feat: add UI for the Import Positions modal

therealemjy:
## Jira ticket(s)

VEN-3235

- add `ImportPositionsModal` container
- add `importPositions` feature flag
- remove outdated comments
- fix `className` prop of `Apy` component
- rename `lightening` icon to `lightning`
- add graph, shield2 and lightning2 icons
- update `Modal` component container to use Tailwind classes instead of Emotion CSS

The Import Positions modal is currently enabled on testnet and automatically opens on launch, displaying fake opportunities. The following PRs will fetch actual opportunities and allow users to migrate their positions to Venus.


apps/evm/src/libs/translations/translations/en.json
```diff
@@ -364,6 +364,40 @@
...
+      "1": {
+        "description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
+        "title": "Better yields"
+      },
+      "2": {
+        "description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
+        "title": "Fastest growing protocol"
+      },
+      "3": {
+        "description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
```
gleiser-oliveira, line 383: Are these yet to be defined?

apps/evm/src/hooks/useIsFeatureEnabled/index.tsx
```diff
@@ -110,6 +110,16 @@ export const featureFlags = {
...
+  importPositions: [
+    ChainId.BSC_TESTNET,
+    ChainId.SEPOLIA,
+    ChainId.OPBNB_TESTNET,
+    ChainId.ARBITRUM_SEPOLIA,
+    ChainId.ZKSYNC_SEPOLIA,
+    ChainId.OPTIMISM_SEPOLIA,
+    ChainId.BASE_SEPOLIA,
+    ChainId.UNICHAIN_SEPOLIA,
+  ],
```
gleiser-oliveira, line 122: nit: one thing I'd start doing in this file at is to have lists like `testnets`, `mainnets` and `allChains`, it would be easier to mantain this and other feature flags

gleiser-oliveira (commented):
What about adding a temporary button to open the modal instead of opening it automatically when going to testnet? I imagine this would be a bit annoying if people just want to navigate to the governance page, for example

apps/evm/src/libs/translations/translations/en.json
therealemjy: Yes

apps/evm/src/hooks/useIsFeatureEnabled/index.tsx
therealemjy: Yeah that's a good point, I can do that in a separate PR (just to keep the scope of this one small).

therealemjy:
This PR is just the UI base for this modal, so this behavior won't be on testnet for very long at all. I'm already working on the following PR that will fetch actual positions and only show the modal when there's real opportunities.

gleiser-oliveira (approved):

PR #4435, opened 2025-06-23
feat: wire up Import Positions feature

therealemjy:
## Jira ticket(s)

VEN-3225

- add `@biconomy/abstractjs-canary` and `@rhinestone/module-sdk` libraries to add support for sending transactions via Biconomy
- add `VITE_BICONOMY_API_KEY` env variable
- add `useShouldDisplayImportUi` hook
- add `ImportPositionsModal` container
- add `importablePositions` mock model
- add `getImportablePositions` query function
- add `useGetProfitableImports` hook
- add `useImportSupplyPosition` mutation function
- invalidate `GET_TOKEN_ALLOWANCE` query cache after supplying tokens
- add `Import` page that is only displayed if user has importable positions
- enable `importPositions` on mainnet networks supported by Biconomy
- update `useSendTransaction` hook to handle sending transactions via Biconomy
- add records of `AaveUiPoolDataProvider`, `AavePoolAddressesProvider` and `AaveV3Pool` contracts
- add `useMeeClient` hook to wallet lib
- update `formatPercentageToReadableValue` utility function to use `ROUND_HALF_UP` rounding method
- update `generateExplorerUrl` utility function to support Biconomy transactions
- add `getBoostedAssetSupplyApy` utility function


apps/evm/src/libs/wallet/hooks/useMeeClient/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+      }
+
+      const transport = http(config.rpcUrls[chainId][0]);
+      const chain = chainMapping[chainId];
+
+      const nexusAccount = await toMultichainNexusAccount({
+        chains: [chain],
+        transports: [transport],
+        signer: walletClient,
+      });
```
gleiser-oliveira, line 61: I'm assuming this would be more reliable:

```suggestion
      const transports = config.rpcUrls[chainId].map(rpcUrl => http(rpcUrl));
      const chain = chainMapping[chainId];

      const nexusAccount = await toMultichainNexusAccount({
        chains: [chain],
        transports,
        signer: walletClient,
      });
```

apps/evm/src/clients/api/queries/getImportablePositions/getImportableAaveSupplyPositions/index.ts
```diff
@@ -0,0 +1,142 @@
...
+        const userABalances = userATokenBalanceMapping[tokenAddress];
+
+        if (userABalances.supplyBalanceMantissa === 0n) {
+          return;
+        }
+
+        const userSupplyBalanceMantissa = BigInt(
+          new BigNumber(Number(userABalances.supplyBalanceMantissa))
+            .multipliedBy(Number(liquidityIndex))
+            .dividedBy(Number(RAY))
```
gleiser-oliveira: Is it possible to cause any precision loss by converting these values to numbers? Would it be safer to pass them as strings `bigInt.toString()`?

apps/evm/src/clients/api/queries/getImportablePositions/getImportableAaveSupplyPositions/index.ts
```diff
@@ -0,0 +1,142 @@
...
+            .toFixed(0, BigNumber.ROUND_DOWN),
+        );
+
+        // Skip positions for which pool doesn't have enough liquidity available to withdraw
+        if (availableLiquidity < userSupplyBalanceMantissa) {
+          return;
+        }
+
+        // Convert rate to APY
+        const ratePerSecond = Number(liquidityRate) / Number(RAY);
```
gleiser-oliveira, line 118: Same here (`RAY` could maybe be a number right from its creation, no?)

apps/evm/src/store/index.ts
```diff
@@ -36,25 +37,22 @@ const useStore = create<State>()(
...
+                ...acc,
+                [chainId]: {
+                  ...state.userSettings[chainId],
+                  ...settings,
+                },
+              }),
+              {},
+            ),
+          },
+        })),
```
gleiser-oliveira: nit: maybe we should consider something like https://zustand.docs.pmnd.rs/integrations/immer-middleware

apps/evm/src/libs/wallet/hooks/useMeeClient/index.tsx
therealemjy: Unfortunately the `toMultichainNexusAccount` function only accepts one transport per chain. It's counter intuitive, but the `transports` array needs to match the `chains` array so that each chain gets a transport. Hopefully they'll support setting multiple transports for the same chain to make this more reliable.

apps/evm/src/clients/api/queries/getImportablePositions/getImportableAaveSupplyPositions/index.ts
therealemjy: You're right that we lose some precision by converting these bigint to numbers. Using BigNumber.js and the `toString` function of `BigInt` should indeed fix the issue.

apps/evm/src/clients/api/queries/getImportablePositions/getImportableAaveSupplyPositions/index.ts
therealemjy: In this case, losing precision isn't really an issue because we're calculating an approximate APY.

I think `RAY` should remain a bigint, so that we can keep as much precision as possible when calculating the user supply balance in the block above.

apps/evm/src/store/index.ts
therealemjy: I didn't know about Immer, that's nice. I've updated the logic to use it, it's indeed cleaner.

gleiser-oliveira (approved):

PR #4465, opened 2025-06-30
feat: Alchemy RPC on mainnets

gleiser-oliveira:


apps/evm/src/config/rpcUrls.ts
```diff
@@ -6,6 +6,7 @@ export const rpcUrls: {
   [chainId in ChainId]: string[];
 } = {
   [ChainId.BSC_MAINNET]: [
+    `https://bnb-mainnet.g.alchemy.com/v2/${envVariables.VITE_ALCHEMY_API_KEY}`,
```
therealemjy, line 9: Alchemy should be set as the second available option, otherwise it will be used as the main RPC provider

therealemjy (approved):

PR #4500, opened 2025-07-08
feat: use new price structure from the API

gleiser-oliveira:
## Jira ticket(s)

VEN-3275

- This PR updates the way we fetch token prices from the API, considering the `TokenMetadata` and `TokenPrice` structures


apps/evm/src/clients/api/queries/getPendingRewards/index.ts
```diff
@@ -245,7 +245,10 @@ export const getPendingRewards = async ({
     [address: string]: BigNumber;
   }>((acc, tokenPriceMantissaTuple) => {
     const rewardTokenAddress = tokenPriceMantissaTuple[0];
-    const rewardTokenPriceMantissa = tokenPriceMantissaTuple[1];
+    const availableRewardTokenPrices = tokenPriceMantissaTuple[1];
+    const referenceRewardTokenPrice =
+      availableRewardTokenPrices.find(tokenPrice => tokenPrice.priceSource === 'oracle') ||
+      availableRewardTokenPrices.find(tokenPrice => tokenPrice.priceSource === 'coingecko');
```
therealemjy: Since we are fetching token prices in relation to rewards, perhaps we should prioritize getting prices from Merkl rather than the oracle? I'd also add a fallback mechanism:


```suggestion
    const referenceRewardTokenPrice =
      availableRewardTokenPrices.find(tokenPrice => tokenPrice.priceSource === 'merkl') ||
      availableRewardTokenPrices.find(tokenPrice => tokenPrice.priceSource === 'coingecko') ||
      availableRewardTokenPrices[0];
```

apps/evm/src/clients/api/queries/useGetPools/getPools/index.ts
```diff
@@ -22,7 +22,7 @@ export const getPools = async ({
   tokens,
 }: GetPoolsInput) => {
   const [
-    { pools: apiPools },
+    { pools: apiPools, tokenMetadata: apiTokenMetadata },
```
therealemjy: Sorry if you talked about this before and I missed it, but why don't we keep using the `tokenPriceCents` property of the `ApiMarket` type? The frontend only uses the underlying token price coming from the oracle (as highlighted by the logic you add to the `getPools/formatOutput` function in this PR).

If the reason is that we want to use a different price source for distributions, we can still do that by reusing the `priceMantissa` field of the `ApiReward` type (here as well the backend would make the connection so that we don't need any extra logic on the frontend).

apps/evm/src/clients/api/queries/getPendingRewards/__tests__/index.spec.ts
```diff
@@ -92,20 +93,18 @@ const fakePublicClient = {
...
-);
-
+// const fakeTokenPriceMapping = tokens.reduce(
+//   (acc, token) => ({
+//     ...acc,
+//     [token.address]: [],
+//   }),
+//   {},
+// );
+//
```
therealemjy: This can be removed then

therealemjy (changes requested):

apps/evm/src/clients/api/queries/useGetPools/getPools/index.ts
gleiser-oliveira: The idea is to eventually remove these fields. I kept them for backwards compatibility in case we need to rollback changes, but basically once everything is settled we decouple any price related info from the `Market` and `RewardDistributor` models.

apps/evm/src/clients/api/queries/useGetPools/getPools/formatOutput/formatDistributions/index.ts
```diff
@@ -67,6 +69,20 @@ export const formatDistributions = ({
         return;
       }
 
+      const correspondingRewardTokenPrice = apiTokenMetadata
+        .find(tm => areAddressesEqual(rewardTokenAddress, tm.address))
+        ?.tokenPrices.find(tokenPrice =>
+          rewardType === 'venus'
+            ? tokenPrice.priceSource === 'oracle'
+            : tokenPrice.priceSource === 'merkl',
+        );
```
therealemjy: Rather than having to go through the array of prices for every asset, you can pass a mapping of the prices by token address. You can then read from this mapping using the token address.
We currently do this for the prices of reward tokens: https://github.com/VenusProtocol/venus-protocol-interface/blob/4a193d1dfa19f32e2866c7e64e4ae5f4218b9a7d/apps/evm/src/clients/api/queries/getPendingRewards/index.ts#L242-L243

Also, I would use any price available if the source we want isn't available, rather than filtering out the reward if we don't have a price for the exact source we want.

apps/evm/src/clients/api/queries/useGetPools/getPools/index.ts
therealemjy: Got it, I'll review the PR given this then.

therealemjy (changes requested):

apps/evm/src/clients/api/queries/useGetPools/getPools/formatOutput/formatDistributions/index.ts
gleiser-oliveira: Refactored to pass down the price mapping, thanks!

apps/evm/src/clients/api/queries/getPendingRewards/index.ts
gleiser-oliveira: I think all 3 options are valid here, so instead of adding a fallback, I added a 3rd `find`. The intention is to have order of preference for the price source.

therealemjy: That's fine because the array is small, but for future reference I'd normally suggest sorting the array by order of preference then returning the first item.

apps/evm/src/clients/api/queries/useGetPools/getPools/formatOutput/formatDistributions/index.ts
therealemjy: Looks good, thanks. There's still one part of my comment to address: I would use any price available if the source we want isn't available, rather than filtering out the reward if we don't have a price for the exact source we want (otherwise users won't see their rewards and won't be able to claim them).
Something like:

```ts
const tokenPrices = tokenPricesMapping[
  market.underlyingAddress.toLowerCase()
];

const correspondingOraclePrice = tokenPrices.find(
  p =>
    p.priceOracleAddress &&
    areAddressesEqual(apiPool.priceOracleAddress, p.priceOracleAddress),
) || tokenPrices[0];
```

apps/evm/src/clients/api/queries/getPendingRewards/getApiTokenPrice/index.ts
```diff
@@ -24,16 +34,19 @@ export const getApiTokenPrice = async ({
...
+  const result = response.data?.result || [];
 
-  return Object.entries(result).reduce(
-    (acc, tokenPriceTuple) => ({ ...acc, [tokenPriceTuple[0]]: new BigNumber(tokenPriceTuple[1]) }),
-    {},
-  );
+  return result.reduce((acc, tokenMetadata) => {
+    return {
+      ...acc,
+      [tokenMetadata.address]: tokenMetadata.tokenPrices,
```
therealemjy: Just to be safe:

```suggestion
      [tokenMetadata.address.toLowerCase()]: tokenMetadata.tokenPrices,
```

therealemjy (changes requested):

apps/evm/src/clients/api/queries/useGetPools/getPools/formatOutput/formatDistributions/index.ts
gleiser-oliveira: Would that really be preferrable? I mean, a reward is, currently, always internal (oracle price) or external (merkl price), so it has to have one of them. Now, if for any reason this array comes empty, if I look for the first element here I cause a crash instead of filtering the bad record out 🤔

therealemjy: I see where your mind is. You're right that if the array is empty then we shouldn't make the app crash.

I think both our arguments are valid: so I'd sort the array by order of preference and grab the first item. If the first item does not exist, then I'd skip the record.

therealemjy (approved):

PR #4515, opened 2025-07-10
refactor: stop logging warning headers on Sentry

gleiser-oliveira:
- We don't need to log warning headers on Sentry, they are expected when the backend is in the process of deprecating a certain endpoint version

### DIFF
--- a/apps/evm/src/utilities/restService.ts
+++ b/apps/evm/src/utilities/restService.ts
@@ -88,13 +88,8 @@ export async function restService<D>({
 
       try {
         data = await response.json();
-        const warning = response.headers.get('Warning');
-
-        if (warning) {
-          logError(warning);
-        }
-      } catch {
-        // Do nothing
+      } catch (err) {
+        logError(err);
       }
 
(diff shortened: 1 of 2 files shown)


apps/evm/src/utilities/restService.ts
```diff
@@ -88,13 +88,8 @@ export async function restService<D>({
...
         data = await response.json();
-        const warning = response.headers.get('Warning');
-
-        if (warning) {
-          logError(warning);
-        }
-      } catch {
-        // Do nothing
+      } catch (err) {
+        logError(err);
```
gleiser-oliveira, line 92: @therealemjy do you recall why we were ignoring errors here? I think at some point there was an error being spammed, but maybe we can have it being logged by the catch block again

therealemjy: I believe we stopped logging those because they ended up spamming Sentry. We can enable it back and see if that's still the case.

therealemjy (approved):

PR #4535, opened 2025-07-16
feat: mobile web UI improvements

gleiser-oliveira:
## Jira ticket(s)

VEN-XXX

- Series of improvements for the mobile web UI. We can refer to Figma to check in detail the expected outcome.


apps/evm/src/components/Table/TableCards.tsx
```diff
@@ -122,7 +109,11 @@ export function TableCards<R>({
           return (
             <Card
               key={rowKey}
-              css={styles.tableWrapperMobile({ clickable: !!(rowOnClick || getRowHref) })}
+              className={cn(
+                '[&:not(:last-of-type)]:mb-4',
```
therealemjy: Tailwind has a built-in class for this: `space-y-4`, which you can add to the parent to apply a gap of 4 between each child.

apps/evm/src/pages/Market/Page/AssetWarning/index.tsx
```diff
@@ -63,20 +65,24 @@ export const AssetWarning: React.FC<AssetWarningProps> = ({
...
+        noHorizontalPadding={isSmOrUp}
         title={t('assetWarning.modalTitle', {
           poolName: pool.name,
         })}
       >
         <MarketTable
           data-testid={TEST_IDS.marketTable}
           rowOnClick={handleHideAssets}
+          controls={false}
+          selectVariant={isSmOrUp ? 'tertiary' : 'quaternary'}
```
therealemjy: It seems the select is only shown in the XS resolution, so I don't think you need to the ternary operator.
```suggestion
          selectVariant="quaternary"
```

therealemjy (approved):

apps/evm/src/components/Table/TableCards.tsx
```diff
@@ -88,7 +75,7 @@ export function TableCards<R>({
           value={selectedOption?.value || selectOptions[0].value}
           onChange={handleOrderChange}
           css={styles.cardsSelect}
-          variant="tertiary"
+          variant={selectVariant || 'tertiary'}
```
therealemjy: nit: you can use a default value for the `selectVariant` prop instead

apps/evm/src/components/charts/TooltipContent/index.tsx
```diff
@@ -18,7 +18,10 @@ const TooltipContent: React.FC<TooltipContentProps> = ({ items }) => {
   return (
     <div css={styles.container}>
       {items.map(item => (
-        <div css={styles.item} key={`tooltip-content-item-${item.label}`}>
+        <div
+          className="[&:not(:last-of-type)]:mb-1 [&:not(:last-of-type)]:sm:mb-2 flex items-center mr-auto"
```
therealemjy: Same comment as previously: I'd recommend using the `space-y-X` class on the parent tag instead

apps/evm/src/containers/Layout/Header/MarketInfo/index.tsx
```diff
@@ -90,30 +92,22 @@ export const MarketInfo = () => {
...
+          handleGoBack={handleGoBack}
+        />
+      ) : (
+        <XsMarketInfoHeader
+          asset={asset}
+          pool={pool}
+          isUserConnected={isUserConnected}
+          handleGoBack={handleGoBack}
+        />
+      )}
```
therealemjy: I'd use Tailwind's responsive classes (e.g.: `hidden sm:block`) to handle showing/hiding each UI (better performance over using JS).

Example: https://github.com/VenusProtocol/venus-protocol-interface/blob/7c5c2b4043f4da96806e8e8ce9fab4001205e907/apps/evm/src/containers/Layout/Header/TopBar/index.tsx

apps/evm/src/containers/Layout/index.tsx
```diff
@@ -19,6 +24,14 @@ export const Layout: React.FC = () => {
     const shouldShowScrollToTopVisible =
       scrollElem?.scrollTop && scrollElem.scrollTop > viewportHeight;
 
+    const scrollToTopRect = scrollToTopRef.current?.getBoundingClientRect();
+    const footerRect = footerRef.current?.getBoundingClientRect();
+    const footerTopPos = footerRect ? footerRect.top - footerRect.height : undefined;
+    const isCloseToBottom =
+      scrollToTopRect && footerTopPos ? scrollToTopRect.bottom >= footerTopPos : false;
+
+    setIsCloseToBottom(isCloseToBottom);
```
therealemjy: This will trigger a re-render for every pixel scrolled. You can limit the amount of updates:

```suggestion
if (isCloseToBottom !== store.use.isCloseToBottom()) {
    setIsCloseToBottom(isCloseToBottom);
}
```

apps/evm/src/containers/MarketTable/styles.ts
```diff
@@ -19,7 +19,11 @@ export const useStyles = () => {
     cardContentGrid: css`
       .table__table-cards__card-content {
         grid-template-columns: 1fr 1fr 1fr;
-        row-gap: ${theme.spacing(5)};
+        row-gap: ${theme.spacing(4)};
+
+        ${theme.breakpoints.down('md')} {
+          row-gap: ${theme.spacing(5)};
+        }
```
therealemjy: If you don't mind, I think we should follow the same pattern as Tailwind which is to write CSS with a mobile-first vision (mobile is the default style, then any responsive styling for bigger screen sizes is applied from smallest to largest).

Just to clarify, this would translate to this:

```suggestion
        row-gap: ${theme.spacing(5)};

        ${theme.breakpoints.up('md')} {
          row-gap: ${theme.spacing(4)};
        }
```

therealemjy (approved):

PR #4536, opened 2025-07-16
feat: increase Port limit to $100k + UI improvements

therealemjy:
- add `importAavePositions` feature flag
- lowercase ABI variable names in `contracts` lib config 
- increase Port feature limit to $100 000
- improve error handling. We now log every error passed to the `handleError` function, ignoring errors due to willingly canceled transactions
- add "Position import canceled" analytic event that is sent when a user has requested to import a position but refused to sign the transaction
- small UI improvements on the Port feature following design changes


apps/evm/src/hooks/useIsFeatureEnabled/index.tsx
```diff
@@ -120,16 +121,31 @@ export const featureFlags = {
...
+    ChainId.SEPOLIA,
+    ChainId.OPTIMISM_MAINNET,
+    ChainId.OPTIMISM_SEPOLIA,
+    ChainId.ZKSYNC_MAINNET,
+    ChainId.ZKSYNC_SEPOLIA,
+    ChainId.BASE_MAINNET,
+    ChainId.BASE_SEPOLIA,
+    ChainId.UNICHAIN_MAINNET,
+    ChainId.UNICHAIN_SEPOLIA,
+  ],
```
gleiser-oliveira, line 131: What's the idea here? I see opBNB is left out

therealemjy: So the idea is to list the chains we can import certain protocol positions from.
We end up with two lists:
- a list of chains Biconomy is supported on
- a list of chains various external protocols are deployed on

We dissociate both because Biconomy might be available on certain chains certain external protocols aren't deployed on, and vice versa.

gleiser-oliveira (approved):

PR #4556, opened 2025-07-21
refactor: remove unused code in the supply form

gleiser-oliveira:
- I believe the Uniswap notice is not needed anymore, the wrap and unwrap feature has been stable for a while
- Same with the supply preset percentages


therealemjy (approved):
Good call.
You'll just need to run `yarn extract-translation-keys` to get rid of the translations linked to the keys removed in this PR.

PR #4560, opened 2025-07-22
feat: sync responsive breakpoints with designs

therealemjy:
- sync responsive breakpoints with designs

### DIFF
--- a/.changeset/huge-candles-lead.md
+++ b/.changeset/huge-candles-lead.md
@@ -1,0 +1,6 @@
+---
+"@venusprotocol/ui": minor
+"@venusprotocol/evm": minor
+---
+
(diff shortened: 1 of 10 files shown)


Frayst2:
Hey,
Not sure if this commit is done but the current result does not look good with an iPhone 16. The resolution was way better before. 

IMG_0315

IMG_0316

IPhone horizontal : the menu is vertical 🫠
IMG_0318

I could send pics of previous resolution if needed

therealemjy:
Hi @Frayst2 
Thanks for highlighting those issues. We talked with our design team and while the breakpoints used in the dApp will change in the future, we've decided to put this on hold for now. Once we do get to modifying them, we'll make sure mobile views display correctly.

PR #4577, opened 2025-07-24
feat: account performance graph

therealemjy:
## Jira ticket(s)

VEN-3288

- add `AreaChart` component
- remove `charts` component directory
- rename `TooltipContent` component to `ChartTooltipContent`
- add performance chart to new Account page
- update `InterestRateChart` of Market page to remove split between container and UI
- update `ApyChart` of Market page to use `AreaChart` component


apps/evm/src/pages/Account/NewPage/PerformanceChart/fakeData.ts
```diff
@@ -0,0 +1,602 @@
+export const data = [
+  {
+    apyPercentage: 463.22072070832434,
```
gleiser-oliveira: I was thinking about this field, what is the meaning of this value? Is it the the sum of the APYs accross all the participating markets at this point in time minus the borrow APYs?

apps/evm/src/pages/Account/NewPage/PerformanceChart/fakeData.ts
```diff
@@ -0,0 +1,602 @@
...
+    timestampMs: 1750752158000,
```
gleiser-oliveira, line 3: Without formatting, the stored timestamps from the subgraph are in microseconds. Meaning they would have to be either converted to miliseconds, or change the expectation here to microseconds. This can be done by the backend before sending out the response, or it can be sent as is to the frontend.

therealemjy: If we can make the backend return milliseconds, that would be great (that's even less computation to do on the frontend).
That said, the `/market/history` endpoint currently returns data points in microseconds and so we convert them to milliseconds on the frontend. I think the backend should use milliseconds here too, but let me know what you think.

apps/evm/src/pages/Account/NewPage/PerformanceChart/fakeData.ts
therealemjy: It doesn't correspond to anything: I just copied some data from the market history endpoint to build the UI and forgot to remove this property. I'll remove it.

apps/evm/src/pages/Account/NewPage/PerformanceChart/fakeData.ts
gleiser-oliveira: Let's start with this one in miliseconds then, it's basically just removing the least significant digits. And keep here as it already is.

gleiser-oliveira (approved):

apps/evm/src/pages/Account/NewPage/PerformanceChart/fakeData.ts
therealemjy: Sounds good, thanks

PR #4589, opened 2025-07-28
feat: read market's bad debt from the API

gleiser-oliveira:
- Read market's bad debt from the API
- Use the value when calculating the interest and utilization rates


apps/evm/src/clients/api/queries/useGetPool/__snapshots__/index.spec.tsx.snap
```diff
@@ -5,6 +5,7 @@ exports[`useGetPool > returns the correct asset 1`] = `
   "pool": {
     "assets": [
       {
+        "badDebtMantissa": 0n,
```
therealemjy: Just to cover this new feature with tests, it would be good to have some test assets with a non-zero bad debt.

apps/evm/src/clients/api/queries/useGetPools/getPools/getApiPools/index.ts
```diff
@@ -80,6 +80,7 @@ export interface ApiMarket {
   lastCalculatedXvsAccruedBlockNumber: string | null;
   supplyCapsMantissa: string;
   borrowCapsMantissa: string;
+  badDebt: string;
```
therealemjy: I understand this is the schema from the API, but I'd recommend updating it at some point to be consistent with the naming convention: `badDebtMantissa`.

gleiser-oliveira: Agreed, let's do it before going live

therealemjy (approved):

PR #4597, opened 2025-07-30
feat: update core analytic events

therealemjy:
## Jira ticket(s)

VEN-3348

- add `isUserRejectedTxError` utility function to detect if an RPC error is due to the user rejecting it
- refactor analytics lib to export only one function or component per file
- refactor `useAnalytics` hook to map events using a flat type structure
- log wallet address, provider, origin and page with every analytic event
- log analytic events in the console in test environments
- update how core analytic events (supply/withdraw/borrow/repay) are logged, according to the new specs (see Notion doc)
- mock analytics lib in all tests


alicecha (approved):
All good 👍

PR #4629, opened 2025-08-05
feat: send analytic events related to wallet actions

therealemjy:
## Jira ticket(s)

VEN-3348

- send analytic events related to wallet actions

### DIFF
--- a/.changeset/puny-times-like.md
+++ b/.changeset/puny-times-like.md
@@ -1,0 +1,5 @@
+---
+"@venusprotocol/evm": minor
+---
+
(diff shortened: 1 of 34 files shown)


alicecha:
Can we add "wallet_provider_selected" when the user selects the wallet in the wallet pop-up, along with parameter "selection" or "option" to differentiate between the wallet provider that the user chose to connect: rabby, metamask, coinbase, other, family. Thanks!

Rest looks great!

alicecha (approved):
Not possible to track wallet provider chosen pre-connection so happy with current implementation.

PR #4654, opened 2025-08-11
feat: update page architecture

therealemjy:
## Jira ticket(s)

VEN-3354

- add `useGetHomePagePath` hook
- remove Dashboard page
- add Isolated pools page
- add Core pool page and make it the new home page
- update `useGetPools` hook so that data is kept between refetches, if the `chainId` parameter did not change and `accountAddress` was previously `undefined`. This covers a very specific case: a user visits the dApp, which first loads the pools without their account in mind, then the account address is retrieved a few seconds later therefore triggering a refetch of the pools, this time with the account data added. With the new logic, users will always see the pools from the first fetch instead of a blinking loading state
- update `CellGroup` component to follow new design requirements
- update `Breadcrumbs` component to support new pages
- update menu items to reflect new page architecture


Frayst2:
IMG_0440

With the new page structure, the hyperlink of isolated pool may be removed, wdyt? when you click on it, the structure path is not displayed. @therealemjy

therealemjy:
Hi @Frayst2 
Thanks for the heads up, this is actually a mistake on my end. Indeed the pool of each asset should not be displayed here, since we're already on the page on that pool. I'll get this fixed, thank you.

gleiser-oliveira (commented):
@therealemjy I noticed that refreshing on the `Account` and `Port` pages will take you back to the Core Pool page, but shouldn't the behavior be staying on them?

therealemjy:
@gleiser-oliveira It should, but unfortunately when reloading the page there's a delay to retrieve the connected wallet. In certain cases, this delay is long enough to trigger the redirection you mentioned. I've made modifications to improve this behavior, but we can't remove this behavior completely as we don't control the delay.

apps/evm/src/pages/Account/OldPage/AccountBreakdown/index.tsx
```diff
@@ -83,6 +83,7 @@ const Account: React.FC = () => {
   if (!hasPositions) {
     return <AccountPlaceholder />;
   }
+  console.log('CALLED');
```
gleiser-oliveira: Stray log

therealemjy: Good catch, thanks

gleiser-oliveira (approved):

PR #4745, opened 2025-09-08
feat: add account historical transactions

gleiser-oliveira:
## Jira ticket(s)

VPD-58

- Added tab, components and logic to fetch and display the account's historical transactions


apps/evm/src/pages/Account/NewPage/Transactions/TransactionsList/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+    <>
+      <ul
+        className={cn(
+          'flex flex-col w-full items-center justify-evenly space-y-6 md:bg-cards md:p-4 md:rounded-xl',
+          className,
+        )}
+      >
+        {Object.keys(transactionsGroupedByDate).map(day => (
+          <div className="flex flex-col w-full space-y-3">
+            <span className="text-base font-semibold">{day}</span>
```
therealemjy: Semantically speaking, you can't put block-level tag inside a `ul` tag.
A possible solution could be to update the `TransactionRow` component so that it can optionally render the date you want to render

apps/evm/src/pages/Account/NewPage/Transactions/TransactionsList/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+  return (
+    <>
+      <ul
+        className={cn(
+          'flex flex-col w-full items-center justify-evenly space-y-6 md:bg-cards md:p-4 md:rounded-xl',
+          className,
+        )}
+      >
+        {Object.keys(transactionsGroupedByDate).map(day => (
+          <div className="flex flex-col w-full space-y-3">
```
therealemjy: According to the designs, there's no spacing between each list item on md and up. Instead, they each have a fixed height (64px) assigned to them (I'll leave another comment to update the `TransactionRow` component).

```suggestion
      <ul
        className={cn(
          'flex flex-col w-full items-center justify-evenly space-y-6 md:bg-cards md:py-4 md:rounded-xl md:space-y-0',
          className,
        )}
      >
        {Object.keys(transactionsGroupedByDate).map(day => (
          <div className="flex flex-col w-full space-y-3">
```

apps/evm/src/pages/Account/NewPage/Transactions/TransactionRow/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }, [t, transaction]);
+
+  return (
+    <li
+      className={cn(
+        'flex flex-col px-4 py-3 md:p-0 bg-cards rounded-xl md:rounded-none w-full md:flex-row items-center justify-evenly',
+        className,
+      )}
+    >
```
therealemjy: On md and up, this component has fixed height of 64px and a hover effect assigned to it:
```suggestion
    <li
      className={cn(
        'flex flex-col px-4 py-3 bg-cards rounded-xl  w-full items-center justify-evenly md:flex-row md:p-0 md:h-16 md:hover:bg-lightGrey md:rounded-none',
        className,
      )}
    >
```

apps/evm/src/pages/Account/NewPage/Transactions/TransactionRow/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+              hash: transaction.txHash,
+              urlType: 'tx',
+              chainId: transaction.chainId,
+            })}
+            target="_blank"
+            rel="noreferrer"
+          >
+            <span className="hidden lg:block text-sm underline">
+              {t('account.transactions.view')}
+            </span>
```
therealemjy: When the component is hovered on md and up, this text should be white. You can use the group feature from Tailwind to detect that the root tag is hovered and apply styling to this element in consequence: https://v3.tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state

apps/evm/src/pages/Account/NewPage/Transactions/TransactionRow/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+          <div className="p-[6px] md:p-2 mr-2 md:mr-3 flex items-center justify-center bg-lightGrey rounded-full">
+            <Icon className="h-3 w-3 md:h-4 md:w-4" name={getTransactionIcon(transaction.txType)} />
+          </div>
+          <div className="flex flex-row w-full md:w-auto items-center md:items-start justify-between md:flex-col">
+            <span className="text-sm">{transactionTitle}</span>
+            <span className="text-grey text-xs">{format(transaction.txTimestamp, 'hh:mm aa')}</span>
+          </div>
+        </div>
+
+        <hr className="block w-full text-lightGrey my-3 md:hidden" />
```
therealemjy: This line can be replaced with the `Delimiter` component.

therealemjy (commented):
According to the designs, the second select is meant to enable filtering by token, not by specific market. I looked for conversations you might have had with Pavel about this but couldn't find anything.
Designs:
Image
PR:
Screenshot 2025-09-08 at 08 22 40

Also, in this PR we show the vToken name instead of the transaction value in dollars of each item transaction:
Designs:
Screenshot 2025-09-08 at 08 48 51
PR: 
Screenshot 2025-09-08 at 08 50 05

apps/evm/src/clients/api/queries/getAccountHistoricalTransactions/types.ts
```diff
@@ -0,0 +1,52 @@
...
+  results: ApiAccountHistoricalTransaction[];
+}
+
+export interface GetAccountHistoricalTransactionsInput {
+  accountAddress: string;
+  contractAddress: string;
+  chainId: ChainId;
+  getPoolsData: GetPoolsOutput | undefined;
+  type?: number | string;
+  page?: number | string;
```
therealemjy: To avoid having to do what you do [here](https://github.com/VenusProtocol/venus-protocol-interface/pull/4745/files#diff-50cd1d8e563cb7ebdb8b26c414f13ac06dd32f58dc425483375f2dda156c3c71R26), I'd use only one type for each prop:

```suggestion
  type?: number;
  page?: number;
```

apps/evm/src/clients/api/queries/getAccountHistoricalTransactions/index.ts
therealemjy, line 1: Just to be consistent with the name of the `getAccountPerformanceHistory` query function, which also fetches historical data, I'd rename this file and the function it contains `getAccountTransactionHistory`.

apps/evm/src/clients/api/queries/getAccountHistoricalTransactions/index.ts
```diff
@@ -0,0 +1,65 @@
...
+    ) || [];
+
+  const formattedResponse = txsResponse.data.results.map(tx => ({
+    transaction: {
+      ...tx,
+      txType: tx.txType as TxType,
+      txTimestamp: new Date(tx.txTimestamp),
+    },
+    asset: allAssets.find(a => areAddressesEqual(a.vToken.address, tx.contractAddress))!,
+  }));
```
therealemjy: You don't need the entire asset, only the underlying token.

We also need to think that more transaction types will be added in the near future, so I'd suggest creating a formatter function that returns, for each transaction type, exactly the information needed to display them on the UI.
To put it in code, I'd imagine something like this:

```typescript
export interface TransactionBase {
  txType: TxType;
  id: string;
  hash: string;
  blockTimestamp: Date;
  blockNumber: string;
  accountAddress: string;
  contractAddress: string;
  chainId: ChainId;
}

export interface AmountTransaction extends TransactionBase {
  txType: TxType.Mint | TxType.Borrow | TxType.Redeem | TxType.Repay | TxType.Approve;
  poolName: string;
  amountTokens: BigNumber;
  amountCents: number;
  token: Token;
}

export interface CollateralTransaction extends TransactionBase {
  txType: TxType.EnterMarket | TxType.ExitMarket;
  poolName: string;
  token: Token;
}

export type Transaction = AmountTransaction | CollateralTransaction;

const formatTransaction = ({
  tokenPriceMap,
  vTokenMap,
  apiTransaction,
}: {
  tokenPriceMap: Record<Address, number>;
  vTokenMap: Record<
    Address,
    {
      underlyingToken: Token;
      poolName: string;
    }
  >;
  apiTransaction: ApiAccountHistoricalTransaction;
}) => {
  const tokenPriceCents = Object.prototype.hasOwnProperty.call(
    tokenPriceMap,
    apiTransaction.underlyingAddress,
  )
    ? tokenPriceMap[apiTransaction.underlyingAddress as keyof typeof tokenPriceMap]
    : undefined;

  const vTokenData = Object.prototype.hasOwnProperty.call(vTokenMap, apiTransaction.contractAddress)
    ? vTokenMap[apiTransaction.underlyingAddress as keyof typeof vTokenMap]
    : undefined;

  const sharedProps = {
    txType: apiTransaction.txType as TxType,
    id: apiTransaction.id,
    hash: apiTransaction.txHash,
    blockTimestamp: new Date(apiTransaction.txTimestamp),
    blockNumber: apiTransaction.blockNumber,
    accountAddress: apiTransaction.accountAddress,
    contractAddress: apiTransaction.contractAddress,
    chainId: apiTransaction.chainId,
  };

  switch (apiTransaction.txType) {
    case TxType.Mint:
    case TxType.Borrow:
    case TxType.Redeem:
    case TxType.Repay:
    case TxType.Approve: {
      if (!vTokenData || !tokenPriceCents) {
        return undefined;
      }

      const amountTokens = convertMantissaToTokens({
        value: apiTransaction.amountUnderlyingMantissa,
        token: vTokenData.underlyingToken,
      });

      const amountCents = amountTokens.multipliedBy(tokenPriceCents);

      const tx: AmountTransaction = {
        ...sharedProps,
        amountTokens,
        amountCents,
        poolName: vTokenData.poolName,
        token: vTokenData.underlyingToken,
      };

      return tx;
    }
    // TODO: handle other transaction types
    default:
      undefined;
  }
};
```

Note that `formatTransaction` can return `undefined` if some data is missing, which can happen if the API was to return a new type of transaction it can't handle yet. This makes it easy to filter out unsupported transactions.

gleiser-oliveira:
> According to the designs, the second select is meant to enable filtering by token, not by specific market. I looked for conversations you might have had with Pavel about this but couldn't find anything. Designs: Image PR: Screenshot 2025-09-08 at 08 22 40
> 
> Also, in this PR we show the vToken name instead of the transaction value in dollars of each item transaction: Designs: Screenshot 2025-09-08 at 08 48 51 PR: Screenshot 2025-09-08 at 08 50 05

I think filtering down by market (or by single origin contract let's say) makes much more sense, because underlying tokens are shared in multiple markets. If one selects USDT, then there's a situation where it returns all USDT markets TXs and no way to filter to a single one. Same when other events get added, if XVS is selected then it is the XVS market along with vault stakes, votes, etc.

therealemjy:
Is it not the purpose of this feature? I can imagine a user who holds USDT in their wallet and supplied some to multiple pools would want to see their history for this token specifically, across all pools/vaults.

gleiser-oliveira:
> Is it not the purpose of this feature? I can imagine a user who holds USDT in their wallet and supplied some to multiple pools would want to see their history for this token specifically, across all pools/vaults.

In my opinion that would be very annoying if you wanted to check, let's say, the last time you supplied to a market but you had to go through more recent txs from other sources that share the same underlying token until you find the one you were looking for, instead of just filtering the exact market (or vault, if we consider XVS as another example of this case).

therealemjy:
I get your point. If Pavel is ok with this UX then that works for me 👍

apps/evm/src/pages/Account/NewPage/Transactions/TransactionsList/index.tsx
gleiser-oliveira: I changed this div to a `li`, the way I see it is a list of grouping by date

apps/evm/src/clients/api/queries/getAccountHistoricalTransactions/types.ts
gleiser-oliveira: Updated to numbers only. But I believe I still have to sanitize the output of `useSearchParams` because they come as any strings

apps/evm/src/pages/Account/NewPage/Transactions/TransactionRow/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+      <a
+        className="flex flex-col md:flex-row w-full"
+        href={generateExplorerUrl({
+          hash,
+          urlType: 'tx',
+          chainId,
+        })}
+        target="_blank"
+        rel="noreferrer"
+      >
```
therealemjy: Nit: you could remove the `div` and apply it's styling to the `a` so that the whole component is clickable

apps/evm/src/pages/Account/NewPage/Transactions/TransactionRow/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+          </div>
+        </div>
+
+        <Delimiter className="my-3 md:hidden" />
+
+        <div className="flex w-full items-center">
+          <TokenIcon className="mr-2 self-start" token={token} />
+          <TransactionDetails className="flex-1" transactionData={transactionData} />
+          <a
+            className="hidden md:flex items-center self-start text-grey md:group-hover:text-offWhite space-x-1 mt-1"
```
therealemjy: This can be a `div` since the whole component is a link already.
Also, according to the designs this element should be vertically aligned in the center.

```suggestion
          <a
            className="hidden md:flex items-center text-grey md:group-hover:text-offWhite space-x-1"
            href={generateExplorerUrl({
              hash,
              urlType: 'tx',
              chainId,
            })}
            target="_blank"
            rel="noreferrer"
          >
```

apps/evm/src/clients/api/queries/getAccountTransactionHistory/formatApiTransaction.ts
```diff
@@ -0,0 +1,81 @@
...
+  const { poolName, vToken } = contractToTokenMap[contractAddress];
+
+  const vTokenSymbol = vToken.symbol;
+  const token =
+    txType === TxType.Approve
+      ? {
+          ...vToken,
+          asset: vToken.asset ?? vToken.underlyingToken.asset,
+        }
+      : vToken.underlyingToken;
```
therealemjy: I know that technically users are allowing a vToken contract to spend their tokens, but I'd use the underlying token asset for all transactions as that's the one users are familiar with.

gleiser-oliveira: Wouldn't it be misleading? At the very least I need to format the amount with the vToken decimals

Screenshot 2025-09-11 at 9 32 16 a m

chechu: Regular users won't approve vTokens, so they won't see many Approval events in their timeline. I would use the symbol of the approved token (vUSDT, for example). I know it's ugly, but as I said regular users won't see that, and expert users (who approved VTokens) could expect that unit

Finally, the conversion vToken/underlying should consider the exchange when the event was emitted. I think that could overcomplicate the development

therealemjy: Anyone supplying ERC20 tokens to the protocol will need to approve their spending, so I'd say that's a type of transaction most users will get to see in their history.

I get your point regarding the conversion, but in my opinion since users are approving a quantity of vTokens, then the corresponding amount of underlying tokens evolves with time (if I approve 10 vFLOKI a year ago when it corresponded to 20 FLOKI, then today it might represent 40 FLOKI). We use the most recent price reference anyway when calculating the corresponding USD amount of each transaction, so I'd say this follows the same logic.

therealemjy (approved):

therealemjy:
PR looks good, but I approved right before finding a bug with the filtering system.

On an account with several supply transactions, if I filter by "Supplies" transactions I see none of them.

Without filtering:
Screenshot 2025-09-12 at 08 30 01

With filtering:
Screenshot 2025-09-12 at 08 30 05

I'm seeing a similar behavior with other transaction types, where some or all transactions of that type are being filtered out.

apps/evm/src/clients/api/queries/getAccountTransactionHistory/formatApiTransaction.ts
gleiser-oliveira: Just to clarify, we are tracking the approval event on the vToken contract, not the approval on the underlying token contract, so this is not the most frequent one in the timeline of txs.

Also, the price being shown is the price in the block number of the tx, this info is sent by the backend.

therealemjy: Ahhh sorry, that's my bad. That makes sense now then, sorry for the confusion.

PR #4747, opened 2025-09-08
feat: improve Tabs to handle search param nav

therealemjy:
- add `initialActiveTabId` and `navType` props to `useTabs` hook
- update `Tabs` component to use `useTabs` hook
- E-mode: make sure user is redirected to Repay tab of the Market page when clicking on the CTA of a blocking borrow position

### DIFF
--- a/apps/evm/src/containers/Vault/WithdrawFromVestingVaultModal/index.tsx
+++ b/apps/evm/src/containers/Vault/WithdrawFromVestingVaultModal/index.tsx
@@ -2,6 +2,7 @@
 import { useState } from 'react';
 
 import { Modal, type ModalProps, Tabs, TextButton } from 'components';
+import type { Tab } from 'hooks/useTabs';
 import { useTranslation } from 'libs/translations';
 import type { Token } from 'types';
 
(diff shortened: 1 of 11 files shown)


apps/evm/src/containers/Vault/WithdrawFromVestingVaultModal/index.tsx
```diff
@@ -32,11 +35,36 @@ const WithdrawFromVestingVaultModal: React.FC<WithdrawFromVestingVaultModalProps
     setShouldDisplayWithdrawalRequestList(true);
     // Set initial active tab index to 1 so that if user clicks on modal back button
     // they get redirect to the "Request withdrawal" tab
-    setInitialActiveTabIndex(1);
+    setInitialActiveTabId(tabs[1].id);
   };
 
-  const { t } = useTranslation();
-  const styles = useStyles();
+  const tabs: Tab[] = [
```
gleiser-oliveira, line 41: nit: it's a minor thing, but when I read line 38 it wasn't so clear that the `tabs` variable was declared after it

therealemjy: I get your point, unfortunately I can't declare the `tabs` variable before `handleDisplayWithdrawalRequestList` because the latter uses the former :/

gleiser-oliveira (approved):

PR #4781, opened 2025-09-15
feat: add legal pages

therealemjy:
## Jira ticket(s)

VPD-125

### [landing app](https://github.com/VenusProtocol/venus-protocol-interface/tree/main/apps/landing/)
- update file architecture to support multiple pages
- add `date-fns`
- add `/privacy-policy` and `/terms-of-use` routes
- update footer
- rewrite `Container` component with Tailwind
- add compiler options to TS config to allow absolute imports


apps/landing/src/pages/PrivacyPolicy/index.tsx
```diff
@@ -0,0 +1,438 @@
+import { Legal } from 'components/Legal';
+import { LANDING_PAGE_URL } from 'constants/production';
+
+const LAST_UPDATED_AT = 'September 20, 2025';
```
gleiser-oliveira: Is this date correct? Also in the terms of use

therealemjy: It depends on when we release. Once the PR approved, I can update it to the day's date then merge and deploy so it corresponds to the first release date.

gleiser-oliveira (approved):

PR #4807, opened 2025-09-19
feat: wire up E-mode feature

therealemjy:
## Jira ticket(s)

VPD-65

- add analytic events to cover E-mode feature
- add `useSetEModeGroup` mutation function
- remove `liquidityCents` and `liquidityTokens` properties from `EModeAssetSettings` type
- fetch `isBorrowable` asset property from API
- fetch E-mode groups from API
- wire up E-mode UI


apps/evm/src/pages/Market/MarketHistory/Card/index.tsx
```diff
@@ -65,7 +65,7 @@ export const Card: React.FC<CardProps> = ({
   );
 
   const liquidationIncentivePercentage =
-    getPoolLiquidationIncentiveData?.liquidationIncentivePercentage;
+    getPoolLiquidationPenaltyData?.liquidationIncentivePercentage;
```
gleiser-oliveira: It's more of a nitpick, but I feel the naming was better before.
```
const liquidationIncentivePercentage = getPoolLiquidationIncentiveData?.liquidationIncentivePercentage;
```
Because it would all match everywhere.

therealemjy: I agree, let me change that

gleiser-oliveira: Nice, do you want to also rename the hook and function?

therealemjy: The current name should be correct. While we fetch the liquidation incentive from the contract, we convert it into a liquidation penalty and return that.

gleiser-oliveira (approved):

PR #4868, opened 2025-10-02
feat: support intrinsic APYs

gleiser-oliveira:
## Jira ticket(s)

VPD-162

- Support intrinsic APYs


apps/evm/src/clients/api/queries/useGetPools/getPools/formatOutput/formatDistributions/formatRewardDistribution/index.ts
```diff
@@ -9,26 +10,31 @@ import type {
...
-  rewardDetails: TType extends 'merkl'
-    ? {
-        appName: string;
-        merklCampaignIdentifier: string;
-        description: string;
-        claimUrl: string;
-        tags: string[];
-      }
-    : null;
+  rewardDetails: TType extends 'merkl' ? MerklRewardDetails : IntrinsicApyRewardDetails | null;
```
therealemjy, line 34: I cant see this type used a `null` before and so you didn't introduce it, but in general I try to avoid using `null` in the frontend codebase at all due to its weird behavior.
How about replacing this line with that?

```suggestion
  rewardDetails?: TType extends 'merkl' ? MerklRewardDetails : IntrinsicApyRewardDetails;
```

therealemjy (approved):

apps/evm/src/clients/api/queries/useGetPools/getPools/formatOutput/formatDistributions/formatRewardDistribution/index.ts
gleiser-oliveira: `null` here is for `venus` rewards, which always return `rewardDetails: null`, I think this is the correct definition rather than make it optional, which would mean `rewardDetails` is not in the object.

therealemjy: The API returns `null`, but the frontend doesn't have to follow that logic. To me it makes more sense than in the case of Venus rewards, `rewardDetails` is `undefined` as it does not apply to this type of rewards.

gleiser-oliveira: Yeah, it already does it, this is the input type for the format function (input being the API data). Then after it's formatted rewardDetails is optional and only present if the type is `merkl` or `intrinsic`.

therealemjy: Oh yeah, turns out I can't read code. Ignore my stupidity...

gleiser-oliveira: Nah, don't say that, it was good to double check anyway! I'll proceed releasing this

PR #4914, opened 2025-10-15
feat: add vToken records to chains package

therealemjy:
## Jira ticket(s)

VPD-190

- add `useGetVTokens` hook
- retrieve vTokens from chains package

- generate and export vTokens for each chain by making RPC calls
- add `getRpcUrls` utility function
- export Viem chains associated with each chain ID


apps/evm/src/__mocks__/models/vTokens.ts
```diff
@@ -23,20 +23,69 @@ export const vUsdc: VToken = {
...
-export const vUsdt: VToken = {
+export const vUsdt_Core_Pool: VToken = {
   address: '0xb7526572FFE56AB9D7489838Bf2E18e3323b441A',
   decimals: 8,
   symbol: 'vUSDT',
   underlyingToken: usdt,
 };
 
-export const vBusd: VToken = {
+export const vUsdt_DeFi: VToken = {
```
gleiser-oliveira: nit: I think they could be just `camelCased`

apps/evm/src/clients/api/queries/getProposal/useGetProposal.ts
```diff
@@ -25,7 +25,7 @@ type Options = QueryObserverOptions<
   ]
 >;
 
-const { blockTimeMs: BSC_BLOCK_TIME_MS } = chains[governanceChain.id];
+const { blockTimeMs: BSC_BLOCK_TIME_MS } = chains[governanceChain.id as ChainId];
```
gleiser-oliveira: nit: wouldn't the type cast be better in the wallet lib?

packages/chains/src/tokens/underlyingTokens/bscMainnet.ts
```diff
@@ -401,4 +401,16 @@ export const tokens: Token[] = [
...
+    decimals: 6,
+    symbol: 'LUNA',
+    iconSrc: iconSrcs.luna,
+  },
+  {
+    address: '0x3d4350cD54aeF9f9b2C29435e0fa809957B3F30a',
+    decimals: 6,
+    symbol: 'UST',
+    iconSrc: iconSrcs.ust,
+  },
```
gleiser-oliveira, line 415: These intentionally came back?

therealemjy: Yes, because they are still returned by the contracts, even though the corresponding markets are unlisted.

apps/evm/src/clients/api/queries/getProposal/useGetProposal.ts
therealemjy: You're right. `governanceChain` is of the `Chain` type from viem, but doesn't need to be. I'll make a change.

apps/evm/src/pages/Proposal/Commands/NonBscCommand/index.tsx
```diff
@@ -1,15 +1,15 @@
...
+import { governanceChainId, useChainId } from 'libs/wallet';
 import { useMemo } from 'react';
 import { type RemoteProposal, RemoteProposalState } from 'types';
 import { Command } from '../Command';
 import { Description } from '../Description';
 import { CurrentStep } from './CurrentStep';
 import { ExecuteButton } from './ExecuteButton';
 
-const governanceChainMetadata = chains[governanceChain.id];
+const governanceChainMetadata = chains[governanceChainId as ChainId];
```
gleiser-oliveira: No need to cast anymore now that `governanceChainId` is already `ChainId`

gleiser-oliveira (approved):

PR #4941, opened 2025-10-22
feat: support off-chain APYs

gleiser-oliveira:
## Jira ticket(s)

VPD-204

- Support off-chain APYs


apps/evm/src/types/index.ts
```diff
@@ -97,12 +98,17 @@ export interface IntrinsicApyDistribution {
   };
 }
 
+export type OffChainApyDistribution = Omit<IntrinsicApyDistribution, 'type' | 'rewardDetails'> & {
+  type: 'off-chain';
+};
```
therealemjy: I'd suggest sharing a type here. Since this type could correspond to a lot of distribution types, I'd make it the generic type:

```suggestion
export interface GenericDistribution {
	type: 'intrinsic' | 'off-chain';
	token: Token;
	apyPercentage: BigNumber;
	dailyDistributedTokens: BigNumber;
	isActive: boolean;
	rewardDetails: {
	name: string;
	description: string;
	};
}
```

therealemjy (approved):

PR #4993, opened 2025-11-05
refactor: logic to simulate user operations

therealemjy:
Sorry for the massive PR, most of the line changes concern snapshots.
The most important changes to look at are these new functions:
- `apps/evm/src/hooks/useSimulateBalanceMutations/index.ts`
- `apps/evm/src/utilities/calculateUserPoolValues/index.ts`

And the updates to: `apps/evm/src/clients/api/queries/useGetPools/getPools/index.ts`, which uses the new utility functions to construct pool objects.

## Context

So far, each component (examples: `AccountData`, `OperationDetails`) that simulated account updates as the result of a transaction had to make the calculations itself. This lead to duplicated calculations and an implementation that assumed every transaction sent from the Market page forms would only affect one asset.

The goal of this PR is to start updating the logic so that any amount of mutations made to asset balances can be simulated and displayed on the UI.

- add `getSimulatedPool` query function that takes a list of balance mutations and a pool and returns a simulation of that pool after mutations
- add `useSimulateBalanceMutations` hook that uses the `getSimulatedPool` query function to simulate balance mutations
- add `AccountData2` component that acts as a temporary replacement for the `AccountData` component. This PR only updates the VAI page forms to use the new component, as updating all the forms within the dApp would make the PR even longer than it is. In a following PR, I'll update every form within the dApp that uses the old `AccountData` component then remove it from the codebase
- rename `userPercentOfLimit` property of `Asset` type to `userBorrowLimitSharePercentage`
- replace `userVaiBorrowBalanceTokens` and `userVaiBorrowBalanceCents` properties of `Pool` type with a `vai` object property that contains generic and user data related to VAI
- add `userYearlyEarningsCents` property to `Pool` type
- add `BalanceUpdates` component
- add `calculateUserPoolValues` utility function
- add `clampToZero` utility function
- add `addUserBorrowLimitShares` utility function
- remove `getVaiRepayApr` query function


apps/evm/src/pages/Vai/Repay/index.tsx
```diff
@@ -90,16 +89,16 @@ export const Repay: React.FC = () => {
...
-
   const { data: getPoolData, isLoading: isGetPoolDataLoading } = useGetPool({
     poolComptrollerAddress: chain.corePoolComptrollerContractAddress,
     accountAddress,
   });
+  const legacyPool = getPoolData?.pool;
+  const userVaiBorrowBalanceTokens = legacyPool?.vai?.userBorrowBalanceTokens;
 
-  const userVaiBorrowBalanceTokens = getPoolData?.pool.userVaiBorrowBalanceTokens;
+  console.log(userVaiBorrowBalanceTokens?.toFixed());
```
gleiser-oliveira: Stray log

therealemjy: Always and ever... Thanks, it's removed.

gleiser-oliveira (approved):

PR #5007, opened 2025-11-10
feat: add support for Fermi hardfork on BNB testnet

therealemjy:
- add support for Fermi hardfork on BNB testnet


apps/evm/src/clients/api/queries/getProposal/__tests__/__snapshots__/index.spec.ts.snap
```diff
@@ -21,7 +21,7 @@ exports[`getProposal > returns proposal in the correct format 1`] = `
       "version": "v2",
     },
     "endBlock": 44354696,
-    "endDate": 2025-04-03T08:07:46.250Z,
+    "endDate": 2024-10-31T07:54:17.750Z,
```
gleiser-oliveira, line 24: The calculated `endDate` is older because the block rate got faster, but ideally the snapshots wouldn't change (since the actual endDate already happened). I don't think this affects the UI in practice though - the date is registered in the subgraph instead of calculated by the start and end block diff.

therealemjy: So are you saying that `startBlock` and `endBlock` are calculated by the subgraph based on the block time?
Because we do calculate `startDate` and `endDate` on the frontend based on these and the block time: https://github.com/VenusProtocol/venus-protocol-interface/blob/main/apps/evm/src/clients/subgraph/utilities/formatToProposal/index.ts#L51-L56

So to me it makes sense that the snapshots are changing when the block time changes.

gleiser-oliveira: I mean there is no calculation on the subgraph, if a proposal is queued, executed or canceled it just gets the block timestamp of the event and stores it, so then it isn't affected by block times.

Here I just want to make sure this calculation isn't really used for past VIPs as it would alter their dates, ideally we just need the stored tx timestamp:

https://github.com/VenusProtocol/subgraphs/blob/main/subgraphs/venus-governance/src/operations/update.ts#L46

therealemjy: I get your point now, but unfortunately I don't think events are recorded for what `endDate` and `startDate` are recording here.
For example, no event is emitted when a proposal is defeated :/

gleiser-oliveira: Indeed, the date differences are quite significant here:

https://venus-testnet.vercel.app/#/governance/proposal/633?chainId=97
compared to
https://dapp-testnet-git-feat-bnb-testnet-fermi-hardfork-venus-protocol.vercel.app/#/governance/proposal/633?chainId=97

Should we support historical block times? Similar to what the API does

therealemjy: Sorry for the delay in my response, I was focused on the leveraged position features. When you talk about adding support for historical block times, do you mean on the subgraph side so that we could query the date when a proposal was deemed defeated for example?

gleiser-oliveira: No, it would actually be a frontend change, because the subgraph has no event to rely on.

What I mean here is that we can have a more accurate calculation, by using multiple block rates with timestamps associated with them, like in the backend code, instead of just the current block rate:

```
    case ChainId.BSC_TESTNET:
      return [
        { startTimestampMs: 1762741500000n, secondsPerBlock: 0.45 }, // block rate changed at 2025-11-10 02:25:00 AM UTC
        { startTimestampMs: 1748243100000n, secondsPerBlock: 0.75 }, // block rate changed at 2025-05-26 07:05:00 AM UTC
        { startTimestampMs: 1744097580000n, secondsPerBlock: 1.5 }, // block rate changed at 2025-04-08 07:33:00 AM UTC
        { startTimestampMs: 0n, secondsPerBlock: 3 },
 ```
 
And the logic to change would be getting the diff between the timestamps and calculating the number of blocks between them based on the block rate at that point in time. Which then means that for old proposal dates we would be using the block rate at that point in time, does it make sense? Second step to really be accurate is to consider all the block rates in that period of time (eg, if a proposal was created before a hardfork and then defeated after it - there would be two block rates to consider)

therealemjy: I see, you're right. I'll implement this logic in this PR.

therealemjy: I'll close this PR for now, David is working on implementing the solution you proposed and I don't think this PR brings much.

PR #5062, opened 2025-11-21
feat: add Repay with collateral UI

therealemjy:
## Jira ticket(s)

VPD-213

## Notes

This PR contains temporary code (encapsulated in `DEV ONY` and `END DEV ONLY` comments) to connect to a virtual testnet so that we can run tests before the mainnet environment is available. The changes can still be reviewed, but the PR should not be merged with `main` yet.

- add `useRepayWithCollateral` mutation function
- reduce default buffer amount when repaying full loan from 0.1% to 0.00001%
- fix `getSwapQuote` query function to handle all swap types properly
- update `SwapDetails` component to allow user to set the slippage up to 100% and display slippage >= 5% in red
- small design fix on `SelectTokenTextField` component
- update Boost form to allow users with no borrowing power if they have supplied tokens
- add inner tabs within Repay tab when `repayWithCollateral` feature is enabled


gleiser-oliveira (commented):
Overall I think the changes look good, I'd just wait until we are able to replace the dev sections with the actual prod code to give this a final look.

Other than that, I was thinking if you'd want the `find-swap/pcs` route to become just `find-swap`. I don't think we will have a use case for the Balmy SDK version and I could just remove it.

therealemjy:
If that's ok for you, I'd like to get the code merged (without the dev code nor the feature flag so that nothing is visible in production), just so we can continue working on it (there's several PRs incoming to polish things up).

We will test the feature extensively once the VIP is executed on Friday, if you're happy with the current code then any change needed should be to fix bugs (since the UI has only been QA tested and approved already).

About the route, we could yes. I also don't think we will be using Balmy at all so PCS will be the default provider for sure.

gleiser-oliveira (approved):

PR #5108, opened 2025-12-03
feat(evm,chain): add support to dynamic block times

david-sun-venus:
## Jira ticket(s)

VPD-336, VPD-337

-

### [landing app](https://github.com/VenusProtocol/venus-protocol-interface/tree/main/apps/landing/)
-

-

### [UI package](https://github.com/VenusProtocol/venus-protocol-interface/tree/main/apps/ui/)
-

### DIFF
--- a/apps/evm/src/clients/subgraph/utilities/formatToProposal/__tests__/index.spec.ts
+++ b/apps/evm/src/clients/subgraph/utilities/formatToProposal/__tests__/index.spec.ts
@@ -34,9 +34,11 @@ describe('getEstimatedDateByBlockHeight', () => {
   const maxwellBlockTime = blockTimes[2];
   const fermiBlockTime = blockTimes[3];
 
+  const LORENTZ_BLOCK_HEIGHT = 49791365;
+  const MAXWELL_BLOCK_HEIGHT = 52552978;
+
   // maxwell ->- target ->- current
   it('current time and target time (earlier) are both in the same block time range:', () => {
-    const MAXWELL_BLOCK_HEIGHT = 52552978;
     const nowDate = new Date('2025-05-26 07:05:16 AM UTC');
     vi.useFakeTimers().setSystemTime(nowDate);
 
@@ -54,8 +56,6 @@ describe('getEstimatedDateByBlockHeight', () => {
 
   //  target ->- maxwell ->- current
   it('current time and target time (earlier) are in different block time ranges (1 diff):', () => {
-    const MAXWELL_BLOCK_HEIGHT = 52552978;
-
     const currentBlock = MAXWELL_BLOCK_HEIGHT + 10; // now: maxwell upgrade block + 10
     const targetBlock = MAXWELL_BLOCK_HEIGHT - 10; // target: maxwell upgrade block - 10
 
@@ -75,8 +75,6 @@ describe('getEstimatedDateByBlockHeight', () => {
 
   // target ->- lorentz ->- maxwell ->- current
   it('current time and target time (earlier) are across two block time ranges:', () => {
-    const MAXWELL_BLOCK_HEIGHT = 52552978;
-
     const currentBlock = MAXWELL_BLOCK_HEIGHT + 10; // now: maxwell upgrade block + 10
 
     const lorentzEstimateHeight =
@@ -104,8 +102,6 @@ describe('getEstimatedDateByBlockHeight', () => {
 
   // target ->- maxwell ->- fermi ->- current
   it('current time (later after fermi) and target time (earlier) are across two block time ranges:', () => {
-    const MAXWELL_BLOCK_HEIGHT = 49791365;
-
     const fermiEstimateHeight =
       MAXWELL_BLOCK_HEIGHT +
       (fermiBlockTime.startTimestamp - maxwellBlockTime.startTimestamp) /
@@ -134,7 +130,6 @@ describe('getEstimatedDateByBlockHeight', () => {
 
   // maxwell ->- current ->- target -> fermi
   it('current time and target time (later) are both in the same block time range:', () => {
-    const MAXWELL_BLOCK_HEIGHT = 52552978;
     const nowDate = new Date('2025-05-26 07:05:08 AM UTC');
     vi.useFakeTimers().setSystemTime(nowDate);
 
@@ -152,8 +147,6 @@ describe('getEstimatedDateByBlockHeight', () => {
 
   // current ->- maxwell ->- target
   it('current time and target time (later) are in different block time ranges (1 diff):', () => {
-    const MAXWELL_BLOCK_HEIGHT = 52552978;
-
     const currentBlock = MAXWELL_BLOCK_HEIGHT - 10; // now: maxwell upgrade block - 10
     const targetBlock = MAXWELL_BLOCK_HEIGHT + 10; // target: maxwell upgrade block + 10
 
@@ -173,8 +166,6 @@ describe('getEstimatedDateByBlockHeight', () => {
 
   // current ->- lorentz ->- maxwell ->- target
   it('current time and target time (later) are across two block time ranges:', () => {
-    const LORENTZ_BLOCK_HEIGHT = 49791365;
-
     const currentBlock = LORENTZ_BLOCK_HEIGHT - 10; // now: lorentz upgrade block - 10
 
     const maxwellEstimateHeight =
@@ -203,8 +194,6 @@ describe('getEstimatedDateByBlockHeight', () => {
 
   // current ->- maxwell ->- fermi ->- target
   it('current time and target time (later after Fermi) are across two block time ranges:', () => {
-    const MAXWELL_BLOCK_HEIGHT = 49791365;
-
     const currentBlock = MAXWELL_BLOCK_HEIGHT - 10; // now: maxwell upgrade block - 10
 
     const fermiEstimateHeight =
--- a/apps/evm/src/clients/subgraph/utilities/formatToProposal/__tests__/index.spec.ts
+++ b/apps/evm/src/clients/subgraph/utilities/formatToProposal/__tests__/index.spec.ts
@@ -6,7 +6,7 @@ import BigNumber from 'bignumber.js';
 import type { ProposalsQuery } from 'clients/subgraph/gql/generated/governanceBsc';
 import { governanceChainId } from 'libs/wallet';
 import { formatToProposal } from '..';
-import { getEstimateDateByBlockHeight } from '../getEstimateDateByBlockHeight';
+import { getEstimatedDateByBlockHeight } from '../getEstimatedDateByBlockHeight';
 
 describe('formatToProposal', () => {
   beforeEach(() => {
@@ -26,49 +26,47 @@ describe('formatToProposal', () => {
   });
 });
 
-describe('getEstimateDateByBlockHeight', () => {
+describe('getEstimatedDateByBlockHeight', () => {
   const { blockTimes = [] } = chains[governanceChainId] ?? {};
-  const genesisBlockTime = blockTimes.find(
-    item => item.blockTimeMs === 3000,
-  ) as (typeof blockTimes)[number];
 
-  const lorentzBlockTime = blockTimes.find(
-    item => item.blockTimeMs === 1500,
-  ) as (typeof blockTimes)[number];
+  const genesisBlockTime = blockTimes[0];
+  const lorentzBlockTime = blockTimes[1];
+  const maxwellBlockTime = blockTimes[2];
+  const fermiBlockTime = blockTimes[3];
 
-  const maxwellBlockTime = blockTimes.find(
-    item => item.blockTimeMs === 750,
-  ) as (typeof blockTimes)[number];
-
-  const fermiBlockTime = blockTimes.find(
-    item => item.blockTimeMs === 450,
-  ) as (typeof blockTimes)[number];
-
-  // maxwell ->- target ->- current ->- fermi
+  // maxwell ->- target ->- current
   it('current time and target time (earlier) are both in the same block time range:', () => {
-    const maxwellBlockHeight = 52552978;
+    const MAXWELL_BLOCK_HEIGHT = 52552978;
     const nowDate = new Date('2025-05-26 07:05:16 AM UTC');
     vi.useFakeTimers().setSystemTime(nowDate);
 
-    const currentBlock = maxwellBlockHeight + 20; // now: maxwell upgrade block + 20
-    const targetBlock = maxwellBlockHeight + 10; // target: maxwell upgrade block + 10
+    const currentBlock = MAXWELL_BLOCK_HEIGHT + 20; // now: maxwell upgrade block + 20
+    const targetBlock = MAXWELL_BLOCK_HEIGHT + 10; // target: maxwell upgrade block + 10
 
-    const estimateTime = getEstimateDateByBlockHeight(targetBlock, currentBlock, blockTimes);
+    const estimateTime = getEstimatedDateByBlockHeight({
+      targetBlockHeight: targetBlock,
+      currentBlockHeight: currentBlock,
+      blockTimes,
+    });
 
     expect(estimateTime).toBeCloseTo(nowDate.getTime() - maxwellBlockTime.blockTimeMs * 10);
   });
 
   //  target ->- maxwell ->- current
-  it('current time and target time (eariler) are in different block time ranges (1 diff):', () => {
-    const maxwellBlockHeight = 52552978;
+  it('current time and target time (earlier) are in different block time ranges (1 diff):', () => {
+    const MAXWELL_BLOCK_HEIGHT = 52552978;
 
-    const currentBlock = maxwellBlockHeight + 10; // now: maxwell upgrade block + 10
-    const targetBlock = maxwellBlockHeight - 10; // target: maxwell upgrade block - 10
+    const currentBlock = MAXWELL_BLOCK_HEIGHT + 10; // now: maxwell upgrade block + 10
+    const targetBlock = MAXWELL_BLOCK_HEIGHT - 10; // target: maxwell upgrade block - 10
 
     const nowDate = new Date(maxwellBlockTime.startTimestamp + 10 * maxwellBlockTime.blockTimeMs);
     vi.useFakeTimers().setSystemTime(nowDate);
 
-    const estimateTime = getEstimateDateByBlockHeight(targetBlock, currentBlock, blockTimes);
+    const estimateTime = getEstimatedDateByBlockHeight({
+      targetBlockHeight: targetBlock,
+      currentBlockHeight: currentBlock,
+      blockTimes,
+    });
 
     expect(estimateTime).toBeCloseTo(
       nowDate.getTime() - maxwellBlockTime.blockTimeMs * 10 - lorentzBlockTime.blockTimeMs * 10,
@@ -76,13 +74,13 @@ describe('getEstimateDateByBlockHeight', () => {
   });
 
   // target ->- lorentz ->- maxwell ->- current
-  it('current time and target time (eariler) are across two block time ranges:', () => {
-    const maxwellBlockHeight = 52552978;
+  it('current time and target time (earlier) are across two block time ranges:', () => {
+    const MAXWELL_BLOCK_HEIGHT = 52552978;
 
-    const currentBlock = maxwellBlockHeight + 10; // now: maxwell upgrade block + 10
+    const currentBlock = MAXWELL_BLOCK_HEIGHT + 10; // now: maxwell upgrade block + 10
 
     const lorentzEstimateHeight =
-      maxwellBlockHeight -
+      MAXWELL_BLOCK_HEIGHT -
       (maxwellBlockTime.startTimestamp - lorentzBlockTime.startTimestamp) /
         lorentzBlockTime.blockTimeMs;
     const targetBlock = lorentzEstimateHeight - 10; // target: estimated lorentz upgrade block - 10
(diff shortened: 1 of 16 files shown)


packages/chains/src/chains/chainMetadata/index.ts
```diff
@@ -26,8 +42,28 @@ export const chains: {
...
-    blocksPerDay: 115200,
+    blockTimes: [
+      {
+        startTimestamp: new Date('2020-06-09 07:57:19 AM UTC').getTime(), // Block #0
+        blockHeight: 0,
+        blockTimeMs: 3000,
+      },
+      {
+        startTimestamp: new Date('2025-04-08 07:33:00 AM UTC').getTime(), // Lorentz Upgrade
+        blockHeight: 49791365,
```
therealemjy: I'd suggest not setting the block height of each hardfork, because we won't know in advance the block time of a hardfork that's yet to happen. We should only use the timestamp as reference.

packages/chains/src/chains/chainMetadata/index.ts
```diff
@@ -160,3 +236,31 @@ export const chains: {
     nativeToken: eth,
   },
 };
+
+const MS_PER_DAY = 1000 * 60 * 60 * 24;
+
+export const getBlockTimeByChainId = (chainId: ChainId, targetTimestamp = new Date().getTime()) => {
+  const targetChain = chains?.[chainId ?? ''];
```
therealemjy: `chains` and `chainId` can not be `undefined`, so the optional chaining and the fallback are unnecessary.

```suggestion
  const targetChain = chains[chainId];
```

packages/chains/src/chains/chainMetadata/index.ts
```diff
@@ -160,3 +236,31 @@ export const chains: {
...
+  if (!targetChain || !Array.isArray(targetChain.blockTimes)) return null;
```
therealemjy: This code is unnecessary since `targetChain` will always be defined.

packages/chains/src/chains/chainMetadata/index.ts
```diff
@@ -160,3 +236,31 @@ export const chains: {
...
+
+  // default to latest
+  let blockTime = targetChain.blockTimes?.slice(-1)?.[0];
```
therealemjy:
```suggestion
  let blockTime = targetChain.blockTimes[0];
```

packages/chains/src/chains/chainMetadata/index.ts
```diff
@@ -160,3 +236,31 @@ export const chains: {
...
+  }
+
+  const { blockTimeMs, blockHeight, startTimestamp } = blockTime;
+
+  return {
+    startTimestamp,
+    blockTimeMs,
+    blockHeight,
+    blocksPerDay: blockTime.blockTimeMs !== 0 ? MS_PER_DAY / blockTime.blockTimeMs : 0,
+  };
```
therealemjy: I think there's two issues here. The first one is that the first block will always return the first timestamp, because your timestamps are ordered in ascending order in the array.

The second one is that you're only considering one block time, but the block time of a network is evolving over time. Let's imagine that the following hardforks happened:

Hard fork #1
Timestamp: 13/01/24
New block time: 1s

Hard fork #2
Timestamp: 01/04/24
New block time: 500ms

Hard fork #3
Timestamp: 07/10/24
New block time: 250ms

Now let's say that you want to calculate the timestamp of a block that happened after hardfork #3. You'll need to account for all three hardforks, as some blocks happened between the genesis block (the first block of a chain) and hardfork #1, others happened between hardfork #1 and #2, some happened between hardfork #2 and #3, and some happened between hardfork #3 and now.

packages/chains/src/chains/chainMetadata/index.ts
```diff
@@ -160,3 +236,31 @@ export const chains: {
     nativeToken: eth,
   },
 };
+
+const MS_PER_DAY = 1000 * 60 * 60 * 24;
```
therealemjy: We have [this file](https://github.com/VenusProtocol/venus-protocol-interface/blob/b1e40a4e380b7d6fac0cba0063ca347ef5f4dc05/apps/evm/src/constants/time.ts) that contains some time constants, you could add this one to it.

david-sun-venus: This is in package/chains, and constant file is in apps/evm. I could move it to constants under chains

packages/chains/src/chains/chainMetadata/index.ts
david-sun-venus: Not all chains config has blockTimes, and targetChain could be undefined with incorrect chainId, so I would say it's safer to keep the check rather than having users seeing the "cannot get 'blockTimes' of undefined" error.

packages/chains/src/chains/chainMetadata/index.ts
david-sun-venus: I had the startTimestamp in config which was the actual timestamp from the bscScan / opBNBScan for the corresponding block height which starts with the new block time, so my take is that if the estimate time is between #2 and #3, I'll just use startTimestamp#2 + (targetBlockHeight - BlockHeight#2) * BlockTime#2.

packages/chains/src/chains/chainMetadata/index.ts
david-sun-venus: I think the latest block time is most likely to be the default to fallback to.

packages/chains/src/chains/chainMetadata/index.ts
therealemjy: `chainId` will never be incorrect, or a TS error would be thrown at build time. That's why `targetChain` will always be defined.

packages/chains/src/chains/chainMetadata/index.ts
therealemjy: Good point, let's leave this constant here then.

packages/chains/src/chains/chainMetadata/index.ts
therealemjy: But then how will you know the exact block height of hard forks that are yet to happen?

apps/evm/src/clients/subgraph/utilities/formatToProposal/index.ts
```diff
@@ -26,7 +26,24 @@ import {
 import { formatToProposalActions } from './formatToProposalActions';
 import { formatToRemoteProposal } from './formatToRemoteProposal';
 
-const { blockTimeMs: BSC_BLOCK_TIME_MS } = chains[governanceChainId];
+const { blockTimes = [] } = chains[governanceChainId] ?? {};
+
+const getEstimateDateByBlockHeight = (
+  targetBlockHeight: number,
+  _blockTimes: Chain['blockTimes'] = [],
```
therealemjy: Conventionally speaking, prefixed a parameter with an underscore indicates that it is unused.
```suggestion
blockTimes: Chain['blockTimes'] = [],
```

apps/evm/src/clients/subgraph/utilities/formatToProposal/index.ts
```diff
@@ -26,7 +26,24 @@ import {
...
+) => {
+  let closestBlockTime: Chain['blockTimes'][number] | null = null;
```
therealemjy: Let's not use `null` ever.

apps/evm/src/utilities/generatePseudoRandomRefetchInterval/index.ts
```diff
@@ -1,3 +1,12 @@
-export const generatePseudoRandomRefetchInterval = () =>
+export const generatePseudoRandomRefetchInterval = (speed?: 'FAST' | 'MODERATE') => {
```
therealemjy: To follow the codebase standards:
```suggestion
export const generatePseudoRandomRefetchInterval = (speed?: 'fast' | 'moderate') => {
```

packages/chains/src/chains/chainMetadata/index.ts
therealemjy: You don't need to default to the latest block time since the block right after would find the right block time by going through the array.

packages/chains/src/chains/chainMetadata/index.ts
david-sun-venus: We can't. Instead we could deploy a new version with updated configs after the hard fork happens. The deployment of our app based on new hard forks is inevitable, and it's just the matter of before or after.

therealemjy: That's not the goal no, the dApp needs to update in realtime without needing to be redeployed so that users get accurate block times when the hardfork happens.

apps/evm/src/clients/subgraph/utilities/formatToProposal/index.ts
david-sun-venus: > Conventionally speaking, prefixed a parameter with an underscore indicates that it is unused.

From my experience, it's either that or being used locally (within a inner context) when it's hard to find a different variable name from outer context. I've defined blockTimes in line#29, so it'll be confusing to use the same name inside.

packages/chains/src/chains/chainMetadata/index.ts
```diff
@@ -160,3 +236,31 @@ export const chains: {
...
+
+export const getBlockTimeByChainId = (chainId: ChainId, targetTimestamp = new Date().getTime()) => {
+  const targetChain = chains?.[chainId ?? ''];
+  if (!targetChain || !Array.isArray(targetChain.blockTimes)) return null;
+
+  // default to latest
+  let blockTime = targetChain.blockTimes?.slice(-1)?.[0];
+
+  // Find the right blockTime based on targetTimestamp
+  if (targetChain.blockTimes.length > 1) {
```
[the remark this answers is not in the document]
david-sun-venus: ```!Array.isArray(targetChain.blockTimes)``` in line#244 will make sure it's an array (isArray() will return false for nil values, too)

apps/evm/src/clients/subgraph/utilities/formatToProposal/getEstimateDateByBlockHeight.ts
```diff
@@ -0,0 +1,111 @@
+import type { Chain } from '@venusprotocol/chains';
+
+export const getEstimateDateByBlockHeight = (
+  targetBlockHeight: number,
+  currentBlockHeight: number,
+  blockTimes: NonNullable<Chain['blockTimes']>,
```
therealemjy: `NonNullable` isn't necessary here. If you use a separate type to define `blockTimes` (I'll write a comment about this in the `types/index.ts` file), then you can simplify things:

```suggestion
  blockTimes: BlockTime[];
```

apps/evm/src/clients/subgraph/utilities/formatToProposal/getEstimatedDateByBlockHeight/index.ts
therealemjy, line 1: getEstimateDateByBlockHeight -> getEstimatedTimestampByBlockHeight
("estimate" to "estimated" and "date" to "timestamp" since that's the term you use in your logic).

To follow the codebase practices, let's create a `getEstimatedDateByBlockHeight` folder then add an `index.ts` file inside it that contains the `getEstimatedDateByBlockHeight` function.

packages/chains/src/types/index.ts
```diff
@@ -51,6 +51,7 @@ export interface Chain {
   safeWalletApiUrl?: string;
   proposalExecutionGracePeriodMs?: number;
   blockTimeMs?: number;
+  blockTimes?: { startTimestamp: number; blockTimeMs: number; blockHeight?: number }[];
```
therealemjy: I'd recommend defining this as a separate type so that it can be used in other places instead of having to type `typeof blocktimes`.

Also, `blockHeight` should be removed since it isn't used.

apps/evm/src/clients/subgraph/utilities/formatToProposal/getEstimateDateByBlockHeight.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  const getBlockTimeByTimestamp = (timestamp: number) => {
+    let targetBlockTime: (typeof blockTimes)[number] | undefined;
+
+    blockTimesWithIndex.forEach(item => {
+      if (timestamp > item.startTimestamp) {
+        targetBlockTime = item;
+      }
+    });
+    return targetBlockTime as (typeof blockTimes)[number] & { index: number };
```
therealemjy: While the `targetBlockTime` will always be defined because the first timestamp in the `blockTimes` array is 0, TypeScript doesn't know that and casting the type here removes this typing error which could lead to problems. To prevent that, you can set the first item in the list as a default.

```suggestion
    let targetBlockTime: BlockTime = blockTimes[0];

    blockTimes.forEach(item => {
      if (timestamp > item.startTimestamp) {
        targetBlockTime = item;
      }
    });

    return targetBlockTime;
```

apps/evm/src/clients/subgraph/utilities/formatToProposal/getEstimateDateByBlockHeight.ts
```diff
@@ -0,0 +1,111 @@
...
+  // When target time falls in the same block time as now, no further calcuation required
+  if (estimateBlockTime.blockTimeMs === nowBlockTime.blockTimeMs) {
+    return new Date(estimateTargetTime);
+  }
+
+  // When target is earlier than now, remove the blockTimes after now and reverse; otherwise, remove the blockTimes before now.
+  const amendedBlockTimes =
+    totalBlockDiff > 0
+      ? blockTimesWithIndex.filter(item => item.index <= nowBlockTime.index).reverse()
+      : blockTimesWithIndex.filter(item => item.index >= nowBlockTime.index);
```
therealemjy: This logic seems to be the reason why you've indexed block times, but since each block time has a `startTimestamp` property you don't need the index at all:

```suggestion
      ? blockTimes.filter(item => item.startTimestamp <= nowBlockTime.startTimestamp).reverse()
      : blockTimes.filter(item => item.startTimestamp >= nowBlockTime.startTimestamp);
```

This means you can also get rid of `blockTimesWithIndex`.

apps/evm/src/clients/subgraph/utilities/formatToProposal/getEstimateDateByBlockHeight.ts
```diff
@@ -0,0 +1,111 @@
...
+        const remainingBlockCount = accu.remainingBlockCount - blockDiff;
+
+        if (remainingBlockCount <= 0) {
+          return {
+            timestamp: accu.timestamp - accu.remainingBlockCount * curr.blockTimeMs,
+            remainingBlockCount,
+          };
+        }
+
+        // need to go further blockTime
```
therealemjy:
```suggestion
        // need to traverse more block times
```

apps/evm/src/clients/subgraph/utilities/formatToProposal/getEstimatedDateByBlockHeight/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    totalBlockDiff > 0
+      ? blockTimesWithIndex.filter(item => item.index <= nowBlockTime.index).reverse()
+      : blockTimesWithIndex.filter(item => item.index >= nowBlockTime.index);
+
+  let ret: number;
+
+  if (totalBlockDiff > 0) {
+    const { timestamp } = amendedBlockTimes.reduce(
+      (accu, curr) => {
+        if (accu.remainingBlockCount <= 0) return accu;
```
therealemjy, line 53: Let's not use nested IF statements. You can use early returns or functions to avoid this.

apps/evm/src/clients/subgraph/utilities/formatToProposal/getEstimateDateByBlockHeight.ts
```diff
@@ -0,0 +1,111 @@
...
+    return targetBlockTime as (typeof blockTimes)[number] & { index: number };
+  };
+
+  const now = new Date().getTime();
+  const totalBlockDiff = currentBlockHeight - targetBlockHeight; // negative: target is in the future
+  const nowBlockTime = getBlockTimeByTimestamp(now);
+
+  // Try to calculate target time using block time for now.
+  const estimateTargetTime = now - nowBlockTime.blockTimeMs * totalBlockDiff;
+  const estimateBlockTime = getBlockTimeByTimestamp(estimateTargetTime);
```
therealemjy:
```suggestion
  const estimatedTargetTime = now - nowBlockTime.blockTimeMs * totalBlockDiff;
  const estimatedBlockTime = getBlockTimeByTimestamp(estimateTargetTime);
```

packages/chains/src/chains/chainMetadata/index.ts
```diff
@@ -160,3 +236,30 @@ export const chains: {
     nativeToken: eth,
   },
 };
+
+const MS_PER_DAY = 1000 * 60 * 60 * 24;
+
+export const getBlockTimeByChainId = (chainId: ChainId, targetTimestamp = new Date().getTime()) => {
```
therealemjy: This function should be in its own file, inside the `apps/packages/chains/src/utilities` folder.

Also and to follow codebase practices, we generally pass arguments in the form of an object (so that each property has a name when being called, which can prevent mistakes):

```suggestion
export const getBlockTimeByChainId = ({
    chainId,
    targetTimestamp = new Date().getTime(),
  }: { chainId: ChainId; targetTimestamp: number }) => {};
```

packages/chains/src/chains/chainMetadata/index.ts
```diff
@@ -26,8 +42,28 @@ export const chains: {
...
+      {
+        startTimestamp: new Date('2025-05-26 07:05:00 AM UTC').getTime(), // Maxwell Upgrade
+        // blockHeight: 52552978,
+        blockTimeMs: 750,
+      },
+      {
+        startTimestamp: new Date('2025-11-10 02:25:00 AM UTC').getTime(), // Fermi Upgrade
+        blockTimeMs: 450,
+        // blockHeight: 71859051,
+      },
```
therealemjy, line 58: Let's remove the comment `blockHeight` properties since they aren't used.

packages/chains/src/types/index.ts
```diff
@@ -51,6 +51,7 @@ export interface Chain {
   safeWalletApiUrl?: string;
   proposalExecutionGracePeriodMs?: number;
   blockTimeMs?: number;
```
therealemjy: This property needs to be removed since it's being replaced by `blockTimes`.

apps/evm/src/clients/subgraph/utilities/formatToProposal/getEstimateDateByBlockHeight.ts
```diff
@@ -0,0 +1,111 @@
...
+
+export const getEstimateDateByBlockHeight = (
+  targetBlockHeight: number,
+  currentBlockHeight: number,
+  blockTimes: NonNullable<Chain['blockTimes']>,
+) => {
+  const blockTimesWithIndex = blockTimes.map((item, index) => ({
+    ...item,
+    index,
+  }));
```
therealemjy: You don't need to index the block times since they already come with a timestamp property you can use for comparison (I'll add another comment about this).

apps/evm/src/clients/subgraph/utilities/formatToProposal/getEstimateDateByBlockHeight.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  const getBlockTimeByTimestamp = (timestamp: number) => {
```
therealemjy: This can be defined as a separate function so that it's not encapsulated here (to make the code easier to read). To follow the codebase practices, this function would be inside a separate getBlockTimeByTimestamp folder (within the getEstimateDateByBlockHeight folder) containing an index.ts file with this code in it.

apps/evm/src/clients/subgraph/utilities/formatToProposal/__tests__/index.spec.ts
```diff
@@ -22,3 +25,187 @@ describe('formatToProposal', () => {
...
+    item => item.blockTimeMs === 1500,
+  ) as (typeof blockTimes)[number];
+
+  const maxwellBlockTime = blockTimes.find(
+    item => item.blockTimeMs === 750,
+  ) as (typeof blockTimes)[number];
+
+  const fermiBlockTime = blockTimes.find(
+    item => item.blockTimeMs === 450,
+  ) as (typeof blockTimes)[number];
```
therealemjy:
```suggestion
  const genesisBlockTime = blockTimes[0];
  const lorentzBlockTime = blockTimes[1];
  const maxwellBlockTime = blockTimes[2];
  const fermiBlockTime = blockTimes[3];
```

apps/evm/src/clients/subgraph/utilities/formatToProposal/__tests__/index.spec.ts
```diff
@@ -22,3 +25,187 @@ describe('formatToProposal', () => {
...
+
+  // maxwell ->- target ->- current ->- fermi
```
therealemjy: According to the data used in the test:
```suggestion
  // maxwell ->- target ->- current
```

apps/evm/src/clients/subgraph/utilities/formatToProposal/__tests__/index.spec.ts
```diff
@@ -22,3 +25,187 @@ describe('formatToProposal', () => {
...
+
+    const estimateTime = getEstimateDateByBlockHeight(targetBlock, currentBlock, blockTimes);
+
+    expect(estimateTime).toBeCloseTo(
+      nowDate.getTime() - maxwellBlockTime.blockTimeMs * 10 - lorentzBlockTime.blockTimeMs * 10,
+    );
+  });
+
+  // target ->- lorentz ->- maxwell ->- current
+  it('current time and target time (eariler) are across two block time ranges:', () => {
```
therealemjy:
```suggestion
  it('current time and target time (earlier) are across two block time ranges:', () => {
```

apps/evm/src/clients/subgraph/utilities/formatToProposal/__tests__/index.spec.ts
```diff
@@ -22,3 +25,187 @@ describe('formatToProposal', () => {
...
+    const targetBlock = maxwellBlockHeight + 10; // target: maxwell upgrade block + 10
+
+    const estimateTime = getEstimateDateByBlockHeight(targetBlock, currentBlock, blockTimes);
+
+    expect(estimateTime).toBeCloseTo(nowDate.getTime() - maxwellBlockTime.blockTimeMs * 10);
+  });
+
+  //  target ->- maxwell ->- current
+  it('current time and target time (eariler) are in different block time ranges (1 diff):', () => {
+    const maxwellBlockHeight = 52552978;
```
therealemjy: Since you're reusing this, you can set it as a constant at the top of the file.


```suggestion
    const MAXWELL_BLOCK_HEIGHT = 52552978;
```

therealemjy (changes requested):

apps/evm/src/clients/subgraph/utilities/formatToProposal/__tests__/index.spec.ts
```diff
@@ -22,3 +25,209 @@ describe('formatToProposal', () => {
...
+  const { blockTimes = [] } = chains[governanceChainId] ?? {};
+
+  const genesisBlockTime = blockTimes[0];
+  const lorentzBlockTime = blockTimes[1];
+  const maxwellBlockTime = blockTimes[2];
+  const fermiBlockTime = blockTimes[3];
+
+  // maxwell ->- target ->- current
+  it('current time and target time (earlier) are both in the same block time range:', () => {
+    const MAXWELL_BLOCK_HEIGHT = 52552978;
```
therealemjy: This is shared by multiple tests, so you can declare it at the top of the file and share the same constant between all tests (without having to redeclare it in each test).

therealemjy (commented):
Nice, only one minor comment about a constant that can be shared across tests, but that's a detail.

Feel free to move your code to a new branch coming from the `main` branch so that the CI pipelines can run. This will also provide us test environment to check the result on proposals.

david-sun-venus:
Migrated to this PR: https://github.com/VenusProtocol/venus-protocol-interface/pull/5130

PR #5130, opened 2025-12-10
feat: support dynamic block times

david-sun-venus:
## Jira ticket(s)

VPD-336, VPD-337

- Improve estimation precision for proposals' date displayed under governance

- Add block times of hardforks config to chains metadata, up to Fermi hardfork and Fourier upgrade.
- Support getting the blockTimeMs and blocksPerDay of the correct hardfork from giving timestamp.


apps/evm/src/clients/api/queries/useGetPools/__tests__/__snapshots__/index.spec.ts.snap
```diff
@@ -76,8 +76,8 @@ exports[`useGetPools > returns pools in the correct format 1`] = `
           "borrowPointDistributions": [],
           "borrowTokenDistributions": [
             {
-              "apyPercentage": "5.623734327743302",
-              "dailyDistributedTokens": "1000.0000512",
+              "apyPercentage": "4.188904668290827",
+              "dailyDistributedTokens": "750.0000384",
```
therealemjy: It seems odd that these would get updated, as they are based on the latest block time anyway (so even after your updates, the block time reference should be the same).

therealemjy (approved):

PR #5132, opened 2025-12-11
feat(evm,chain): add auto-reload when passing hardforks

david-sun-venus:
## Jira ticket(s)

VPD-366

- Add auto-reload ui after passing hardfork

- Remove unused constant

### DIFF
--- a/apps/evm/src/App/ChainUpgradeHandler/__tests__/index.spec.tsx
+++ b/apps/evm/src/App/ChainUpgradeHandler/__tests__/index.spec.tsx
@@ -1,0 +1,37 @@
+import { renderComponent } from 'testUtils/render';
+
+import { ChainUpgradeHandler } from 'App/ChainUpgradeHandler';
+import type { MockInstance } from 'vitest';
+
+const runtimeTimestamp = new Date().getTime();
+const upgradeTimestamps = [new Date(runtimeTimestamp + 2000)];
+const fakeNowMs = runtimeTimestamp + 12000;
+
+describe('ChainUpgradeHandler', () => {
+  let reloadSpy: MockInstance;
+
+  beforeEach(() => {
+    // 1. Redefine the 'reload' property on window.location to be configurable.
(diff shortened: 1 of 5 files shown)


apps/evm/src/App/index.tsx
```diff
@@ -22,61 +21,65 @@ const GaslessChecker = safeLazyLoad(() => import('containers/GaslessChecker'));
...
-          <meta name="robots" content="noindex" />
-        </Helmet>
-      )
-    }
+const App = () => {
+  const { hardforks } = useChain();
+  const upgradeTimestamps = useMemo(
+    () => hardforks?.map(hardfork => new Date(hardfork.startTimestamp)) ?? [],
+    [hardforks],
+  );
```
therealemjy: React 19 has auto memoization built in, so for cases like this you don't need to use`useMemo` (especially since this is anyway a cheap operation).

```suggestion
  const upgradeTimestamps = (hardforks || []).map(hardfork => new Date(hardfork.startTimestamp));
```

apps/evm/src/App/ChainUpgradeHandler/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+    // This is necessary because it is non-configurable by default in jsdom.
+    Object.defineProperty(window, 'location', {
+      configurable: true,
+      value: () => {}, // Provide a dummy function initially
+    });
+
+    Object.defineProperty(window.location, 'reload', {
+      configurable: true,
+      value: () => {}, // Provide a dummy function initially
+    });
```
therealemjy, line 25: I wonder if this test suite should reset the window.location.reload property after running (so in an `afterEach` function call)?

david-sun-venus: Yes, but it used to give me errors when I added it, so I removed it. Now I think I know why it caused the error before, let me add it back with fixes.

therealemjy (approved):

PR #5151, opened 2025-12-16
feat(evm): add auto-reload when passing hardforks (updated)

david-sun-venus:
## Jira ticket(s)

VPD-366

- Add auto-reload ui after passing hardfork

- Remove unused constant


apps/evm/src/App/index.tsx
```diff
@@ -1,5 +1,3 @@
-import { opBnbTestnetFourierForkTimestamp } from '@venusprotocol/chains';
```
therealemjy, line 1: This variable doesn't need to be exported from the chains package anymore then.

therealemjy (approved):

PR #5196, opened 2025-12-25
fix: only filter by user assets if they have some

therealemjy:
- only filter by user assets if they have some


david-sun-venus (approved):
lgtm

PR #5197, opened 2025-12-30
feat: add high price impact acknowledgement to Boost form

therealemjy:
- add high price impact acknowledgement to Boost form
- fix issue with supply cap being incorrectly validated during "Repay with collateral" flow


apps/evm/src/pages/Market/OperationForm/Repay/index.tsx
```diff
@@ -39,6 +40,12 @@ export const Repay: React.FC<RepayProps> = ({ asset, pool }) => {
     return repayWithWalletBalanceFormDom;
   }
 
+  const hasCollateral = pool.assets.some(
+    asset =>
+      // Skip vBNB && Skip tokens for which user has no supply
+      asset.vToken.symbol !== 'vBNB' && asset.userSupplyBalanceCents.isGreaterThan(0),
+  );
```
therealemjy: There's no need to filter vBNB out, we already do that line 25 because the `repayWithCollateral` token action is disabled for vBNB. Also, it's not just the user supply balance of the asset that needs to be checked, but rather the borrow limit of the user within that pool (other assets can contribute towards the borrow limit of the user is they have enabled them as collateral).

The check should be:
```tsx
const hasCollateral = pool.userBorrowLimitCents?.isGreaterThan(0);
```

apps/evm/src/pages/Market/OperationForm/Repay/index.tsx
```diff
@@ -48,7 +55,11 @@ export const Repay: React.FC<RepayProps> = ({ asset, pool }) => {
     {
       id: 'repayWithCollateral',
       title: t('operationForm.repayTab.collateralTabTitle'),
-      content: <RepayWithCollateralForm pool={pool} asset={asset} />,
+      content: hasCollateral ? (
+        <RepayWithCollateralForm pool={pool} asset={asset} />
+      ) : (
+        <DisabledActionNotice token={asset.vToken.underlyingToken} action={'repay'} />
```
therealemjy: The action isn't disabled, instead it's a warning with the message I gave you in DM that we should display ("You do not have any collateral to repay with"):

```tsx
 <NoticeWarning
    description={t('operationForm.repayTab.warning.noCollateral')}
 />
```

david-sun-venus (approved):
lgtm

PR #5217, opened 2026-01-07
feat(evm): migrate landing to evm

david-sun-venus:
## Jira ticket(s)

VPD-416

- New Landing


apps/evm/src/components/Link/index.tsx
therealemjy, line 1: This component is specific to the Landing page, so it shouldn't be added to the `components` directory

apps/evm/src/components/Container/index.tsx
therealemjy, line 1: This component is specific to the Landing page, so it shouldn't be added to the `components` directory

apps/evm/src/clients/api/queries/getMarketsAndTvl/index.ts
therealemjy, line 1: No need to import his query function, it won't be used in the final landing page

apps/evm/src/pages/Landing/index.tsx
therealemjy, line 1: The name of this directory should be lowercased (`landing`).

apps/evm/src/constants/routing.ts
```diff
@@ -23,6 +26,9 @@ export enum Subdirectory {
 
 const routeSubdirectories = {
   landing: [Subdirectory.LANDING],
+  landingTerms: [Subdirectory.LANDING, Subdirectory.LANDING_TERMS],
+  landingPolicy: [Subdirectory.LANDING, Subdirectory.LANDING_POLICY],
+  landingDiscord: [Subdirectory.LANDING, Subdirectory.LANDING_DISCORD],
```
therealemjy: Since these pages will concern the entire app (and not just its landing page), they can be placed at the root.

```suggestion
  terms: [Subdirectory.TERMS],
  policy: [Subdirectory.POLICY],
  discord: [Subdirectory.DISCORD],
```

apps/evm/public/images/landing/governance/planets.png
therealemjy: Let's move this image and the other images placed in the public folder inside their respective components.

The public folder is used to make images public, generally because they'll be used in externals apps. For images that are only relevant to the dApp, they should be placed inside their component's directory (you can check existing pages/components to see how this is done).

apps/evm/src/assets/styles/index.css
```diff
@@ -9,6 +9,13 @@
 @import '@venusprotocol/ui/fonts/proximaNova' layer(base);
 @import '@venusprotocol/ui/theme' layer(theme);
 
+@theme {
+  /* Legacy colors which is used by old landing: */
+  --color-background-secondary: #1E2431;
+  --color-icon-background: #2D3549;
+  --global-max-width: 1280px;
```
therealemjy: Use arbitrary values for these instead of updating the main theme, as we don't want to keep these here.

apps/evm/src/assets/styles/index.css
```diff
@@ -72,6 +79,14 @@
...
     }
   }
+
+  .Container {
+    @apply w-full px-4 sm:px-6 md:px-8 lg:px-12 lg:mx-auto
+  }
+
+  .LandingLink {
+    @apply hover:no-underline! hover:text-blue-hover! duration-250;
+  }
```
therealemjy: Same idea here, don't create custom classes for Tailwind class chains. If `LandingLink` is used in multiple places, then it should be component not a class.

apps/evm/src/constants/landing.ts
therealemjy, line 1: Sorry I should have spotted this earlier, these constants should be in [this file](https://github.dev/VenusProtocol/venus-protocol-interface/apps/evm/src/constants/production.ts) in my opinion (just another file containing constants for production URLs).

apps/evm/src/clients/api/queries/getProposalsCountFromApi/useGetProposalsCountFromApi.ts
```diff
@@ -0,0 +1,8 @@
+import { useQuery } from '@tanstack/react-query';
+import { fetchProposalCount } from '.';
+
+export const useProposalsCountFromApi = () =>
+  useQuery({
+    queryKey: ['proposalCount'],
```
therealemjy: The keys used in queries are centralized in [this file](https://github.dev/VenusProtocol/venus-protocol-interface/apps/evm/src/constants/functionKey.ts), so you'll have to add a key to it then reference it in this query.

The query keys are centralized in a shared file so that it makes it easy to manipulate React Query's cache when making mutations.

apps/evm/src/containers/Container/index.tsx
therealemjy, line 1: The `containers` directory is for components with side effect (you could not know that), so this component shouldn't be here.

This component should be removed anyway and placed in the Landing page directory since that's the only place where it's used.

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -524,6 +524,45 @@
...
+    "governance": {
+      "forum": "Governance forum",
+      "proposalCounting": "Venus Improvement Proposals Executed and counting",
+      "text": "Venus is a community-driven decentralized protocol delivering best-in-class functionality for crypto money markets.",
+      "title": "Money Markets governed <br />by community"
+    },
+    "protection": {
+      "challengeText": "We encourage all to challenge our code and search for vulnerabilities. Read about our bug bounty rewards, and please submit any bug you identify.",
+      "challengeTitle": "Challenge our code and be rewarded",
+      "protectionText": "Maintaining a fallback pool to keep us all safe in the case of outlier events",
```
```suggestion
      "protectionText": "Maintaining a fallback pool to keep us all safe in the case of outlier events.",
```

apps/evm/src/pages/Landing/Protection/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+import bugBountyImg from './assets/bugBounty.png';
+import protectionImg from './assets/protection.png';
+
+interface IProtectionProps {
+  className?: string;
+}
+
+const itemClassName =
+  'relative flex flex-1 flex-col min-h-105 justify-between p-6 pb-0 border border-solid border-lightGrey rounded-3xl sm:min-h-100 xl:min-h-139 xl:p-10 xl:pb-0';
+const bgImgClassName = 'w-full bottom-0';
```
therealemjy: I recommend wrapping these with `cn()`, so that you get autocompletion for Tailwind classes in the IDE. If you do that, please update the other components too (I won't comment on all of them not to spam).

```suggestion
const itemClassName =
  cn('relative flex flex-1 flex-col min-h-105 justify-between p-6 pb-0 border border-solid border-lightGrey rounded-3xl sm:min-h-100 xl:min-h-139 xl:p-10 xl:pb-0');
const bgImgClassName = cn('w-full bottom-0');
```

apps/evm/src/pages/Landing/Protection/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+            itemClassName,
+            'bg-radial-[62.14%_57.90%_at_50.00%_50.00%,rgba(23,46,98,0.50)_0%,rgba(18,22,32,0.50)_100%]',
+          )}
+          key="bounty"
+        >
+          <div className={'xl:mb-21'}>
+            <h2 className="mb-4">{t('landing.protection.challengeTitle')}</h2>
+            <p className="mb-8 md:mb-0">{t('landing.protection.challengeText')}</p>
+          </div>
+          <img loading="lazy" className={cn(bgImgClassName)} src={bugBountyImg} alt="bugBounty" />
```
therealemjy: No need to wrap the variable with `cn()` since you aren't appending any class to it. Same thing here, I won't comment on all components but please scan through them.


```suggestion
          <img loading="lazy" className={bgImgClassName} src={bugBountyImg} alt="bugBounty" />
```

apps/evm/src/pages/Landing/Governance/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { t, Trans } = useTranslation();
+  const { data: proposalsCount, isLoading } = useProposalsCountFromApi();
+
+  return !isLoading && proposalsCount ? (
+    <Container className={cn('mt-15 md:mt-20 xl:mt-25', textClassName)}>
+      <div
+        className={
+          'flex flex-col-reverse justify-between overflow-y-hidden border border-solid border-lightGrey rounded-3xl h-135.5 bg-[#1E2431] sm:h-94 sm:flex-row xl:gap-6 xl:h-125'
+        }
+        key="bounty"
```
therealemjy: This shouldn't be necessary

apps/evm/src/pages/Landing/Protection/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <Container className={cn('mt-15 md:mt-20 lg:mt-25', textClassName, className)}>
+      <ul className={'flex flex-col justify-between gap-6 sm:flex-row xl:gap-8'}>
+        <li
+          className={cn(
+            itemClassName,
+            'bg-radial-[62.14%_57.90%_at_50.00%_50.00%,rgba(23,46,98,0.50)_0%,rgba(18,22,32,0.50)_100%]',
+          )}
+          key="bounty"
```
therealemjy: This shouldn't be necessary

apps/evm/src/pages/Landing/Protection/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+            <p className="mb-8 md:mb-0">{t('landing.protection.challengeText')}</p>
+          </div>
+          <img loading="lazy" className={cn(bgImgClassName)} src={bugBountyImg} alt="bugBounty" />
+        </li>
+        <li
+          className={cn(
+            itemClassName,
+            'bg-radial-[62.14%_57.90%_at_50.00%_50.00%,rgba(30,75,100,0.50)_0%,rgba(18,22,32,0.50)_100%]',
+          )}
+          key="protection"
```
therealemjy: This shouldn't be necessary

therealemjy (commented):
This looks good.

I noticed a few UI glitches, but nothing crazy:

1) All the content below the Prime section has additional left and right padding
Image

Expected:
Image

2) The "_14 audits..." text isn't centrally aligned and has no border
Image

Expected:
Image

3) The text below the "Safety before all" text should be grey

Image

Expected:
Image

apps/evm/src/libs/translations/translations/en.json
david-sun-venus: I noticed the missing period when I migrated, but the original one doesn't have it, so I kept it the same.

apps/evm/src/constants/functionKey.ts
```diff
@@ -70,6 +70,7 @@ enum FunctionKey {
   GET_ACCOUNT_TRANSACTION_HISTORY = 'GET_ACCOUNT_TRANSACTION_HISTORY',
   GET_SIMULATED_POOL = 'GET_SIMULATED_POOL',
   GET_SWAP_QUOTE = 'GET_SWAP_QUOTE',
+  GET_PROPOSALS_COUNT_FROM_API = 'GET_PROPOSALS_COUNT_FROM_API',
```
therealemjy: `GET_PROPOSALS_COUNT` (no need to specify it comes from the API).

apps/evm/src/pages/Landing/index.tsx
therealemjy: I think this should be moved into the [dApp's root](https://github.com/VenusProtocol/venus-protocol-interface/blob/main/apps/evm/src/App/Routes/index.tsx), so that it is triggered when visiting https://app.venus.io/discord (it redirects to our Discord server).

apps/evm/src/pages/Landing/Wallets/index.tsx
```diff
@@ -74,7 +73,7 @@ export const Wallets: React.FC = () => {
 
   return (
     <section className="my-15 md:my-20 xl:my-25">
-      <Container>
+      <div>
```
therealemjy: No need for this unstyled `div` tag then.

apps/evm/src/pages/Landing/Protection/index.tsx
```diff
@@ -34,20 +34,29 @@ export const Protection: React.FC<IProtectionProps> = ({ className }) => {
...
           <div className={'xl:mb-21'}>
             <h2 className="mb-4">{t('landing.protection.protectionTitle')}</h2>
             <p className="mb-8 md:mb-0">{t('landing.protection.protectionText')}</p>
           </div>
-          <img loading="lazy" className={bgImgClassName} src={protectionImg} alt="protection" />
+          <img
+            loading="lazy"
+            className={bgImgClassName}
+            src={protectionImg}
+            alt={t('landing.protection.protectionAlt', 'Protection Illustration')}
```
therealemjy: You don't need to set a default string, just the i18n key is enough since we then use a parser to gather all the translations into files anyway. This applies to all translations.

apps/evm/src/libs/translations/translations/en.json
therealemjy: Let's add the dot then so we're consistent

therealemjy (approved):
Nice

PR #5218, opened 2026-01-07
feat: start updating layout

therealemjy:
## Jira ticket(s)

VPD-414

- rename Vaults page to Staking
- update layout. In this version, only the mobile layout is supported. It is sufficient to navigate throughout the app while we're making updates, so I figured I'd stop the PR here to keep it manageable. I will continue to work on the layout in the following PRs to add support for other resolutions


david-sun-venus (approved):
lgtm

PR #5226, opened 2026-01-12
feat: add support for multiple languages

therealemjy:
## Jira ticket(s)

VPD-423

- add "language" setting
- add support for the following languages: Japanase, Thai, Turkish, Vietnamese, Chinese (simplified) and Chinese (traditional)


apps/evm/src/libs/translations/constants/index.ts
```diff
@@ -0,0 +1,17 @@
+export interface SupportedLanguage {
+  name: string;
+  tag: string;
```
david-sun-venus: I think "code" is a better term than "tag" in this context.

therealemjy: The name "tag" comes from "BCP 47 language tags" (https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag), which is the format used for those.
Maybe I can make this clearer by naming the property `bcp47Tag`?

david-sun-venus: I see. It's my first time knowing about the BCP_47 tag and I think bcp47Tag is clearer for sure. 

Another thing I'd like to address is the code for Simplified/Traditional Chinese. I believe the following statement from [here](https://en.wikipedia.org/wiki/IETF_language_tag#List_of_common_primary_language_subtags) makes sense in our case (instead of having zh-CN and zh-TW, it's more appropriate to have zh-hans and zh-hant). 
Image

It might also depends on if you want language tag displayed in the url.

therealemjy: Same to be honest. I'll make the property change.

Works for me regarding the Chinese tags, thanks for researching this. I'll also make this change.

david-sun-venus (approved):
lgtm

PR #5232, opened 2026-01-13
feat: add temporary fix to handle VenusLens address change

therealemjy:
- add temporary fix to handle VenusLens address change after the Fermi upgrade


david-sun-venus (approved):
lgtm

PR #5236, opened 2026-01-14
feat: add chainId property to tokens

therealemjy:
## Changes

This PR is scary, but the high number of line changes is simply due to snapshot updates. The logics to check are the following ones:
- packages/chains/src/types/index.ts: a `chainId` property was added to the `Token` type, which is the source of all the other changes
- packages/chains/src/tokens/underlyingTokens: each token in the lists has been appended with a `chainId` property
- packages/chains/src/tokens/nativeTokens/index.ts: native tokens are now generated as mappings, then referenced using a chain ID (e.g.: bnb[ChainId.BSC_MAINNET])


david-sun-venus (approved):
lgtm

PR #5243, opened 2026-01-15
feat: add ad carousel and top markets to Dashboard page

therealemjy:
## Jira ticket(s)

VPD-415

- remove current Dashboard page
- add ad carousel and top markets to Dashboard page
- add `AssetCard` component


apps/evm/src/components/TokenIconWithSymbol/index.tsx
```diff
@@ -1,22 +1,35 @@
...
+  return (
+    <div className={cn(className, 'flex items-center gap-x-2')}>
+      <TokenIcon
+        token={token}
+        className={tokenIconClassName}
+        displayChain={displayChain}
+        {...otherProps}
+      />
+
+      <div className="space-y-1">
```
david-sun-venus:
```suggestion
      <div>
```

apps/evm/src/components/TokenIconWithSymbol/index.tsx
```diff
@@ -1,22 +1,35 @@
...
+        <p className={cn(displayChain && 'text-sm leading-[1.2]')}>{token.symbol}</p>
+
+        {displayChain && <p className="text-light-grey text-xs leading-[1.2]">{chain.name}</p>}
```
david-sun-venus:
```suggestion
        {displayChain && <p className="text-light-grey text-xs leading-[1.2] mt-1">{chain.name}</p>}
```

therealemjy: The `space-y-1` class does the same thing

apps/evm/src/components/TokenIconWithSymbol/index.tsx
therealemjy: This class is necessary to apply spacing between children.

apps/evm/src/components/AssetCard/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+  className?: string;
+}
+
+export const AssetCard: React.FC<AssetCardProps> = ({
+  asset,
+  poolComptrollerContractAddress,
+  type,
+  className,
+}) => {
+  const to = routes.market.path
```
david-sun-venus, line 22: I've seen this piece of code in several places. Maybe it might be better to extract it to a hook, and instead of replacing hardcoded  ':poolComptrollerAddress' and ':vTokenAddress' every time, maybe it can get the poolComptrollerContractAddress from hook itself and we don't need to pass it down?

apps/evm/src/pages/Dashboard/AdCarousel/index.tsx
```diff
@@ -0,0 +1,33 @@
+import { Carousel, CarouselItem, Icon } from 'components';
+import { PrimeBanner } from './PrimeBanner';
+
+const slides: React.ReactNode[] = [<PrimeBanner />];
+
+export const AdCarousel: React.FC = () => {
+  const closeCarousel = () => {}; // TODO: wire up
```
david-sun-venus, line 7: Just checking if this TODO is intended for later implementations.

apps/evm/src/pages/Dashboard/index.tsx
```diff
@@ -1,178 +1,19 @@
...
+        </div>
 
-        <Summary
-          className="lg:basis-4/12"
-          pools={pools}
-          vaults={vaults}
-          xvsPriceCents={xvsPriceCents}
-          vaiPriceCents={vaiPriceCents}
-        />
+        <div className="xl:grow xl:order-1">TODO: add content</div>
```
david-sun-venus, line 15: Contents?

therealemjy: Yeah, this will be added in the following PRs

apps/evm/src/pages/Dashboard/AdCarousel/index.tsx
therealemjy: I ended up implementing this in a following PR :)

apps/evm/src/components/AssetCard/index.tsx
therealemjy: We can't do that actually, because the `poolComptrollerContractAddress` parameter is specific to the pool. Currently, we support isolated pools which each use a different `poolComptrollerContractAddress` value, so we need to keep this component modular.

david-sun-venus (approved):

PR #5245, opened 2026-01-15
feat: add overview section to Dashboard page

therealemjy:
## Jira ticket(s)

VPD-415

- add account overview section to Dashboard page


apps/evm/src/pages/Dashboard/Overview/index.tsx
```diff
@@ -0,0 +1,370 @@
...
+export const Overview: React.FC<OverviewProps> = ({ className, ...otherProps }) => {
+  const { t } = useTranslation();
+
+  const isVaiFeatureEnabled = useIsFeatureEnabled({
+    name: 'vaiRoute',
+  });
+
+  const [userChainSettings] = useUserChainSettings();
+
+  console.log(store.use.setUserSettings());
```
david-sun-venus: remove console

therealemjy: Good catch, will do now

apps/evm/src/pages/Dashboard/Overview/index.tsx
```diff
@@ -0,0 +1,368 @@
...
+      : undefined;
+
+  const dailyChangePercentage =
+    dailyChangeCents !== undefined && netWorthCents !== undefined && netWorthCents !== 0
+      ? (dailyChangeCents * 100) / netWorthCents
+      : undefined;
+
+  let readableDailyChangePercentage = formatPercentageToReadableValue(dailyChangePercentage);
+  // Remove "-" sign
+  if (readableDailyChangePercentage[0] === '-') {
```
david-sun-venus, line 178: Why not just using `Math.abs()` in line 173?

therealemjy: Because `readableDailyChangePercentage` is a string that contains a `%` sign at the end.

apps/evm/src/containers/HidableUserBalance/index.tsx
```diff
@@ -0,0 +1,13 @@
+import { HIDDEN_BALANCE_KEY } from 'constants/placeholders';
+import { useUserChainSettings } from 'hooks/useUserChainSettings';
+
+export interface HidableUserBalanceProps {
+  children: React.ReactNode;
```
david-sun-venus, line 5: I believe React.FC already includes the `children` in its type, so no need to declare it specifically.

therealemjy: It does not unfortunately

david-sun-venus: I see. It has been changed since react 18. Now we could do something like this instead: `React.FC<PropsWithChildren>`

therealemjy: Yeah I believe that's what I'm doing here

david-sun-venus: I mean we could reuse `PropsWithChildren` type provided by React lib.

therealemjy: Not in this case. `PropsWithChildren` sets an optional `children` property, in this case I want to be required.

david-sun-venus: I see.

david-sun-venus (approved):

PR #5264, opened 2026-01-20
feat: list PT-clisBNBx-25JUN2026 token

therealemjy:
## Jira ticket(s)

VPD-473

- list PT-clisBNBx-25JUN2026 token


packages/chains/src/scripts/generateVTokens/writeVTokens/index.ts
```diff
@@ -79,6 +79,8 @@ export const writeVTokens = async ({
     }
 
     if (underlyingTokenIndex < 0) {
+      console.log(underlyingAddress);
```
gleiser-oliveira: Log 👀

therealemjy: Good catch, thanks

gleiser-oliveira (approved):

PR #5271, opened 2026-01-21
chore: do not block build pipeline when vToken is missing

therealemjy:
- emit warning instead of throwing error when a VToken record is missing


david-sun-venus (approved):
lgtm

PR #5439, opened 2026-03-04
fix: responsive issue with max buttons

therealemjy:
## Changes

- fix responsive issue with max buttons


therealemjy:
@greptile

PR #5445, opened 2026-03-09
feat(evm): vault page revamp

david-sun-venus:
## Jira ticket(s)

VEN-730

## Changes

- Vault Page revamp


therealemjy:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -1800,21 +1804,58 @@
...
+      "vaults": "Vaults",
+      "yieldTokens": "Yield tokens"
+    },
+    "overview": {
+      "bannerCoinIllustration": "{{token}} coins illustration",
+      "bannerTitle": "Enjoy <strong>{{apyPercentage}}</strong> APR with <strong>{{tokenDashMarket}}</strong>",
+      "bannerVaultIllustration": "Vault illustration",
+      "description": "Yield made simple, returns made real — deposit across multiple vaults and watch your crypto grow <strong>on your terms</strong>.",
+      "highestApr": "Highest APR",
+      "title": "Venus Vault",
```
```suggestion
      "title": "Vaults",
```

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -1800,21 +1804,58 @@
...
+    },
+    "filter": {
+      "active": "Active",
+      "allCategories": "All categories",
+      "allManagers": "All managers",
+      "allStates": "All states",
+      "claim": "Claim",
+      "deposit": "Deposit",
+      "earning": "Earning",
+      "inputPlaceholder": "Search asset or market",
```
therealemjy: "Search asset"

apps/evm/src/pages/Vaults/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { type ActiveModal, VaultModals } from 'containers/Vault';
+import { useState } from 'react';
+import { Overview } from './Overview';
+import { Vaults } from './Vaults';
+
+const StakingPage: React.FC = () => {
+  const [activeModal, setActiveModal] = useState<ActiveModal | undefined>(undefined);
+  const [activeVault, setActiveVault] = useState<Vault | undefined>(undefined);
+
+  const openModal = (_vault: Vault, _activeModal?: ActiveModal) => {
```
therealemjy:
```suggestion
  const openModal = (vault: Vault, activeModal?: ActiveModal) => {
```

Prefixing a variable with `_` indicates it is unused, but it's used here.

apps/evm/src/containers/Vault/VaultCard/Legacy/index.tsx
therealemjy, line 1: This needs to be an index file inside a directory with the name `Legacy`

apps/evm/src/containers/Vault/VaultCard/Simplified/index.tsx
therealemjy, line 1: This needs to be an index file inside a directory with the name `Simplified`

apps/evm/src/clients/api/queries/getTokenUsdPrice/useGetTokenListUsdPrice.ts
```diff
@@ -0,0 +1,60 @@
...
+    queries: (tokens ?? []).map(token => ({
+      queryKey: [FunctionKey.GET_TOKEN_USD_PRICE, { tokenAddress: token?.address ?? '', chainId }],
+      queryFn: () =>
+        callOrThrow({ token, resilientOracleAddress }, params =>
+          getTokenUsdPrice({ publicClient, ...params }),
+        ),
+      enabled:
+        (options?.enabled === undefined || options?.enabled) && !!token && !!resilientOracleAddress,
+    })),
+  });
```
therealemjy: You can make this more efficient by using a single multical; to get the price of all the tokens in one call.

apps/evm/src/clients/api/queries/getTokenUsdPrice/useGetTokenListUsdPrice.ts
therealemjy, line 1: This needs to be in a seperate directory inside the query client

apps/evm/src/containers/Vault/hooks/useVaultUsdValues.ts
```diff
@@ -0,0 +1,50 @@
...
+    dailyEmissionMantissa,
+  } = vault;
+
+  const { data: stakedTokenPrice, isLoading: isStakedTokenPriceLoading } = useGetTokenUsdPrice({
+    token: stakedToken,
+  });
+
+  const { data: rewardTokenPrice, isLoading: isRewardTokenPriceLoading } = useGetTokenUsdPrice({
+    token: rewardToken,
+  });
```
therealemjy: You've created a `useGetTokenListUsdPrice` hook to fetch multiple prices in one query, you should use it here.

apps/evm/src/pages/Vaults/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <Page>
+      <div className={cn('flex flex-col gap-6 sm:gap-12')}>
+        <Overview vaults={vaults} onOpenModal={openModal} />
+
+        <Vaults vaults={vaults} openModal={openModal} />
+      </div>
+      {activeVault && (
+        <VaultModals vault={activeVault} activeModal={activeModal} onClose={closeModal} />
+      )}
```
therealemjy: Why not keep following the current logic, which is to render the modals linked to a certain type of vaults from the vault itself?

You then wouldn't have to store the vault in a state like you're here, which can lead to issues down the line (for example if the data changes while the user is on the form, they won't get the update).

apps/evm/src/pages/Vaults/Vaults/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+          /> */}
+        </div>
+        <TextField
+          value={search}
+          onChange={onChange}
+          size="sm"
+          leftIconSrc="search"
+          placeholder={t('vault.filter.inputPlaceholder')}
+          className="w-full lg:w-80"
+        />
```
therealemjy, line 76: According to the designs, this form should use a 14px normal font face (it currently uses 16px bold font face).

apps/evm/src/pages/Vaults/Vaults/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+            size="medium"
+            variant="tertiary"
+            placeLabelToLeft
+            options={curatorOptions}
+            optionClassName="px-3 h-10 scrollbar-track-cards"
+            dropdownClassName="overflow-y-auto max-h-70 scrollbar-thin scrollbar-track-cards scrollbar-thumb-grey"
+            buttonClassName="sm:min-w-45"
+            value={curator}
+            onChange={newValue => setCurator(newValue.toString())}
+          />
```
therealemjy, line 67: This can be made into a reusable component within the Vaults page (so that you don't duplicate `optionClassName`, `dropdownClassName` and all other duplicated props).

apps/evm/src/pages/Vaults/Vaults/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+    // statusOptions,
+  } = useFilterOptions();
+
+  const [search, setSearch] = useState('');
+  const onChange: TextFieldProps['onChange'] = e => {
+    setSearch(e.currentTarget?.value);
+  };
+
+  const filteredVaults = (vaults ?? []).filter(vault => {
+    const { category: _category, curator: _curator, status: _status } = getVaultMetadata(vault);
```
therealemjy:
```suggestion
    const { category, curator, status } = getVaultMetadata(vault);
```

apps/evm/src/components/Select/index.tsx
```diff
@@ -67,11 +68,12 @@ export const Select = forwardRef<HTMLInputElement, SelectProps>(
             'border-transparent bg-lightGrey rounded-xl hover:bg-lightGrey hover:border-grey active:bg-lightGrey active:border-blue',
             isDropdownOpened && 'border-blue bg-lightGrey hover:border-blue',
           );
+        */
         // primary
         default:
           return cn(
-            'border-dark-blue-disabled/50 bg-dark-blue hover:bg-dark-blue-hover active:bg-dark-blue-hover',
-            isDropdownOpened && 'bg-dark-blue-hover hover:border-blue',
+            'rounded-xl border-dark-blue-hover bg-transparent hover:bg-transparent hover:border-light-grey active:bg-dark-blue',
```
therealemjy:
```suggestion
            'rounded-lg border-dark-blue-hover bg-transparent hover:bg-transparent hover:border-light-grey active:bg-dark-blue',
```

apps/evm/src/components/Icon/icons/search.tsx
therealemjy, line 1: We already have an icon for this: `magnifier`

apps/evm/src/pages/Vaults/Vaults/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+            buttonClassName="sm:min-w-45"
+            value={curator}
+            onChange={newValue => setCurator(newValue.toString())}
+          />
+        </div>
+        <TextField
+          value={search}
+          onChange={onChange}
+          size="sm"
+          leftIconSrc="search"
```
therealemjy:
```suggestion
          leftIconSrc="magnifier"
```

apps/evm/src/clients/api/queries/getTokenUsdPrice/useGetTokenListUsdPrice.ts
```diff
@@ -0,0 +1,60 @@
...
+
+type TrimmedGetTokenUsdPriceInput = Omit<
+  GetTokenUsdPriceInput,
+  'publicClient' | 'resilientOracleAddress'
+>;
+
+export type UseGetTokenUsdPriceQueryKey = [
+  FunctionKey.GET_TOKEN_USD_PRICE,
+  {
+    tokenAddress: string;
```
therealemjy:
```suggestion
    tokenAddress: Address;
```

apps/evm/src/clients/api/queries/getTokenUsdPrice/useGetTokenListUsdPrice.ts
```diff
@@ -0,0 +1,60 @@
...
+type Options = QueryObserverOptions<
+  GetTokenUsdPriceOutput,
+  Error,
+  GetTokenUsdPriceOutput,
+  GetTokenUsdPriceOutput,
+  UseGetTokenUsdPriceQueryKey
+>;
+
+export interface UseGetTokenUsdPriceInput extends Omit<TrimmedGetTokenUsdPriceInput, 'token'> {
+  tokens?: (Token | undefined)[];
```
therealemjy: This should be stricter, so that you don't need to check for `undefined` values inside the hook itself:
```suggestion
  tokens: Token[];
```

apps/evm/src/pages/Dashboard/Vaults/index.tsx
```diff
@@ -14,26 +19,89 @@ export const Vaults: React.FC<VaultsProps> = ({ vaults }) => {
   // Filter out vaults user has not staked in
   const filteredVaults = vaults.filter(vault => vault.userStakedMantissa?.isGreaterThan(0));
 
+  const stakedTokenPriceResults = useGetTokenListUsdPrice({
+    tokens: filteredVaults.map(vault => vault.stakedToken),
+  });
+  const rewardTokenPriceResults = useGetTokenListUsdPrice({
+    tokens: filteredVaults.map(vault => vault.rewardToken),
+  });
```
therealemjy: You can merge this into one fetch request only.

apps/evm/src/containers/Vault/VaultCard/Simplified/index.tsx
therealemjy, line 1: There's duplicated UI code here: use reusable components within your code as much as possible. If a UI  is duplicated, it should be made into a component. This limits the amount of code and makes it easier to maintain.

apps/evm/src/libs/translations/translations/en.json
david-sun-venus: Title is "Venus Vault" from the design:
image

apps/evm/src/libs/translations/translations/en.json
david-sun-venus: It can also search for markets, for example "venus". Seems to be what design intends.

apps/evm/src/pages/Vaults/index.tsx
david-sun-venus: image

It could also mean inner scoped variables. Changing them to the same name as outer scope will be shadowing which is confusing to read and higher risks of introducing bugs when modifying later on.

apps/evm/src/components/Icon/icons/search.tsx
david-sun-venus: I see. I tried to check it from storybook, but probably missed it. Didn't expect it to be called this.

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

apps/evm/src/clients/api/queries/getTokenListUsdPrice/index.ts
```diff
@@ -0,0 +1,47 @@
...
+        tokenPriceUsd: undefined,
+        error: new VError({ type: 'unexpected', code: 'somethingWentWrong' }),
+      };
+
+    return {
+      tokenPriceUsd: convertPriceMantissaToDollars({
+        priceMantissa: new BigNumber(resp.result?.toString() ?? 0),
+        decimals: tokens[index].decimals,
+      }),
+    };
```
therealemjy: We don't return errors if one happens, we log it and either we throw it back if it's crucial, or we skip the result if it's skippable.

You can find an example here: https://github.com/VenusProtocol/venus-protocol-interface/blob/65764fec8e6171ae501f4dcef359029b3d76dd24/apps/evm/src/clients/api/queries/getPendingRewards/index.ts#L181

apps/evm/src/clients/api/queries/getTokenListUsdPrice/useGetTokenListUsdPrice.ts
```diff
@@ -0,0 +1,65 @@
...
+  GetTokenListUsdPriceOutput,
+  Error,
+  GetTokenListUsdPriceOutput,
+  GetTokenListUsdPriceOutput,
+  UseGetTokenListUsdPriceQueryKey
+>;
+
+export interface UseGetTokenListUsdPriceInput
+  extends Omit<TrimmedGetTokenListUsdPriceInput, 'tokens'> {
+  tokens?: Token[];
```
therealemjy:
```suggestion
  tokens: Token[];
```

apps/evm/src/libs/translations/translations/en.json
therealemjy: Yeah it's not your responsibility for sure, but it doesn't make sense to use this title if the page is called "Vaults".
@zed-venus I believe the title of the page should match its name in the navigation menu, so it would be "Vaults". What do you think?

apps/evm/src/libs/translations/translations/en.json
therealemjy: Yeah the designs are generally quite inaccurate when it comes to wordings. I did ask whether this content gets proofread but I'm not sure it really does.

An asset and a market are the same thing, hence why it should be "Search asset" (to be consistent with other search inputs in the app).

apps/evm/src/pages/Vaults/index.tsx
therealemjy: Works for me 👍

apps/evm/src/clients/api/queries/getTokenUsdPrice/useGetTokenUsdPrice.ts
```diff
@@ -51,7 +52,7 @@ export const useGetTokenUsdPrice = (
     queryKey: [
       FunctionKey.GET_TOKEN_USD_PRICE,
       {
-        tokenAddress: token ? token.address : '',
+        tokenAddress: token?.address ?? '0x',
```
therealemjy: Small thing: in general, we use the `NULL_ADDRESS` (you can import it from constants) when we want to put an address placeholder.

```suggestion
        tokenAddress: token?.address ?? NULL_ADDRESS,
```

apps/evm/src/containers/Vault/VaultCard/Simplified/index.tsx
```diff
@@ -110,10 +99,12 @@ export const VaultCardSimplified: React.FC<VaultCardSimplifiedProps> = ({
                     token: vault.stakedToken,
                   })}
                 </div>
-              )}
-            </>
-          )}
-        </div>
+              ) : (
```
therealemjy: Avoid using nested ternary statements, this tends to make the code harder to read.

apps/evm/src/pages/Dashboard/Vaults/index.tsx
```diff
@@ -1,6 +1,7 @@
 import BigNumber from 'bignumber.js';
-import { useGetTokenListUsdPrice } from 'clients/api/queries/getTokenUsdPrice/useGetTokenListUsdPrice';
+import { useGetTokenListUsdPrice } from 'clients/api/queries/getTokenListUsdPrice/useGetTokenListUsdPrice';
```
therealemjy: Add `getTokenListUsdPrice` and `useGetTokenListUsdPrice` to the barrel file of the API client ([here](https://github.com/VenusProtocol/venus-protocol-interface/blob/65764fec8e6171ae501f4dcef359029b3d76dd24/apps/evm/src/clients/api/index.ts)) then import it from there:
```suggestion
import { useGetTokenListUsdPrice } from 'clients/api;
```

apps/evm/src/pages/Vaults/Overview/index.tsx
```diff
@@ -1,7 +1,7 @@
 import BigNumber from 'bignumber.js';
 
 import { cn } from '@venusprotocol/ui';
-import { useGetTokenListUsdPrice } from 'clients/api/queries/getTokenUsdPrice/useGetTokenListUsdPrice';
+import { useGetTokenListUsdPrice } from 'clients/api/queries/getTokenListUsdPrice/useGetTokenListUsdPrice';
```
therealemjy:
```suggestion
import { useGetTokenListUsdPrice } from 'clients/api';
```

apps/evm/src/containers/Vault/VaultCard/Simplified/index.tsx
david-sun-venus: It's the ternary statements for content, should not be considered as nested imo.

apps/evm/src/pages/Vaults/index.tsx
david-sun-venus: Previously Modals are mounted inside for each Card which I think would not be as optimized in terms of performance once we have more and more cards. The data change issue for current modals should be fine, but it could still be an potential issue in some cases. I will definitely improve the design along with the new modals.

david-sun-venus:
@greptile

apps/evm/src/pages/Vaults/index.tsx
therealemjy: There's no impact on performances because the modals are conditionally rendered. So although each vault declares its own modals, they aren't in the dom unless the "open" state is true.

The data change issue is not fine either as of could make transactions fail, for no gain. Please stick to the original implementation.

apps/evm/src/containers/Vault/VaultCard/Simplified/index.tsx
therealemjy: I disagree, it makes it harder to read the code, for something you can easily avoid.

david-sun-venus:
@greptile

therealemjy (approved):

therealemjy (approved):

PR #5448, opened 2026-03-11
feat: configure ui-develop skill

cuzz-venus:
## Jira ticket(s)

VPD-722

## Changes
add claude config
-


.claude/skills/ui-i18n/SKILL.md
david-sun-venus, line 1: For i18n, there's an existing script under apps/evm: `yarn i18next`. If you run it, it will gather all keys used in codebase and generate updated json files, sorted and also removing unused. I think it's easier to let AI run this cmd first and then try to translate what needs to be translated.

cuzz-venus: good idea, let me see the behavior by running i18next script.

cuzz-venus:
@greptile

therealemjy (approved):

PR #5449, opened 2026-03-11
fix: display Dashboard stats with no borrow position

therealemjy:
## Changes

- display Dashboard stats with no borrow position


cuzz-venus (approved):
/lgtm

PR #5459, opened 2026-03-16
feat(evm): pendle vault

david-sun-venus:
## Jira ticket(s)

VEN-736

## Changes

- Pendle Vault Card & Flow

## Summary by CodeRabbit

* **New Features**
  * Full Pendle PT vault flow: deposit/withdraw/redeem with integrated swap quotes, conversion details, approvals, analytics, and a modal with Position & Overview tabs
  * Enhanced vault UI: updated vault cards, status badges, strategy diagram, flow arrows/nodes, supply/total-deposits chart, submit flow, token icon with remaining-period, and market info

* **Chores**
  * Added vault product APIs, contract registration, expanded vault type support, manager/status filters, and extensive localized strings for vault modals and UI


david-sun-venus:
@greptile

david-sun-venus:
@greptile

apps/evm/src/clients/api/mutations/usePendlePtVault/index.ts
```diff
@@ -0,0 +1,180 @@
...
+      }
+
+      throw new VError({
+        type: 'unexpected',
+        code: 'incorrectSwapInput',
+      });
+    },
+    onConfirmed: async ({ input }) => {
+      captureAnalyticEvent(`Pendle vault ${input.type}`, {
+        pendleMarketAddress: pendleMarketAddress,
```
therealemjy:
```suggestion
        pendleMarketAddress,
```

apps/evm/src/clients/api/mutations/usePendlePtVault/index.ts
```diff
@@ -0,0 +1,180 @@
...
+          abi: pendlePtVaultAbi,
+          address: pendlePtVaultContractAddress,
+          functionName: 'redeemAtMaturity' as const,
+          args: formatRedeemParams(swapQuote.contractCallParams, amountToken),
+        } as const;
+      }
+
+      throw new VError({
+        type: 'unexpected',
+        code: 'incorrectSwapInput',
```
therealemjy: This error code does not exist in the codebase.

```suggestion
        code: 'somethingWentWrong',
```

apps/evm/src/clients/api/mutations/usePendlePtVault/utils/index.ts
therealemjy, line 1: Only one helper per file

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

david-sun-venus:
@greptile

apps/evm/src/clients/api/mutations/usePendlePtVault/index.ts
```diff
@@ -0,0 +1,174 @@
...
+        return {
+          abi: pendlePtVaultAbi,
+          address: pendlePtVaultContractAddress,
+          functionName: 'redeemAtMaturity' as const,
+          args: formatWithdrawParams(swapQuote.contractCallParams, { fromToken, vToken }), // Share the same format as withdraw.
+        } as const;
+      }
+
+      throw new VError({
+        type: 'unexpected',
```
therealemjy: This hook should be separated into multiple ones. Each hook represents an action, so in this case what seems the most accurate is to create 3 hooks to split each action: deposit (which will support native and non-native tokens, withdraw, redeemAtMaturity.

apps/evm/src/clients/api/queries/getPendleSwapQuote/types.ts
```diff
@@ -0,0 +1,109 @@
...
+  fromToken: Token;
+  toToken: Token;
+  amount: BigNumber;
+  slippagePercentage: number;
+  receiverAddress?: Address;
+}
+
+export type GetPendleSwapQuoteOutput = {
+  estReceiveMantissa: BigNumber;
+  feeUsdCents: BigNumber;
```
therealemjy:
```suggestion
  estimatedReceivedTokensMantissa: BigNumber;
  feeCents: BigNumber;
```

apps/evm/src/clients/api/queries/getPendleSwapQuote/types.ts
```diff
@@ -0,0 +1,109 @@
...
+  }[];
+}
+
+export type PendleSwapQuoteError = VError<'pendleSwapQuote' | 'interaction'>;
+
+export interface GetPendleSwapQuoteInput {
+  chainId: ChainId;
+  fromToken: Token;
+  toToken: Token;
+  amount: BigNumber;
```
therealemjy:
```suggestion
  amountTokens: BigNumber;
```

apps/evm/src/clients/api/queries/getPendleSwapQuote/formatOutput.ts
```diff
@@ -0,0 +1,13 @@
+import BigNumber from 'bignumber.js';
+import _ from 'lodash';
```
therealemjy: This will import the entire `lodash` library. Import the `pick` function individually instead:
```suggestion
import _pick from 'lodash/pick';
```

apps/evm/src/clients/api/mutations/usePendlePtVault/utils.ts
```diff
@@ -0,0 +1,82 @@
+import BigNumber from 'bignumber.js';
+import type { PendleContractCallParams } from 'clients/api/queries/getPendleSwapQuote';
+import _ from 'lodash';
```
therealemjy: Import `lodash` functions individually, otherwise you'll import the whole library.

apps/evm/src/clients/api/mutations/usePendlePtVault/utils/index.ts
therealemjy, line 1: This feels more convoluted than it needs to be. We’re passing raw API contractCallParams through the query layer, then reinterpreting tuple indexes in the mutation layer and mutating nested values via convertKeysToNumber.

A simpler and more codebase-consistent approach would be to normalize the Pendle response in getPendleSwapQuote/formatOutput and return an explicit typed shape there, similar to getSwapQuote/formatSwapQuote.

apps/evm/src/clients/api/queries/getPendleSwapQuote/useGetPendleSwapQuote.ts
```diff
@@ -0,0 +1,49 @@
...
+  const { chainId } = useChainId();
+  const { accountAddress } = useAccountAddress();
+
+  return useQuery({
+    queryKey: [FunctionKey.GET_PENDLE_SWAP_QUOTE, input, chainId, accountAddress ?? NULL_ADDRESS],
+    queryFn: () =>
+      callOrThrow({}, params =>
+        getPendleSwapQuote({ chainId, receiverAddress: accountAddress, ...input, ...params }),
+      ),
+    retry: false,
```
therealemjy, line 45: I don't think this should be here, we do want the request to be retried if it fails.

apps/evm/src/clients/api/queries/getPendleSwapQuote/useGetPendleSwapQuote.ts
```diff
@@ -0,0 +1,49 @@
...
+  input: UseGetPendleSwapQuoteInput,
+  options?: Partial<Options>,
+) => {
+  const { chainId } = useChainId();
+  const { accountAddress } = useAccountAddress();
+
+  return useQuery({
+    queryKey: [FunctionKey.GET_PENDLE_SWAP_QUOTE, input, chainId, accountAddress ?? NULL_ADDRESS],
+    queryFn: () =>
+      callOrThrow({}, params =>
```
therealemjy: The `callOrThrow` wrapper is unnecessary here since you pass it an empty object.

apps/evm/src/clients/api/queries/getPendleSwapQuote/useGetPendleSwapQuote.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const useGetPendleSwapQuote = (
+  input: UseGetPendleSwapQuoteInput,
+  options?: Partial<Options>,
+) => {
+  const { chainId } = useChainId();
+  const { accountAddress } = useAccountAddress();
+
+  return useQuery({
+    queryKey: [FunctionKey.GET_PENDLE_SWAP_QUOTE, input, chainId, accountAddress ?? NULL_ADDRESS],
```
therealemjy: To be consistent with the other query functions:
```suggestion
    queryKey: [FunctionKey.GET_PENDLE_SWAP_QUOTE, {
    	...input,
    	chainId,
    	accountAddress,
    }],
```

apps/evm/src/clients/api/queries/getPendleSwapQuote/useGetPendleSwapQuote.ts
```diff
@@ -0,0 +1,49 @@
...
+import { generatePseudoRandomRefetchInterval } from 'utilities/generatePseudoRandomRefetchInterval';
+import type { Address } from 'viem';
+import {
+  type GetPendleSwapQuoteInput,
+  type GetPendleSwapQuoteOutput,
+  type PendleSwapQuoteError,
+  getPendleSwapQuote,
+} from '.';
+
+export type UseGetPendleSwapQuoteInput = Omit<
```
therealemjy:
```suggestion
export type TrimmedGetPendleSwapQuoteInput = Omit<
```

apps/evm/src/clients/api/queries/getVaultProducts/index.ts
```diff
@@ -0,0 +1,35 @@
+import { VError } from 'libs/errors';
+import { restService } from 'utilities';
+import type {
+  GetVaultProductsInput,
+  GetVaultProductsOutput,
+  GetVaultProductsResponse,
+} from './types';
+
+export const getVaultProducts = async ({
```
therealemjy: Since we already have a `useGetVaults` query function, it feels odd to have a `getVaultProducts` query function (what's the difference between a vault and a vault product?).

Maybe this could be renamed to `getFixedRatedVaults`

apps/evm/src/clients/api/queries/useGetVaults/useGetPendleVaults/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+  pools: Pool[];
+  tokens: Token[];
+  now: number;
+}
+
+interface FormatToPendleVaultsInput extends BaseInput {
+  vaultProducts: GetVaultProductsOutput;
+}
+
+const DEPLOY_DATE_MAP: Record<Address, number> = {
```
therealemjy: This should come from the API, so that the frontend doesn't need to be redeployed every time a new vault is deployed.

apps/evm/src/clients/api/queries/getVaultProducts/useGetVaultProducts.ts
```diff
@@ -0,0 +1,28 @@
...
+>;
+
+export const useGetVaultProducts = (options?: Partial<Options>) => {
+  const { chainId } = useChainId();
+
+  return useQuery({
+    queryKey: [FunctionKey.GET_VAULT_PRODUCTS, chainId],
+    queryFn: () => getVaultProducts({ chainId }),
+    ...options,
+    enabled: options?.enabled === undefined || options?.enabled,
```
therealemjy: This line is unnecessary: the `enabled` prop is already spread out from the `options` object and you aren't adding more logic to it.

apps/evm/src/types/index.ts
```diff
@@ -464,19 +466,69 @@ export interface Transaction {
...
 }
 
+export interface Vault extends VaultData, VaultMetadata {}
+
+export interface PendleVaultData extends BaseVault {
+  key: string;
+  maturityDate?: number;
+  liquidityCents: BigNumber;
+  stakedTokenPriceUsd: BigNumber;
+  rewardTokenPriceUsd: BigNumber;
```
therealemjy: We use cents everywhere:
```suggestion
  stakedTokenPriceCents: BigNumber;
  rewardTokenPriceCents: BigNumber;
```

apps/evm/src/types/index.ts
```diff
@@ -464,19 +466,69 @@ export interface Transaction {
...
+  dailyEmissionMantissa: BigNumber;
+  isPaused: boolean;
   userHasPendingWithdrawalsFromBeforeUpgrade?: boolean;
 }
 
+export interface Vault extends VaultData, VaultMetadata {}
+
+export interface PendleVaultData extends BaseVault {
+  key: string;
+  maturityDate?: number;
```
therealemjy: If it's a date, then it needs to be of the type `Date`. If it's a timestamp in milliseconds, then it needs to be suffixed with `Ms` (`maturityTimestampMs`).

apps/evm/src/types/index.ts
```diff
@@ -464,19 +466,69 @@ export interface Transaction {
...
+  liquidityCents: BigNumber;
+  stakedTokenPriceUsd: BigNumber;
+  rewardTokenPriceUsd: BigNumber;
+  vToken: VToken;
+  underlyingAssetAddress: Address;
+  managerLink?: string;
+  vaultDeploymentTime?: number;
```
therealemjy:
```suggestion
  vaultDeploymentTimestampMs?: number;
```

apps/evm/src/types/index.ts
```diff
@@ -464,19 +466,69 @@ export interface Transaction {
...
+export interface Vault extends VaultData, VaultMetadata {}
+
+export interface PendleVaultData extends BaseVault {
+  key: string;
+  maturityDate?: number;
+  liquidityCents: BigNumber;
+  stakedTokenPriceUsd: BigNumber;
+  rewardTokenPriceUsd: BigNumber;
+  vToken: VToken;
+  underlyingAssetAddress: Address;
```
therealemjy: You don't need this prop, you can already get this value using the `vToken` property: `vToken.underlyingToken.address`

apps/evm/src/types/index.ts
```diff
@@ -464,19 +466,69 @@ export interface Transaction {
...
+  managerLink?: string;
+  vaultDeploymentTime?: number;
+  poolComptrollerAddress: Address;
```
therealemjy:
```suggestion
  poolComptrollerContractAddress: Address;
```

apps/evm/src/clients/api/queries/useGetVaults/useGetFormattedFixedRatedVaults/formatVaults/formatToPendleVault/index.ts
therealemjy, line 1: One helper per file

apps/evm/src/clients/api/queries/getFixedRatedVaults/types.ts
```diff
@@ -0,0 +1,57 @@
...
+  maturityDate: string;
+  createdAt: string;
+  updatedAt: string;
+};
+
+type VaultProduct = {
+  id: string;
+  chainId: string;
+  protocol: string;
+  vaultAddress: Address;
```
therealemjy, line 44: vaultVTokenAddress: Address;

apps/evm/src/clients/api/queries/getFixedRatedVaults/types.ts
```diff
@@ -0,0 +1,57 @@
...
+export type GetVaultProductsInput = {
+  chainId: ChainId;
+};
+
+type VaultProductAsset = {
+  name: string;
+  symbol: string;
+  address: string;
+  decimals: number;
+  priceUsd: number;
```
therealemjy, line 13: You should reuse existing types more:
```suggestion
  token: Token;
  priceCents: number;
```

apps/evm/src/clients/api/queries/useGetVaults/useGetPendleVaults/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+    managerAddress: vaultData.protocolData.pendleMarketAddress,
+    managerLink: vaultData.protocolData.pendleMarketAddress
+      ? `https://app.pendle.finance/trade/pools/${vaultData.protocolData.pendleMarketAddress}/zap/in?chain=bnbchain`
+      : undefined,
+    status,
+    underlyingAssetAddress: vaultData.underlyingAssetAddress,
+    vaultDeploymentTime: DEPLOY_DATE_MAP[vaultData.vaultAddress],
+    vToken: asset.vToken,
+    poolComptrollerAddress,
+  };
```
therealemjy: To make sure the type is respected, you can use a intermediary variable:

```suggestion
  const pendleVault: PendleVault = {
    key: vaultData.id,
    stakedToken,
    rewardToken,
    stakingAprPercentage: new BigNumber(vaultData.fixedApyDecimal).shiftedBy(2).toNumber(),
    userStakedMantissa: convertTokensToMantissa({
      value: asset.userSupplyBalanceTokens,
      token: stakedToken,
    }),
    totalStakedMantissa: convertTokensToMantissa({
      value: asset.supplyBalanceTokens,
      token: stakedToken,
    }),
    stakedTokenPriceUsd: new BigNumber(vaultData.protocolData.ptTokenPriceUsd),
    rewardTokenPriceUsd: new BigNumber(vaultData.protocolData?.accountingAsset?.priceUsd),
    maturityDate,
    liquidityCents: new BigNumber(vaultData.protocolData.liquidityCents),
    category: VaultCategory.YieldTokens,
    manager: VaultManager.Pendle,
    managerIcon: 'pendle' as const,
    managerAddress: vaultData.protocolData.pendleMarketAddress,
    managerLink: vaultData.protocolData.pendleMarketAddress
      ? `https://app.pendle.finance/trade/pools/${vaultData.protocolData.pendleMarketAddress}/zap/in?chain=bnbchain`
      : undefined,
    status,
    underlyingAssetAddress: vaultData.underlyingAssetAddress,
    vaultDeploymentTime: DEPLOY_DATE_MAP[vaultData.vaultAddress],
    vToken: asset.vToken,
    poolComptrollerAddress,
  };

  return pendleVault;
```

apps/evm/src/components/TextField/index.tsx
```diff
@@ -98,7 +98,9 @@ export const TextField: React.FC<TextFieldProps> = forwardRef<HTMLInputElement,
         >
           {(!!leftIconSrc || leftAdornment) && (
             <div className="flex items-center gap-x-2 shrink-0">
-              {typeof leftIconSrc === 'string' && <Icon name={leftIconSrc} className="size-5" />}
+              {typeof leftIconSrc === 'string' && (
+                <Icon name={leftIconSrc as IconName} className="size-5" />
+              )}
```
therealemjy, line 103:
```suggestion
              {!!leftIconSrc && (
                <Icon name={leftIconSrc} className="size-5" />
              )}
```

apps/evm/src/containers/Vault/hooks/useVaultUsdValues.ts
therealemjy, line 1: Rather than using a separate hook to fetch additional data, the values/prices in cents that you need here should be part of the vault objects themselves.

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -1833,11 +1838,57 @@
...
       "stablecoins": "Stablecoins",
       "vaults": "Vaults",
       "yieldTokens": "Yield tokens"
     },
+    "modals": {
+      "available": "Available",
+      "claim": "Claim",
+      "connectWalletMessage": "Please connect your wallet to stake {{tokenSymbol}}",
+      "convert": "Convert",
+      "currentStaked": "Current Staked",
```
therealemjy: "Current stake"

apps/evm/src/containers/Vault/VaultModals/OverviewTab/MarketInfo.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { CopyAddressButton } from 'containers/CopyAddressButton';
+import { Link } from 'containers/Link';
+import { useTranslation } from 'libs/translations';
+import type { AnyVault, PendleVault, VaultManager } from 'types';
+
+interface MarketInfoProps {
+  vault: AnyVault;
+}
+
+const isPendleVault = (vault: AnyVault): vault is PendleVault =>
```
therealemjy: Create a separate file for this helper

apps/evm/src/containers/Vault/VaultModals/OverviewTab/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { MarketInfo } from './MarketInfo';
+import { StrategyDiagram } from './StrategyDiagram';
+import { TotalDeposits } from './TotalDeposits';
+
+interface CollapsibleSectionProps {
+  title: string;
+  children: React.ReactNode;
+}
+
+const CollapsibleSection: React.FC<CollapsibleSectionProps> = ({ title, children }) => {
```
therealemjy: Only one component per file

apps/evm/src/containers/Vault/VaultModals/OverviewTab/TotalDeposits/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <AccordionAnimatedContent isOpen={isOpen}>
+        <div className="mt-4">
+          <div className="flex items-center gap-2 mb-2">
+            <span className="size-2 rounded-full bg-green" />
+            <span className="text-b1r text-grey">{t('vault.modals.overview.totalStaked')}</span>
+          </div>
+
+          <SupplyChart data={supplyChartData} selectedPeriod={selectedPeriod} />
+        </div>
+      </AccordionAnimatedContent>
```
therealemjy: This code is a duplicate of the `CollapsibleSection` here: https://github.com/VenusProtocol/venus-protocol-interface/blob/c7a17f9cac77d7030e87ecbe2e4f5414d0efcf36/apps/evm/src/containers/Vault/VaultModals/OverviewTab/index.tsx

apps/evm/src/containers/Vault/VaultModals/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <div className="flex flex-col min-w-0">
+        <span className="text-b1s truncate">{vault.stakedToken.symbol}</span>
+        {formattedMaturityDate && daysRemaining !== undefined && daysRemaining > 0 && (
+          <span className="text-b2r text-grey">
+            {formattedMaturityDate} ({t('vault.modals.daysRemaining', { days: daysRemaining })})
+          </span>
+        )}
+      </div>
+    </div>
+  );
```
therealemjy: This logic is duplicated here as well: https://github.com/VenusProtocol/venus-protocol-interface/blob/c7a17f9cac77d7030e87ecbe2e4f5414d0efcf36/apps/evm/src/containers/Vault/VaultCard/TokenIconWithPeriod/index.tsx

apps/evm/src/containers/Vault/VaultCard/StatusLabel/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  return (
+    <div className={cn(commonClassName, variantClassName, className)} {...props}>
+      {label}
+    </div>
+  );
+};
+
+StatusLabel.displayName = 'StatusLabel';
```
therealemjy: We don't use display names anywhere in the app, let's remove it

apps/evm/src/utilities/formatDateToUtc/index.ts
therealemjy, line 1: This formatter isn't needed, i18next already offers the possibility to format dates into a readable format.
Check this out: https://github.com/VenusProtocol/venus-protocol-interface/blob/c7a17f9cac77d7030e87ecbe2e4f5414d0efcf36/apps/evm/src/libs/translations/index.tsx#L40-L57

therealemjy (commented):
- User stake doesn’t get updated after withdrawing supply
- “Current Staked”: current stake
- The UI does not show how the stake balance of the user will evolve from this transaction, both in the Stake and Withdraw tabs
- The accordions used in the “Overview” tab seem unnecessary since they’re all open by default
- Inconsistent capitalization: some sentences capitalize every word, some only capitalize the first one. To match the rest of the website, let’s only capitalize the first word of every sentence
- Extra spaces between period parenthesis
- The stake/withdraw modal should be closed when a transaction is successfully submitted
- On the Dashboard page, Pendle vaults are shown with a “Daily emission” cell that shows a placeholder. This data isn’t relevant to this type of vault and should be removed
- On the “Stake” form of Pendle vaults, you need to check what the allowance provided by the user is (the code currently only checks whether an allowance has been provided, but they might have provided an allowance of 1 BNB while wanting to stake 2 BNB). This spending limit will have to be displayed when it isn’t MAX_UINT256. You can find this type of logic here: [https://github.com/VenusProtocol/venus-protocol-interface/blob/c7a17f9cac77d7030e87ecbe2e4f5414d0efcf36/apps/evm/src/containers/Vault/VaultCard/Legacy/VaultModals/TransactionForm/index.ts](https://github.com/VenusProtocol/venus-protocol-interface/blob/c7a17f9cac77d7030e87ecbe2e4f5414d0efcf36/apps/evm/src/containers/Vault/VaultCard/Legacy/VaultModals/TransactionForm/index.tsx)

apps/evm/src/clients/api/mutations/usePendlePtVault/utils.ts
david-sun-venus: I think for FE project, we could switch to `lodash-es` package instead and have nicer way to import things from it.

apps/evm/src/clients/api/mutations/usePendlePtVault/utils/index.ts
david-sun-venus: The reason I keep contractCallParams until before mutating is that both contractCallParams data structure and how they need to be formatted are different depends on which SC function to call. I think it makes more sense and easier to adapt more functions in the future when formatting it before being consumed in this case.

apps/evm/src/clients/api/mutations/usePendlePtVault/index.ts
david-sun-venus: After a second thought, I'd prefer remain current implementation, since it's similar to swapAndSupply and swapAndRepay, which contains multiple functions in one. There isn't much pros splitting them, by having redundant codes and needing multiple hooks in the same component.

apps/evm/src/clients/api/queries/getVaultProducts/index.ts
david-sun-venus: Perfect! I was struggling to find a proper name for this.

apps/evm/src/clients/api/mutations/usePendlePtVault/utils/index.ts
therealemjy: The logic is quite confusing to be honest, more confusing than mapping array arguments in my opinion.

apps/evm/src/components/TextField/index.tsx
david-sun-venus: I'm confused by the suggested logics here, doesn't it always get to the `<Icon />` rendering instead of `<TokenIconWithSymbol token={leftIconSrc} size="md" />`?

therealemjy: Ah yes, sorry ignore my comment I read this wrong.

apps/evm/src/utilities/formatDateToUtc/index.ts
david-sun-venus: This is trying to format and display the UTC time regardless the local timezone. I've checked the date-fns docs but not seeing any straight-forward utils that could do it like dayjs.

apps/evm/src/clients/api/queries/getFixedRatedVaults/types.ts
david-sun-venus: This type is for BE response data

apps/evm/src/containers/Vault/VaultModals/OverviewTab/TotalDeposits/index.tsx
david-sun-venus: Good catch!

apps/evm/src/containers/Vault/hooks/useVaultUsdValues.ts
david-sun-venus: In the first iteration I tried to inject price values into the vault object, but later found out that the pricing data is constantly updating, causing the vault object being refreshed and the vault component re-rendering / flashing every few seconds.

apps/evm/src/libs/translations/translations/en.json
david-sun-venus: It's where it shows how much user has staked, so I think "Current staked" would fit better.

apps/evm/src/utilities/formatDateToUtc/index.ts
therealemjy: We shouldn't display the UTC time without formatting it to the local time of the user. We have the capacity to display a customized date to people so that they don't have to convert it to their time themselves.

apps/evm/src/clients/api/mutations/usePendlePtVault/index.ts
therealemjy: Sure

apps/evm/src/libs/translations/translations/en.json
therealemjy: It's not valid English. It would either be "Currently staked" or "Current stake".

david-sun-venus: Let me update it to "Currently staked".

apps/evm/src/utilities/formatDateToUtc/index.ts
david-sun-venus: Removed and show local time instead.

apps/evm/src/clients/api/mutations/usePendlePtVault/index.ts
therealemjy, line 1: Each of these hooks needs to be in its own directory so that we keep the codebase cohesive.

apps/evm/src/types/index.ts
```diff
@@ -464,19 +466,69 @@ export interface Transaction {
...
+  asset: Asset;
+  managerLink?: string;
+  vaultDeploymentDate?: Date;
+  poolComptrollerContractAddress: Address;
+  poolName: string;
+}
+
+export interface PendleVault extends PendleVaultData, VaultMetadata {}
+
+export type AnyVault = Vault | PendleVault;
```
therealemjy: Having `AnyVault` and `Vault` will be confusing.
I recommend renaming `Vault` to `VenusVault` and `AnyVault` to `Vault`.

apps/evm/src/types/index.ts
```diff
@@ -464,19 +466,69 @@ export interface Transaction {
...
+  Others = 'others',
+}
+interface VaultMetadata {
+  category: VaultCategory;
+  manager: VaultManager;
+  managerIcon: IconName;
+  managerAddress?: Address;
+  status: VaultStatus;
+  key: string;
+}
```
therealemjy: You can merge this type with `BaseVault`

apps/evm/src/types/index.ts
therealemjy: You can simplify this implementation by only sharing a `BaseVault` type:

```suggestion
interface BaseVault {
  category: VaultCategory;
  manager: VaultManager;
  managerIcon: IconName;
  managerAddress?: Address;
  status: VaultStatus;
  key: string;
  stakedToken: Token;
  rewardToken: Token;
  stakingAprPercentage: number;
  totalStakedMantissa: BigNumber;
  dailyEmissionMantissa: BigNumber;
  isPaused: boolean;
  lockingPeriodMs?: number;
  userStakedMantissa?: BigNumber;
  poolIndex?: number;
}

export type VenusVault = BaseVault & {
  userHasPendingWithdrawalsFromBeforeUpgrade?: boolean;
};

export type PendleVault = BaseVault & {
  maturityDate?: Date;
  liquidityCents: BigNumber;
  stakedTokenPriceCents: BigNumber;
  rewardTokenPriceCents: BigNumber;
  asset: Asset;
  managerLink?: string;
  vaultDeploymentDate?: Date;
  poolComptrollerContractAddress: Address;
  poolName: string;
};

export type Vault = VenusVault | PendleVault;
```

apps/evm/src/clients/api/queries/getPendleSwapQuote/formatOutput.ts
```diff
@@ -0,0 +1,13 @@
+import BigNumber from 'bignumber.js';
+import { pick } from 'lodash-es';
+import type { PendleSwapApiResponse } from './types';
+
+export const formatOutput = (input: PendleSwapApiResponse) => {
+  return {
+    estimatedReceivedTokensMantissa: new BigNumber(input.estimatedOutput.amount),
+    feeCents: new BigNumber(input.fee.usd).shiftedBy(2),
```
therealemjy, line 9:
```suggestion
    feeCents: new BigNumber(input.fee.usd).shiftedBy(2).integerValue(),
```

therealemjy:
I'm seeing accordions that are opened by default:
Screenshot 2026-03-30 at 11 02 49

I think we should remove them then, they are only useful if they are closed by default.

therealemjy (approved):

PR #5473, opened 2026-03-27
chore: delete hosted images of old chains package versions

therealemjy:
## Changes

- automatically delete hosted images of old chains package versions


cuzz-venus (approved):
/lgtm

PR #5482, opened 2026-04-03
hotfix(evm): disable xvs pricing on zk chain

david-sun-venus:
## Changes

- Disable XVS price fetching on zk chain


apps/evm/src/utilities/xvsPriceOnZk/constants.ts
```diff
@@ -0,0 +1 @@
+export const XVS_FIXED_PRICE = 2.56;
```
therealemjy: Just to be consistent, let's use cents and suffix the value (XVS_FIXED_PRICE_CENTS).

therealemjy (commented):
It would be good to update the tests to check for the case where the price of XVS is set to a fixed value.

therealemjy:
@greptile

therealemjy (approved):

PR #5492, opened 2026-04-08
feat: support hidden balance

cuzz-venus:
## Jira ticket(s)

[VPD-956](https://venus-labs.atlassian.net/browse/VPD-956)

## Changes

- Hide market & staking balance when user clicks "Hide"

[VPD-956]: https://venus-labs.atlassian.net/browse/VPD-956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/evm/src/components/AccountHealthBar/index.tsx
```diff
@@ -23,134 +22,114 @@ const moderateBorrowLimitPercentage = 100 / HEALTH_FACTOR_MODERATE_THRESHOLD;
 export interface AccountHealthBarProps {
   borrowBalanceCents: number | undefined;
   borrowLimitCents: number | undefined;
-  liquidationThresholdCents: number | undefined;
   className?: string;
+  hiddenValues?: string;
```
therealemjy: To make it obvious what this prop represents, how about going for a name like `hideUserBalances`?

therealemjy (commented):
I think the values inside the account health bar tooltip should also be hidden when hiding balances:
Image

apps/evm/src/components/AccountHealthBar/index.tsx
cuzz-venus: @therealemjy Good suggestion. let me rename it and also add in tooltip 👍

therealemjy (approved):

PR #5535, opened 2026-04-15
fix: prompt users to switch chain if necessary when voting

therealemjy:
## Changes

- prompt users to switch chain if necessary when voting


therealemjy:
@greptile

cuzz-venus (approved):

PR #5548, opened 2026-04-19
feat: protection mode

cuzz-venus:
## Jira ticket(s)

[VPD-1015](https://venus-labs.atlassian.net/browse/VPD-1015)
[VPD-1092](https://venus-labs.atlassian.net/browse/VPD-1092)

## Changes

Data Layer

  - Add tokenSupplyPriceCents, tokenBorrowPriceCents, isProtectionModeEnabled to Asset type
  - Add userSupplyBalanceProtectedCents, userBorrowBalanceProtectedCents to Asset for
  protection-priced balances
  - Add userBorrowLimitProtectedCents, userBorrowBalanceProtectedCents to Pool type
  - Add isPriceProtected, collateralPriceMantissa, borrowPriceMantissa to ApiTokenPrice
  - Derive protection prices in formatOutput from API data, falling back to spot when protection is
   off

  Calculation Logic

  - calculateUserPoolValues: add protected borrow limit aggregation; health factor remains
  spot-based (matches contract LT path)
  - getSimulatedPool: track protected balances alongside spot, using supply/borrow price per action
  - BorrowForm: hard limit uses protected values; safe/moderate limits stay spot-based
  - WithdrawForm: same pattern — hard limit protected, safe/moderate spot
  - BoostForm: borrowing power uses protected values; max borrow tokens divides by
  tokenBorrowPriceCents
  - CapThreshold: pass supply or borrow price based on card type

  UI

  - New ProtectionModeIndicator component with two variants: icon (18×18) and label (16×16 icon +
  "Protected" badge)
  - New protectionShield SVG icon
  - MarketTable asset column: show icon when protection is active
  - MarketTable modal title: show icon before token symbol with collateral/borrow price tooltip
  - Market detail page header: show "Protected" label badge after Resilient Oracle
  - i18n: add protection mode tooltip keys to all 7 languages

  Tests & Mocks

  - Add new fields to all mock assets and fake pool data
  - Update snapshots

image
image
image

image

[VPD-1015]: https://venus-labs.atlassian.net/browse/VPD-1015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[VPD-1092]: https://venus-labs.atlassian.net/browse/VPD-1092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/evm/src/types/index.ts
```diff
@@ -153,8 +156,10 @@ export interface Asset {
   // connected
   userSupplyBalanceTokens: BigNumber;
   userSupplyBalanceCents: BigNumber;
+  userSupplyBalanceProtectedCents: BigNumber;
   userBorrowBalanceTokens: BigNumber;
   userBorrowBalanceCents: BigNumber;
+  userBorrowBalanceProtectedCents: BigNumber;
```
therealemjy, line 163: I don't recommend adding `userSupplyBalanceProtectedCents` and `userBorrowBalanceProtectedCents`, because it means every logic within the app will now have to check whether it should use these props or the regular `userSupplyBalanceCents` and `userBorrowBalanceCents` props.

I recommend updating `userSupplyBalanceCents` and `userBorrowBalanceCents` directly, so that when a token is being protected they use the capped prices. This way we can keep the logic seamless and reduce the complexity of our code.

apps/evm/src/pages/Market/OperationForm/BorrowForm/index.tsx
```diff
@@ -122,19 +124,19 @@ const BorrowForm: React.FC<BorrowFormProps> = ({ asset, pool, onSubmitSuccess })
...
     let marginWithUserSafeBorrowLimitTokens =
       // We base the safe borrow limit on the liquidation threshold because that's the base used to
-      // calculate the health factor
+      // calculate the health factor (spot-based, since liquidation uses spot)
       pool.userLiquidationThresholdCents
         .div(HEALTH_FACTOR_SAFE_MAX_THRESHOLD)
         .minus(pool.userBorrowBalanceCents)
-        // Convert to tokens
+        // Convert to tokens (spot, to match health factor pricing)
         .dividedBy(asset.tokenPriceCents);
```
therealemjy, line 140: This logic should also use `tokenBorrowPriceCents`, as the borrowing power of a user is also dependent on whether protection mode is enabled or not.

apps/evm/src/pages/Market/OperationForm/BorrowForm/index.tsx
```diff
@@ -144,7 +146,7 @@ const BorrowForm: React.FC<BorrowFormProps> = ({ asset, pool, onSubmitSuccess })
     let marginWithUserModerateRiskBorrowLimitTokens = pool.userLiquidationThresholdCents
       .div(HEALTH_FACTOR_MODERATE_THRESHOLD)
       .minus(pool.userBorrowBalanceCents)
-      // Convert to tokens
+      // Convert to tokens (spot, to match health factor pricing)
       .dividedBy(asset.tokenPriceCents);
```
therealemjy, line 150: Same thing here, this logic should use `tokenBorrowPriceCents`

apps/evm/src/libs/translations/translations/zh-Hant.json
```diff
@@ -976,7 +967,12 @@
     "assetColumn": {
       "eMode": "該市場屬於 {{eModeGroupName}} E-mode組",
       "isolationMode": "該市場屬於 {{eModeGroupName}} 隔離模式組",
-      "pausedAssetTooltip": "資產已被 Venus 治理暫停。借款與供應已禁用。"
+      "pausedAssetTooltip": "資產已被 Venus 治理暫停。借款與供應已禁用。",
+      "protectionMode": "Protection Mode is active for {{tokenName}} due to price volatility, affecting its collateral and borrow prices.",
+      "protectionModeSupply": "This market is suffering volatility. The Protected Collateral Value is {{protectedValue}}. The conservative value used to determine your borrow power.",
+      "protectionModeBorrow": "This market is suffering volatility. The Protected Borrow Value is {{protectedValue}}. The conservative value that counts against your borrow power.",
+      "protectionModeDetailCollateral": "This market is suffering volatility. The Protected Collateral Price is {{collateralPrice}}. The conservative price used to determine your borrow power.",
+      "protectionModeDetailBorrow": "The Protected Borrow Price is {{borrowPrice}}. The conservative price used to calculate how many of this asset can be borrowed."
```
therealemjy: This needs to be translated

apps/evm/src/libs/translations/translations/zh-Hans.json
```diff
@@ -976,7 +967,12 @@
     "assetColumn": {
       "eMode": "该市场属于 {{eModeGroupName}} E-mode组",
       "isolationMode": "该市场属于 {{eModeGroupName}} 隔离模式组",
-      "pausedAssetTooltip": "资产已被 Venus 治理暂停。借款与存款已禁用。"
+      "pausedAssetTooltip": "资产已被 Venus 治理暂停。借款与存款已禁用。",
+      "protectionMode": "Protection Mode is active for {{tokenName}} due to price volatility, affecting its collateral and borrow prices.",
+      "protectionModeSupply": "This market is suffering volatility. The Protected Collateral Value is {{protectedValue}}. The conservative value used to determine your borrow power.",
+      "protectionModeBorrow": "This market is suffering volatility. The Protected Borrow Value is {{protectedValue}}. The conservative value that counts against your borrow power.",
+      "protectionModeDetailCollateral": "This market is suffering volatility. The Protected Collateral Price is {{collateralPrice}}. The conservative price used to determine your borrow power.",
+      "protectionModeDetailBorrow": "The Protected Borrow Price is {{borrowPrice}}. The conservative price used to calculate how many of this asset can be borrowed."
```
therealemjy: This needs to be translated

apps/evm/src/libs/translations/translations/vi.json
```diff
@@ -976,7 +967,12 @@
...
       "isolationMode": "Thị trường này thuộc nhóm chế độ cô lập {{eModeGroupName}}",
-      "pausedAssetTooltip": "Tài sản đã bị tạm dừng bởi Venus Governance. Việc vay và cung cấp đã bị vô hiệu hóa."
+      "pausedAssetTooltip": "Tài sản đã bị tạm dừng bởi Venus Governance. Việc vay và cung cấp đã bị vô hiệu hóa.",
+      "protectionMode": "Protection Mode is active for {{tokenName}} due to price volatility, affecting its collateral and borrow prices.",
+      "protectionModeSupply": "This market is suffering volatility. The Protected Collateral Value is {{protectedValue}}. The conservative value used to determine your borrow power.",
+      "protectionModeBorrow": "This market is suffering volatility. The Protected Borrow Value is {{protectedValue}}. The conservative value that counts against your borrow power.",
+      "protectionModeDetailCollateral": "This market is suffering volatility. The Protected Collateral Price is {{collateralPrice}}. The conservative price used to determine your borrow power.",
+      "protectionModeDetailBorrow": "The Protected Borrow Price is {{borrowPrice}}. The conservative price used to calculate how many of this asset can be borrowed."
```
therealemjy: This needs to be translated

apps/evm/src/libs/translations/translations/tr.json
```diff
@@ -976,7 +967,12 @@
...
-      "pausedAssetTooltip": "Varlık Venus Yönetişimi tarafından duraklatıldı. Borç alma ve tedarik devre dışı bırakıldı."
+      "pausedAssetTooltip": "Varlık Venus Yönetişimi tarafından duraklatıldı. Borç alma ve tedarik devre dışı bırakıldı.",
+      "protectionMode": "Protection Mode is active for {{tokenName}} due to price volatility, affecting its collateral and borrow prices.",
+      "protectionModeSupply": "This market is suffering volatility. The Protected Collateral Value is {{protectedValue}}. The conservative value used to determine your borrow power.",
+      "protectionModeBorrow": "This market is suffering volatility. The Protected Borrow Value is {{protectedValue}}. The conservative value that counts against your borrow power.",
+      "protectionModeDetailCollateral": "This market is suffering volatility. The Protected Collateral Price is {{collateralPrice}}. The conservative price used to determine your borrow power.",
+      "protectionModeDetailBorrow": "The Protected Borrow Price is {{borrowPrice}}. The conservative price used to calculate how many of this asset can be borrowed."
```
therealemjy: This needs to be translated

apps/evm/src/libs/translations/translations/th.json
```diff
@@ -976,7 +967,12 @@
...
       "isolationMode": "ตลาดนี้เป็นส่วนหนึ่งของกลุ่มโหมดแยก {{eModeGroupName}}",
-      "pausedAssetTooltip": "สินทรัพย์ถูกหยุดโดยการกำกับดูแลของ Venus การยืมและการฝากถูกปิดใช้งาน"
+      "pausedAssetTooltip": "สินทรัพย์ถูกหยุดโดยการกำกับดูแลของ Venus การยืมและการฝากถูกปิดใช้งาน",
+      "protectionMode": "Protection Mode is active for {{tokenName}} due to price volatility, affecting its collateral and borrow prices.",
+      "protectionModeSupply": "This market is suffering volatility. The Protected Collateral Value is {{protectedValue}}. The conservative value used to determine your borrow power.",
+      "protectionModeBorrow": "This market is suffering volatility. The Protected Borrow Value is {{protectedValue}}. The conservative value that counts against your borrow power.",
+      "protectionModeDetailCollateral": "This market is suffering volatility. The Protected Collateral Price is {{collateralPrice}}. The conservative price used to determine your borrow power.",
+      "protectionModeDetailBorrow": "The Protected Borrow Price is {{borrowPrice}}. The conservative price used to calculate how many of this asset can be borrowed."
```
therealemjy: This needs to be translated

apps/evm/src/libs/translations/translations/ja.json
```diff
@@ -976,7 +967,12 @@
     "assetColumn": {
       "eMode": "このマーケットは{{eModeGroupName}}のEモードグループに含まれます",
       "isolationMode": "このマーケットは{{eModeGroupName}}の隔離モードグループに含まれます",
-      "pausedAssetTooltip": "資産はVenusガバナンスにより停止されています。借入と供給が無効化されています。"
+      "pausedAssetTooltip": "資産はVenusガバナンスにより停止されています。借入と供給が無効化されています。",
+      "protectionMode": "Protection Mode is active for {{tokenName}} due to price volatility, affecting its collateral and borrow prices.",
+      "protectionModeSupply": "This market is suffering volatility. The Protected Collateral Value is {{protectedValue}}. The conservative value used to determine your borrow power.",
+      "protectionModeBorrow": "This market is suffering volatility. The Protected Borrow Value is {{protectedValue}}. The conservative value that counts against your borrow power.",
+      "protectionModeDetailCollateral": "This market is suffering volatility. The Protected Collateral Price is {{collateralPrice}}. The conservative price used to determine your borrow power.",
+      "protectionModeDetailBorrow": "The Protected Borrow Price is {{borrowPrice}}. The conservative price used to calculate how many of this asset can be borrowed."
```
therealemjy: This needs to be translated

apps/evm/src/containers/MarketTable/index.tsx
```diff
@@ -216,7 +217,36 @@ export const MarketTable: React.FC<MarketTableProps> = ({
...
+                        })}
+                      </p>
+                      <p className="mt-1">
+                        {t('marketTable.assetColumn.protectionModeDetailBorrow', {
+                          borrowPrice: formatCentsToReadableValue({
+                            value: selectedAsset.tokenBorrowPriceCents,
+                            shorten: false,
+                          }),
+                        })}
+                      </p>
```
therealemjy: You can merge this into a singular translation key and use a `LineBreak` component within it to break lines (see `liquidationThresholdTooltip` translation key for an example).

apps/evm/src/components/ProtectionModeIndicator/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+        </span>
+      </Tooltip>
+    );
+  }
+
+  return (
+    <Tooltip className={cn('inline-flex items-center', className)} content={tooltipContent}>
+      <Icon name="protectionShield" className="h-[18px] w-[18px]" />
+    </Tooltip>
+  );
```
therealemjy: You can deduplicate code here:

```suggestion
  return (
    <Tooltip className={cn('inline-flex items-center', className)} content={tooltipContent}>
      {variant === 'label' ? (
        <span className="inline-flex items-center gap-x-1 rounded-full border border-green bg-green/10 px-2 py-0.5 text-xs text-offWhite">
          <Icon name="protectionShield" className="h-4 w-4" />
          <span>{t('protectionModeIndicator.label')}</span>
        </span>
      ) : (
        <Icon name="protectionShield" className="h-[18px] w-[18px]" />
      )}
    </Tooltip>
  );
```

apps/evm/src/components/ProtectionModeIndicator/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const ProtectionModeIndicator: React.FC<ProtectionModeIndicatorProps> = ({
+  variant = 'icon',
+  tooltip,
+  tokenName,
+  className,
+}) => {
+  const { t } = useTranslation();
+
+  const tooltipContent = tooltip ?? t('marketTable.assetColumn.protectionMode', { tokenName });
```
therealemjy: I think you can go a step further with this component and include the tooltip inside it directly, based on the variant. You could remove the `tooltip` prop and add `tokenSupplyPriceCents` and `tokenBorrowPriceCents` instead`.

Then you could determine the tooltip content like so:
```tsx
const values = {
    tokenName,
    tokenSupplyPriceCents,
    tokenBorrowPriceCents1,
  };

  const tooltipContent =
    variant === 'label' ? (
      <Trans
        i18nKey="protectionModeIndicator.detailedTooltip"
        components={{
          LineBreak: <br />,
        }}
        values={values}
      />
    ) : (
      t('protectionModeIndicator.detailedTooltip', values)
    );
```

This way places that are rendering the `ProtectionModeIndicator` component do not need to pass down a tooltip (this will remove the duplicate you have here: https://github.com/VenusProtocol/venus-protocol-interface/blob/dbf5ddb9ce8251b9b909af44414b887283a30d07/apps/evm/src/containers/Layout/Header/MarketInfo/index.tsx#L147-L169 and here https://github.com/VenusProtocol/venus-protocol-interface/blob/dbf5ddb9ce8251b9b909af44414b887283a30d07/apps/evm/src/containers/MarketTable/index.tsx#L223-L245)

apps/evm/src/clients/api/queries/useGetPools/getPools/formatOutput/index.ts
```diff
@@ -260,12 +282,19 @@ export const formatOutput = ({
 
       const userSupplyBalanceCents = userSupplyBalanceTokens.multipliedBy(tokenPriceCents);
       const userBorrowBalanceCents = userBorrowBalanceTokens.multipliedBy(tokenPriceCents);
+      const userSupplyBalanceProtectedCents =
+        userSupplyBalanceTokens.multipliedBy(tokenSupplyPriceCents);
+      const userBorrowBalanceProtectedCents =
+        userBorrowBalanceTokens.multipliedBy(tokenBorrowPriceCents);
```
therealemjy, line 288: Following my previous comment, I think you can remove `userSupplyBalanceProtectedCents` and `userBorrowBalanceProtectedCents`:


```suggestion
const userSupplyBalanceCents = userSupplyBalanceTokens.multipliedBy(tokenSupplyPriceCents);
const userBorrowBalanceCents = userBorrowBalanceTokens.multipliedBy(tokenBorrowPriceCents);
```

apps/evm/src/types/index.ts
cuzz-venus: hey @therealemjy Thanks for the suggestion!

I actually considered this approach first and weighed both options before going with the separate
  fields. The reasons I ended up with **ProtectedCents**:
 - Health factor accuracy: If I modify `userSupplyBalanceCents and userBorrowBalanceCents` directly, the
  health factor calculation in calculateUserPoolValues would also use protection prices. But per the PRD,
  liquidation uses spot prices, so health factor should be spot / spot to accurately reflect real
  liquidation risk. With separate fields I can keep health factor spot-based while using protection prices
   for borrow limit.

 - Readability & review clarity: Having explicit **ProtectedCents** fields makes it obvious in the diff and in future code reviews which calculations switched to protection pricing. If the existing fields silently change semantics, it's harder to tell at a glance whether a given calculation is using spot or protected prices.

That said, I understand the trade-off — having two sets of fields adds complexity for consumers. Happy to switch to modifying the existing fields directly if we prefers simplicity. For health factor, I can handle it locally in `calculateUserPoolValues` by computing the spot-based values from `tokens × tokenPriceCents` directly

What do you think?

apps/evm/src/pages/Market/OperationForm/BorrowForm/index.tsx
cuzz-venus: I kept spot here because this calculation is health-factor-based (LT path uses spot in the contract and have refer on PRD), so I matched the divisor to stay consistent. But if we go with modifying `userBorrowBalanceCents` directly per your first comment, then agreed — this should use `tokenBorrowPriceCents`.

apps/evm/src/pages/Market/OperationForm/BorrowForm/index.tsx
cuzz-venus: Same reasoning as above — will update together if we switch to the approach you suggested.

apps/evm/src/types/index.ts
cuzz-venus: I have remove `userSupplyBalanceProtectedCents and userBorrowBalanceProtectedCents`, and do self-test passed

apps/evm/src/libs/translations/translations/zh-Hant.json
cuzz-venus: updated! thanks!

apps/evm/src/libs/translations/translations/zh-Hans.json
cuzz-venus: updated! thanks!

apps/evm/src/libs/translations/translations/ja.json
cuzz-venus: updated! thanks!

apps/evm/src/containers/MarketTable/index.tsx
cuzz-venus: updated! thanks!

apps/evm/src/components/ProtectionModeIndicator/index.tsx
cuzz-venus: agree, I do it before, and concern if its different tooltip, maybe we can change directly, so i set as props. since we have different tooltip most of them

apps/evm/src/clients/api/queries/useGetPools/getPools/formatOutput/index.ts
cuzz-venus: updated with no `userBorrowBalanceProtectedCents` `userSupplyBalanceProtectedCents`

apps/evm/src/clients/api/queries/useGetPools/getPools/formatOutput/index.ts
```diff
@@ -108,6 +108,28 @@ export const formatOutput = ({
...
+
+      const tokenBorrowPriceCents =
+        isProtectionModeEnabled && correspondingOraclePrice.borrowPriceMantissa
+          ? convertDollarsToCents(
+              convertPriceMantissaToDollars({
+                priceMantissa: correspondingOraclePrice.borrowPriceMantissa,
+                decimals: vToken.underlyingToken.decimals,
+              }),
+            )
+          : convertDollarsToCents(tokenPriceDollars);
```
therealemjy: This is a tiny thing, but since that we use `convertDollarsToCents(tokenPriceDollars)` in three places in this file, you could move the `tokenPriceCents` variable instantiation line 200 up and use it.
```suggestion
      const tokenSupplyPriceCents =
        isProtectionModeEnabled && correspondingOraclePrice.collateralPriceMantissa
          ? convertDollarsToCents(
              convertPriceMantissaToDollars({
                priceMantissa: correspondingOraclePrice.collateralPriceMantissa,
                decimals: vToken.underlyingToken.decimals,
              }),
            )
          : tokenPriceCents;

      const tokenBorrowPriceCents =
        isProtectionModeEnabled && correspondingOraclePrice.borrowPriceMantissa
          ? convertDollarsToCents(
              convertPriceMantissaToDollars({
                priceMantissa: correspondingOraclePrice.borrowPriceMantissa,
                decimals: vToken.underlyingToken.decimals,
              }),
            )
          : tokenPriceCents;
```

apps/evm/src/components/ProtectionModeIndicator/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+const LearnMoreLink = (
+  // biome-ignore lint/a11y/useAnchorContent: content is provided by Trans
+  <a
+    href={VENUS_PROTECTION_MODE_DOC_URL}
+    className="text-blue underline"
+    target="_blank"
+    rel="noopener noreferrer"
+  />
+);
```
therealemjy: I know the AIs love to do that, but one of the practices of this codebase is to only define one component per file, so this component should be moved to its own directory.

apps/evm/src/constants/production.ts
```diff
@@ -2,6 +2,8 @@ export const MAIN_PRODUCTION_HOST = 'app.venus.io';
 export const VENUS_FLUX_URL = 'https://flux.venus.io';
 export const VENUS_DOC_URL = 'https://docs-v4.venus.io';
 export const VENUS_PRIME_DOC_URL = `${VENUS_DOC_URL}/whats-new/prime-yield`;
+// TODO: update to the actual protection mode doc page once available
```
therealemjy, line 5: I've created a ticket and assigned it to you just so we don't forget before going live: https://venus-labs.atlassian.net/browse/VPD-1106

apps/evm/src/containers/Layout/Header/MarketInfo/index.tsx
```diff
@@ -134,11 +140,20 @@ export const MarketInfo = () => {
...
+                variant="icon"
+                tokenName={asset.vToken.underlyingToken.symbol}
+                tokenSupplyPriceCents={asset.tokenSupplyPriceCents}
+                tokenBorrowPriceCents={asset.tokenBorrowPriceCents}
+              />
+              {readablePrice}
+            </span>
+          ) : (
+            readablePrice
+        ),
```
therealemjy: A small thing you can do to make this logic even simpler:

```suggestion
		<span className="inline-flex items-center gap-x-2">
            {asset?.isProtectionModeEnabled && (
              <ProtectionModeIndicator
                variant="icon"
                tokenName={asset.vToken.underlyingToken.symbol}
                tokenSupplyPriceCents={asset.tokenSupplyPriceCents}
                tokenBorrowPriceCents={asset.tokenBorrowPriceCents}
              />
            )}

            {readablePrice}
        </span>
```

apps/evm/src/containers/MarketTable/useColumns/index.tsx
```diff
@@ -155,6 +158,18 @@ export const useColumns = ({
...
 
+              {asset.isProtectionModeEnabled && (
+                <ProtectionModeIndicator
+                  variant="icon"
+                  tooltipType={marketType ?? 'list'}
+                  tokenName={asset.vToken.underlyingToken.symbol}
+                  tokenSupplyPriceCents={asset.tokenSupplyPriceCents}
+                  tokenBorrowPriceCents={asset.tokenBorrowPriceCents}
+                  userSupplyBalanceCents={asset.userSupplyBalanceCents}
+                  userBorrowBalanceCents={asset.userBorrowBalanceCents}
```
therealemjy: Since the tooltip displays protected values, you need to pass the protected balances:
```suggestion
                  userSupplyBalanceCents={asset.userSupplyBalanceProtectedCents}
                  userBorrowBalanceCents={asset.userBorrowBalanceProtectedCents}
```

apps/evm/src/types/index.ts
```diff
@@ -222,8 +227,11 @@ export interface Pool {
   userSupplyBalanceCents?: BigNumber;
   userBorrowBalanceCents?: BigNumber;
   userBorrowLimitCents?: BigNumber;
+  userBorrowLimitProtectedCents?: BigNumber;
+  userBorrowBalanceProtectedCents?: BigNumber;
```
therealemjy, line 231: This is a suggestion you can ignore, but perhaps I'd place "Protected" earlier in the variable names to make it easier for a dev to quickly understand what this data represents.
`userSupplyBalanceProtectedCents` -> `userProtectedSupplyBalanceCents`
`userBorrowBalanceProtectedCents` -> `userProtectedBorrowBalanceCents`
`userBorrowLimitProtectedCents` -> `userProtectedBorrowLimitCents`

apps/evm/src/types/index.ts
```diff
@@ -222,8 +227,11 @@ export interface Pool {
...
   userLiquidationThresholdCents?: BigNumber;
   userYearlyEarningsCents?: BigNumber;
+  userHasProtectionModeAssets?: boolean;
```
therealemjy: This prop seems to only be used by the `AccountHealthBar` component.
You could actually remove it from the `Asset` type (one less prop to worry about) and instead determine that protection mode is enabled by checking that `borrowBalanceProtectedCents` passed to `AccountHealthBar` is different than `borrowBalanceCents`:
```tsx
const isProtectionModeEnabled = !!borrowBalanceProtectedCents?.isEqualTo(borrowBalanceCents);
```

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -137,10 +137,12 @@
...
     "borrowed": "Borrowed:",
+    "borrowedSpotTooltip": "This borrow value is calculated based on the spot price of the borrowed asset and will be used to determine your account's health status.",
     "liquidationThreshold": "Liquidation threshold:",
     "liquidationThresholdShort": "LT:",
     "liquidationThresholdTooltip": "LT = Liquidation Threshold<LineBreak/>When your borrow balance reaches this value, your position becomes eligible for liquidation.<LineBreak/><LineBreak/>Learn more about the <Link>calculation.</Link>",
-    "tooltip": "Borrow balance:<LineBreak/>{{borrowBalance}} ({{borrowLimitUsedPercentage}} of your borrow limit)<LineBreak/><LineBreak/>Borrow limit:<LineBreak/>{{borrowLimit}}"
+    "tooltip": "Borrow balance:<LineBreak/>{{borrowBalance}} ({{borrowLimitUsedPercentage}} of your borrow limit)<LineBreak/><LineBreak/>Borrow limit:<LineBreak/>{{borrowLimit}}",
+    "tooltipProtection": "Borrow balance (Protection):<LineBreak/>{{borrowBalanceProtected}} ({{borrowLimitUsedPercentage}} of your borrow limit)<LineBreak/><LineBreak/>Borrow balance (Spot): {{borrowBalance}}<LineBreak/>Borrow limit: {{borrowLimit}}"
```
therealemjy: I know you're only following the PRD/designs, but "Protection" should be "protected" and "Spot" shouldn't be capitalized:

```suggestion
    "tooltipProtection": "Borrow balance (protected):<LineBreak/>{{borrowBalanceProtected}} ({{borrowLimitUsedPercentage}} of your borrow limit)<LineBreak/><LineBreak/>Borrow balance (spot): {{borrowBalance}}<LineBreak/>Borrow limit: {{borrowLimit}}"
```

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -985,7 +987,12 @@
     "assetColumn": {
       "eMode": "This market is part of the {{eModeGroupName}} E-mode group",
       "isolationMode": "This market is part of the {{eModeGroupName}} isolation mode group",
-      "pausedAssetTooltip": "Asset paused by Venus Governance. Borrowing and supplying have been disabled."
+      "pausedAssetTooltip": "Asset paused by Venus Governance. Borrowing and supplying have been disabled.",
+      "protectionMode": "Protection Mode is active for {{tokenName}} due to price volatility, affecting its collateral and borrow prices. <LearnMore>Learn more</LearnMore>",
```
```suggestion
      "protectionMode": "Protection mode is active for {{tokenName}} due to price volatility, affecting its collateral and borrow prices. <LearnMore>Learn more</LearnMore>",
```

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -985,7 +987,12 @@
...
+      "protectionModePair": "Protection Mode is active for {{tokenName}} due to price volatility, affecting their collateral and borrow prices. <LearnMore>Learn more</LearnMore>",
```
```suggestion
      "protectionModePair": "Protection mode is active for {{tokenName}} due to price volatility, affecting their collateral and borrow prices. <LearnMore>Learn more</LearnMore>",
```

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -985,7 +987,12 @@
...
+      "protectionModeSupply": "This market is experiencing high volatility. The protected collateral value is calculated as {{protectedValue}}, representing the conservative value used to determine your borrow power.",
+      "protectionModeBorrow": "This market is experiencing high volatility. The protected borrow value is calculated as {{protectedValue}}, representing the conservative value that counts against your borrow power.",
+      "protectionModeDetail": "This market is experiencing high volatility. The Protected Collateral Price is {{collateralPrice}} — the conservative price used to determine your borrow power. The Protected Borrow Price is {{borrowPrice}} — the conservative price used to calculate how much of this asset can be borrowed."
```
therealemjy: Let's remove the `—` characters, this makes the text look very AI-like. Also, let's space the paragraphs up by using a `LineBreak` component like you did in other translations.
```suggestion
      "protectionModeDetail": "This market is experiencing high volatility.<LineBreak/><LineBreak/>The Protected Collateral Price is {{collateralPrice}}, the conservative price used to determine your borrow power.<LineBreak/><LineBreak/>The Protected Borrow Price is {{borrowPrice}}, the conservative price used to calculate how much of this asset can be borrowed."
```

therealemjy (commented):
Really good work overall, nicely done. My comments are all minor stuff, only one is important; the one about passing the protected user balances to the `ProtectionModeIndicator` component.

apps/evm/src/pages/Market/OperationForm/BorrowForm/index.tsx
therealemjy: Yeah your approach makes complete sense actually.

apps/evm/src/types/index.ts
therealemjy: So at the end you were right that we should have `userSupplyBalanceProtectedCents` and `userBorrowBalanceProtectedCents`.

No need to make any changes to the current PR state (the latest code is correct).

apps/evm/src/pages/Market/MarketHistory/Card/index.tsx
```diff
@@ -115,7 +115,7 @@ export const Card: React.FC<CardProps> = ({
       topContent={
         <CapThreshold
           type={type}
-          tokenPriceCents={asset.tokenPriceCents}
+          tokenPriceCents={type === 'supply' ? asset.tokenSupplyPriceCents : asset.tokenBorrowPriceCents}
```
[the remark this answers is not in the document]
therealemjy: Yeah for this one I don't think you need to pass the supply/borrow prices: we can display the caps using spot prices, since we display balances with spot prices everywhere.

apps/evm/src/clients/api/queries/useGetPools/getPools/getApiPools/index.ts
```diff
@@ -145,6 +145,9 @@ export interface ApiTokenPrice {
   priceOracleAddress: Address | null;
   isPriceInvalid: boolean;
   hasErrorFetchingPrice: boolean;
+  isPriceProtected: boolean;
+  collateralPriceMantissa: string;
+  borrowPriceMantissa: string;
```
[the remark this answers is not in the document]
therealemjy: The AI is right.

apps/evm/src/clients/api/queries/useGetPools/getPools/formatOutput/index.ts
cuzz-venus: Good catch, updated!
Thanks :)

apps/evm/src/components/ProtectionModeIndicator/index.tsx
cuzz-venus: Makes sense, moved it to its own directory. I'd normally do the same — just kept it inline since it's only used in this file, but extracting it is cleaner. 😄

cuzz-venus: updated! thanks

apps/evm/src/constants/production.ts
cuzz-venus: Noted, thanks!

apps/evm/src/containers/Layout/Header/MarketInfo/index.tsx
cuzz-venus: Cleaner, updated!

apps/evm/src/components/ProtectionModeIndicator/index.tsx
therealemjy: Thanks! :)

apps/evm/src/containers/MarketTable/useColumns/index.tsx
cuzz-venus: Good point, fixed!

apps/evm/src/pages/Market/MarketHistory/Card/index.tsx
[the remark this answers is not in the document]
cuzz-venus: Agreed, reverted to spot.

apps/evm/src/clients/api/queries/useGetPools/getPools/getApiPools/index.ts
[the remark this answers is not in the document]
cuzz-venus: Updated to string | null.

apps/evm/src/types/index.ts
cuzz-venus: I originally had it this way too for readability, but your approach keeps the type leaner.
Updated! 😄

apps/evm/src/libs/translations/translations/en.json
cuzz-venus: Really really thanks for catching that! My English casing instincts aren't always on point — appreciate the correction.

apps/evm/src/libs/translations/translations/en.json
cuzz-venus: Fixed

apps/evm/src/libs/translations/translations/en.json
cuzz-venus: Fixed

apps/evm/src/libs/translations/translations/en.json
cuzz-venus: Fixed

apps/evm/src/libs/translations/translations/en.json
therealemjy: No worries at all, this is not on devs to fix content anyway so it's not your responsibility.

therealemjy (approved):
Looks good to me.
The only two things I see left to do:
- update the link to the documentation (once it's live)
- fix the linting issue

PR #5549, opened 2026-04-20
feat: improve Yield+ feature

therealemjy:
## Changes

- add header to Yield+ chart
- improve reduce flows


apps/evm/src/libs/translations/translations/ja.json
```diff
@@ -743,6 +750,7 @@
       "utilizationRate": "利用率"
     },
     "menu": {
+      "beta": "TRANSLATION NEEDED",
```
cuzz-venus: This needs to be translated. The placeholder "TRANSLATION NEEDED" will render literally in the nav for non-English users.

apps/evm/src/constants/fullRepaymentBuffer.ts
```diff
@@ -1 +1 @@
-export const FULL_REPAYMENT_BUFFER_PERCENTAGE = 0.01;
+export const FULL_REPAYMENT_BUFFER_PERCENTAGE = 0.001;
```
cuzz-venus, line 1: Just a heads-up on this one — since `FULL_REPAYMENT_BUFFER_PERCENTAGE` is a global constant, lowering it from 0.01 to 0.001 also affects the other repay paths that import it (**RepayWithWalletBalanceForm, useRepay, useImportSupplyPosition, RepayWithCollateralForm**). 
Does this one we only want effect on Yield+ or also on where we use?

therealemjy: Yeah it's voluntary, as it will also benefit other flows that use a swap. It won't have a real impact on user as the change is minimal.

apps/evm/src/libs/translations/translations/ja.json
therealemjy: Good catch

apps/evm/src/pages/Trade/index.tsx
```diff
@@ -44,11 +47,25 @@ const YieldPlus: React.FC = () => {
...
     rangeMs: ONE_DAY_MS,
   });
 
+  // Determine precision based on the token pair price
+  let pricePrecision = 6;
+
+  if (priceCentsRatio?.isLessThan(1)) {
+    const firstNonZeroIndex = priceCentsRatio
+      ?.toFixed()
+      .split('.')[1]
```
cuzz-venus, line 62: Tiny edge case here — if priceCentsRatio ever ends up as exactly 0 (e.g. an oracle hiccup or empty K-line response), priceCentsRatio.toFixed() returns "0", so .split('.')[1] is undefined and the next .match(...) would throw, which would crash the whole page.
should we catch this case? or this should basically never happen

```suggestion
      .split('.')[1]?
```

apps/evm/src/containers/Layout/NavBar/MenuItem/index.tsx
```diff
@@ -114,8 +115,19 @@ export const MenuItem: React.FC<MenuItemProps> = ({ item, onClick }) => {
...
+    <Link
+      to={item.to}
+      href={item.href}
+      onClick={onClick}
+      className={cn(sharedContainerClassName, 'flex items-center gap-x-2')}
+    >
+      <span>{item.label}</span>
+
+      {item.isBeta && (
+        <div className="bg-blue px-1.5 py-0.5 rounded-full text-xs font-semibold text-white">
```
cuzz-venus: Tiny one — text-xs font-semibold (12px / 600) maps cleanly to the text-b2s token defined in packages/ui/src/theme.css (12px / 600).


```suggestion
        <div className="bg-blue px-1.5 py-0.5 rounded-full text-b2s text-white">
```

apps/evm/src/pages/YieldPlus/index.tsx
```diff
@@ -24,6 +25,8 @@ const YieldPlus: React.FC = () => {
 
   const doNotShowBanner = store.use.doNotShowBanner();
 
+  const [interval, setInterval] = useState<ApiOhlcInterval>(ApiOhlcInterval['1h']);
```
cuzz-venus: setInterval shadows the global window.setInterval, can we consider like this


```suggestion
  const [chartInterval, setChartInterval] = useState<ApiOhlcInterval>(ApiOhlcInterval['1h']);
```

apps/evm/src/pages/Trade/index.tsx
therealemjy: This can't happen (the price of a token would never be exactly 0).

apps/evm/src/containers/Layout/NavBar/MenuItem/index.tsx
therealemjy: good call

apps/evm/src/constants/fullRepaymentBuffer.ts
cuzz-venus: Got it!

apps/evm/src/pages/Trade/index.tsx
cuzz-venus: Got it! Thanks!

apps/evm/src/pages/YieldPlus/index.tsx
therealemjy: Sure, good catch

therealemjy:
@greptile

toolsvenus (approved):

PR #5563, opened 2026-04-28
feat: rename Yield+ to Trade

therealemjy:
## Changes

- rename Yield+ to Trade


cuzz-venus (approved):
Everything looks great to me here. The rename is remarkably thorough — code symbols, file paths, FunctionKey enums, routes, feature flag, and all 7 locale files updated in lockstep.

PR #5581, opened 2026-05-08
feat: landing page style update

cuzz-venus:
## Jira ticket(s)

[VPD-1145](https://venus-labs.atlassian.net/browse/VPD-1145)

### Landing — Hero
  - Moved primary `Start now` CTA from inside the Highest APY card to below the Venus TVL number, linking
  to Venus Core Market (sized to Figma: `py-2 px-20`).
  - Highest APY card now hidden below `xl` (1280px); desktop only.

  ### NavBar — Markets / Others menus
  - Desktop: Markets / Others now open on **hover** (added an opt-in `triggerOnHover` prop to the shared
  `Dropdown`; other consumers unchanged).
  - Mobile hamburger: Markets is **expanded by default** (new `defaultOpenOnMobile` flag on `SubMenu`, set
   only on Markets).

  ### Markets page
  - Header subtitle hidden on mobile.
  - Mobile `Sort by` select hidden — added `disableCardsSorting` to `MarketTable`, which derives a
  `cardColumns` array with `sortRows` stripped. `TableCards` then computes empty options and skips the
  select. Desktop column sorting and the other 4 `MarketTable` callers are unaffected.

[VPD-1145]: https://venus-labs.atlassian.net/browse/VPD-1145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/evm/src/containers/Layout/NavBar/MenuItem/index.tsx
```diff
@@ -97,19 +92,18 @@ export const MenuItem: React.FC<MenuItemProps> = ({ item, onClick }) => {
...
             )}
             type="button"
-            onClick={handleToggleDropdown}
           >
             <span>{item.label}</span>
 
-            <Icon name="chevronDown" className={cn('size-3', isSubMenuOpen && 'rotate-180')} />
+            <Icon name="chevronDown" className={cn('size-3', isDropdownOpened && 'rotate-180')} />
           </button>
         )}
```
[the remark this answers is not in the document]
cuzz-venus: I think keyboard-only desktop users (≥ 1280px, no mouse, no touchscreen) don't exist in DeFi 😢  — wallet flows are inherently mouse/touch driven, so this regression isn't worth fixing.

cuzz-venus:
@greptile review again

apps/evm/src/components/Dropdown/index.tsx
```diff
@@ -53,7 +58,8 @@ export const Dropdown = ({
             <div className="relative z-50 hidden min-w-full md:block">
               <div
                 className={cn(
-                  'border-lightGrey bg-cards absolute top-2 min-w-full overflow-hidden border shadow',
+                  'border-lightGrey bg-cards absolute min-w-full overflow-hidden border shadow',
+                  triggerOnHover ? 'top-0' : 'top-2',
```
therealemjy: I wouldn't expect the `triggerOnHover` property to change the UI (only to change the trigger mechanism to open the dropdown menu).

apps/evm/src/containers/MarketTable/index.tsx
```diff
@@ -44,6 +44,7 @@ export interface MarketTableProps
   };
   marketType?: 'supply' | 'borrow';
   className?: string;
+  disableCardsSorting?: boolean;
```
therealemjy: Actually, this is broader change than that: on mobile, the cards aren't sorted at all (not conditionally).
So you can remove this prop and simply remove the sorting on mobile.

therealemjy (commented):
The alignments in the top landing page section are a bit off compared to the designs on desktop.

Current:
Image

Designs:
Image

apps/evm/src/components/Dropdown/index.tsx
cuzz-venus: Thanks for the catch — you're right. I had tweaked top-* to avoid the cursor crossing the gap from triggering a premature close mid-hover. I've now switched to a different approach: instead of changing the UI, I added a small (~200ms) close delay on mouseLeave, so the cursor has time to cross the gap before the menu actually closes. triggerOnHover no longer touches any styling — it only controls the open/close trigger.

apps/evm/src/containers/MarketTable/index.tsx
cuzz-venus: Done — removed the prop and use css to simply hidden on mobile.

apps/evm/src/components/Dropdown/index.tsx
cuzz-venus: have align the figma

Currently
image

Figma
image

apps/evm/src/components/Dropdown/index.tsx
```diff
@@ -21,13 +21,40 @@ export const Dropdown = ({
...
+    hoverCloseTimerRef.current = setTimeout(() => setIsDropdownOpened(false), 200);
+  };
+  useEffect(
+    () => () => {
+      if (hoverCloseTimerRef.current) {
+        clearTimeout(hoverCloseTimerRef.current);
+      }
+    },
+    [],
+  );
```
therealemjy: This could be brittle I think (if a user moves their cursor between the button and the menu then it will hide after a little while).

I took the liberty to add a commit to this PR to show another way of doing this purely with CSS: https://github.com/VenusProtocol/venus-protocol-interface/pull/5581/changes/a9d2eb95fa0d93dfa72177ff412198b9bfe96b50

Have a look and let me know what you think.

apps/evm/src/utilities/getCombinedDistributionApys/index.ts
```diff
@@ -53,18 +53,33 @@ const aggregatePercentages = ({ distributions }: AggregatePercentagesInput) =>
...
+  const borrowPrimeBoostPercentage = usePrimeMax
+    ? BigNumber.maximum(borrow.apyPrimePercentage, borrow.apyPrimeSimulationPercentage)
+    : borrow.apyPrimePercentage;
+
+  const totalSupplyApyBoostPercentage = supply.apyRewardsPercentage.plus(
+    supplyPrimeBoostPercentage,
+  );
+  const totalBorrowApyBoostPercentage = borrow.apyRewardsPercentage.plus(
+    borrowPrimeBoostPercentage,
+  );
```
therealemjy: Since this is only used on the landing page, I wouldn't update the `getCombinedDistributionApys` utility function but I'd just update the logic in the landing page. Down the line I think the case is too specific, and somewhat awkward (users who are Prime and receiving interests will see an APY that's based on hypothetical values that do not correspond to their account).

apps/evm/src/components/Dropdown/index.tsx
cuzz-venus: Makes sense — I had a look at the solution, The CSS bridge is cleaner than the JS timeout, no timing dependency.  Cleaned up the timer!

Thanks!

apps/evm/src/utilities/getCombinedDistributionApys/index.ts
cuzz-venus: Good catch - updated on https://github.com/VenusProtocol/venus-protocol-interface/pull/5581/changes/391b9daf3dbdb6230fcea460e50c636f930a5a43

therealemjy (approved):

PR #5585, opened 2026-05-13
chore: update yarn and set the minimal age for dependencies

gleiser-oliveira:
## Changes

- Lockfile regenerated, thus leading to a big diff. This PR aims to make it even harder to accidentally install compromised packages, by enforcing this rule: https://yarnpkg.com/configuration/yarnrc#npmMinimalAgeGate
- For this I updated yarn to 4.14.1, set CI/CD installations to immutable and made a few fixes in existing dependencies:
  - @venusprotocol/venus-protocol: use the slim version, only installs ABIs and addresses
  - hono: resolve to 4.12.17, because the latest version is less than a week old. Can be dropped once the maturity is reached.
  - A few inconsistencies between dev packages and postinstall scripts


therealemjy (approved):

cuzz-venus (approved):
lgtm

PR #5594, opened 2026-05-19
fix: issue with borrowable state on Trade page

therealemjy:
## Changes

- fix issue with borrowable state on Trade page


cuzz-venus (approved):
/lgtm

PR #5601, opened 2026-05-21
fix: swap quote slippage

therealemjy:
## Changes

- fix an issue where the slippage would be incorrectly applied to swap quotes coming from the API

### DIFF
--- a/.changeset/rotten-ways-love.md
+++ b/.changeset/rotten-ways-love.md
@@ -1,0 +1,5 @@
+---
+"@venusprotocol/evm": minor
+---
+
+fix an issue with swap quote slippage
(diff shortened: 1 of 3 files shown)


apps/evm/src/clients/api/queries/getSwapQuote/formatSwapQuote/index.ts
```diff
@@ -38,6 +38,8 @@ export const formatSwapQuote = ({
       slippagePercentage: -slippagePercentage,
     });
 
+    console.log(slippagePercentage);
```
gleiser-oliveira: Stray log

therealemjy:
@greptile

apps/evm/src/clients/api/queries/getSwapQuote/formatSwapQuote/index.ts
therealemjy: Yeah just saw that, thanks!

gleiser-oliveira (approved):

PR #5605, opened 2026-05-27
feat: vai error show

cuzz-venus:
## Jira ticket(s)

[VPD-1256](https://venus-labs.atlassian.net/browse/VPD-1256)

## Changes

- remove description

before: 
image

after:

image

[VPD-1256]: https://venus-labs.atlassian.net/browse/VPD-1256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/evm/src/pages/Vai/Borrow/index.tsx
+++ b/apps/evm/src/pages/Vai/Borrow/index.tsx
@@ -13,7 +13,6 @@ import {
   AcknowledgementToggle,
   Delimiter,
   LabeledInlineContent,
-  NoticeError,
   NoticeWarning,
   Spinner,
 } from 'components';
@@ -45,7 +44,7 @@ import { useSimulateBalanceMutations } from 'hooks/useSimulateBalanceMutations';
 import type { BalanceMutation } from 'types';
 import TEST_IDS from './testIds';
 import type { FormValues } from './types';
-import { ErrorCode, useForm } from './useForm';
+import { useForm } from './useForm';
 
 export const Borrow: React.FC = () => {
   const { t, Trans } = useTranslation();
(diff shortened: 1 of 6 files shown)


cuzz-venus:
@greptile review again

apps/evm/src/pages/Vai/Borrow/index.tsx
```diff
@@ -266,6 +266,7 @@ export const Borrow: React.FC = () => {
           name="amountTokens"
           rules={{ required: true }}
           disabled={!isUserConnected || isUserMissingPrimeToken}
+          hideErrorDescription
```
therealemjy: I'd recommend taking a different approach.

This form is behaving differently in the sense that we use a custom UI to show errors. The designs were made quite a while ago and so I think we'd be better off staying consistent with the newest form.

So the approach I propose is to update the `useForm` hook of this component so that the error that's set on the field isn't an error code but rather the full error message. This means the `RhfTokenTextField` component won't need to be updated and will natively display the human-friendly error message.

cuzz-venus: Makes sense, fully agree — going with the consistent form pattern is cleaner. I've updated the useForm hooks for both Repay and Borrow to set the translated message directly on the field, and removed the custom NoticeError UI in the parent components. RhfTokenTextField is back to untouched and now natively shows the human-friendly message. 

One small note — I kept ErrorCode.REQUIRES_RISK_ACKNOWLEDGEMENT as-is on the acknowledgeRisk field since it's only used as an internal state marker (never displayed to the user, and there's no matching i18n key for it). Let me know if you'd prefer to clean that up too. 😄

cuzz-venus:
@greptile review again

apps/evm/src/pages/Vai/Borrow/index.tsx
cuzz-venus: image

therealemjy (approved):
Nice

apps/evm/src/pages/Vai/Borrow/index.tsx
therealemjy: That's perfect.

Good point about `ErrorCode.REQUIRES_RISK_ACKNOWLEDGEMENT`.

PR #5612, opened 2026-06-03
feat: update withdraw from matured Pendle vault flow

therealemjy:
## Jira ticket(s)

VEN-1322

## Changes

- update withdraw from matured Pendle vault flow

Do not merge until the Pendle vault has matured and this update has been QA tested.


apps/evm/src/containers/VaultCard/PendleVaultModal/PositionTab/WithdrawForm/index.tsx
```diff
@@ -156,9 +176,11 @@ export const WithdrawForm: React.FC<WithdrawFormProps> = ({ vault, onClose }) =>
         fromTokenPriceCents={fromTokenPriceCents.toNumber()}
         swapQuote={swapQuote}
         swapQuoteError={swapQuoteError ?? undefined}
-        swapFromToken={actionMode !== 'redeemAtMaturity' ? fromToken : undefined}
-        swapToToken={actionMode !== 'redeemAtMaturity' ? toToken : undefined}
+        swapFromToken={!hasMatured ? fromToken : undefined}
+        swapToToken={!hasMatured ? toToken : undefined}
         isLoading={isGetSwapQuoteLoading}
+        disableFromAmountInput={hasMatured}
+        delegateeAddress={pendlePtVaultAddress}
```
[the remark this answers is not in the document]
therealemjy: This is expected and fixes an issue in production where one can't withdraw from the vault if they have not previously delegated.

therealemjy:
@greptile

apps/evm/src/clients/api/mutations/useWithdrawFromPendleVault/index.ts
```diff
@@ -31,38 +33,81 @@ export const useWithdrawFromPendleVault = (
...
+            orderType: Number(order.orderType),
+            makingAmount: BigInt(order.makingAmount),
+            lnImpliedRate: BigInt(order.lnImpliedRate),
+            failSafeRate: BigInt(order.failSafeRate),
+          },
+          signature,
+          makingAmount: BigInt(makingAmount),
+        }));
+
+      const formatWithdrawParams = ({
```
cuzz-venus: Curious why we moved formatWithdrawParams from its own file into the hook closure here — is it because it no longer has other consumers, or was there a typing reason?

A few concerns with the current shape: it conflicts with our "one export per file" convention, the helpers get re-created on every fn() call, and we lose the standalone unit-testability the original file had. 

Happy to discuss if there's context I'm missing 🙏

cuzz-venus (approved):
LGTM (except the conflicts)

apps/evm/src/clients/api/mutations/useWithdrawFromPendleVault/index.ts
therealemjy: Yes you're right, let me update this.

PR #5622, opened 2026-06-09
feat: countdown &timer i18n

cuzz-venus:
## Jira ticket(s)

[VPD-1331](https://venus-labs.atlassian.net/browse/VPD-1331)

## Changes

- add count down
- update timer

[VPD-1331]: https://venus-labs.atlassian.net/browse/VPD-1331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## implementation 
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,7 @@ Thumbs.db
 
 # AI development artifacts
 .ui-develop
+
+# local file
+*.local
+*.local.*
--- a/.gitignore
+++ b/.gitignore
@@ -46,7 +46,3 @@ Thumbs.db
 
 # AI development artifacts
 .ui-develop
-
-# local file
-*.local
-*.local.*
--- a/apps/evm/src/libs/translations/translations/en.json
+++ b/apps/evm/src/libs/translations/translations/en.json
@@ -1253,7 +1253,7 @@
     "description": "Stake XVS into the governance vault and receive Prime rewards on top of your yields.",
     "endOfCycle": {
       "days": "Days",
-      "deadline": "{{date, MMMM d, yyyy HH:mm a}}",
+      "deadline": "{{date, MMMM d, yyyy h:mm a}}",
       "helper": "Stay in the top 500 until <bold>{{deadline}}</bold> to receive Prime rewards during the next cycle. Check the <summaryLink>See last cycle's Prime summary</summaryLink> now to view your ranking and stay competitive.",
       "helperEnded": "Stay in the top 500 until {{deadline}} to receive Prime rewards during the next cycle.",
       "hours": "Hours",
--- a/apps/evm/src/libs/translations/translations/en.json
+++ b/apps/evm/src/libs/translations/translations/en.json
@@ -1251,6 +1251,17 @@
   },
   "primeLeaderboard": {
     "description": "Stake XVS into the governance vault and receive Prime rewards on top of your yields.",
+    "endOfCycle": {
+      "days": "Days",
+      "deadline": "{{date, MMMM d, yyyy HH:mm a}}",
+      "helper": "Stay in the top 500 until <bold>{{deadline}}</bold> to receive Prime rewards during the next cycle. Check the <summaryLink>See last cycle's Prime summary</summaryLink> now to view your ranking and stay competitive.",
+      "helperEnded": "Stay in the top 500 until {{deadline}} to receive Prime rewards during the next cycle.",
+      "hours": "Hours",
+      "mins": "Mins",
+      "sec": "Sec",
+      "title": "End of cycle"
+    },
+    "tablesRefreshNote": "Refreshed hourly · Last refresh: {{distance}} ago",
     "title": "Prime leaderboard"
   },
   "primeStatusBanner": {
--- a/apps/evm/src/libs/translations/translations/en.json
+++ b/apps/evm/src/libs/translations/translations/en.json
@@ -1254,14 +1254,14 @@
     "endOfCycle": {
       "days": "Days",
       "deadline": "{{date, MMMM d, yyyy h:mm a}}",
-      "helper": "Stay in the top 500 until <bold>{{deadline}}</bold> to receive Prime rewards during the next cycle. Check the <summaryLink>See last cycle's Prime summary</summaryLink> now to view your ranking and stay competitive.",
+      "helper": "Stay in the top 500 until <bold>{{deadline}}</bold> to receive Prime rewards during the next cycle. <summaryLink>See last cycle's Prime summary</summaryLink>.",
       "helperEnded": "Stay in the top 500 until {{deadline}} to receive Prime rewards during the next cycle.",
       "hours": "Hours",
       "mins": "Mins",
       "sec": "Sec",
       "title": "End of cycle"
     },
-    "tablesRefreshNote": "Refreshed hourly · Last refresh: {{distance}} ago",
+    "tablesRefreshNote": "Refreshed hourly · Last refresh: {{date, distanceToNow}} ago",
     "title": "Prime leaderboard"
   },
   "primeStatusBanner": {
--- a/apps/evm/src/pages/PrimeLeaderboard/EndOfCycle/Timer/index.tsx
+++ b/apps/evm/src/pages/PrimeLeaderboard/EndOfCycle/Timer/index.tsx
@@ -1,0 +1,45 @@
+import { Fragment } from 'react';
+
+import { useTranslation } from 'libs/translations';
+
+export interface TimerProps {
+  days: number;
+  hours: number;
+  minutes: number;
+  seconds: number;
+}
+
+export const Timer: React.FC<TimerProps> = ({ days, hours, minutes, seconds }) => {
+  const { t } = useTranslation();
+
+  const pad = (value: number) => String(value).padStart(2, '0');
+
+  const segments = [
+    { label: t('primeLeaderboard.endOfCycle.days'), value: days },
+    { label: t('primeLeaderboard.endOfCycle.hours'), value: hours },
+    { label: t('primeLeaderboard.endOfCycle.mins'), value: minutes },
+    { label: t('primeLeaderboard.endOfCycle.sec'), value: seconds },
+  ];
+
+  return (
+    <div className="flex w-full flex-col items-center">
+      <div className="flex w-full items-center justify-between text-center text-b2s text-light-grey">
+        {segments.map(segment => (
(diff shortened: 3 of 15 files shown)


apps/evm/src/pages/PrimeLeaderboard/index.tsx
```diff
@@ -8,43 +11,53 @@ import { RewardTable } from './RewardTable';
...
-const PrimeLeaderboard: React.FC = () => (
-  <Page>
-    <div
-      aria-hidden
-      className="pointer-events-none absolute left-1/2 top-0 w-screen -translate-x-1/2"
-    >
-      <div className="absolute inset-x-0 -top-40 flex justify-center">
-        <div className="h-200 w-200 rounded-full bg-blue/30 blur-3xl" />
+// TODO: use the cycle end date returned by the API
+const endOfCycleDate = endOfMonth(new Date());
```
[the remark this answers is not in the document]
cuzz-venus: This is a placeholder — the cycle end date will come from the API and be read at fetch/render time, replacing this module-level computation. Left a TODO to that effect.

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -1246,6 +1246,17 @@
   },
   "primeLeaderboard": {
     "description": "Stake XVS into the governance vault and receive Prime rewards on top of your yields.",
+    "endOfCycle": {
```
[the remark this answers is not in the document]
cuzz-venus: Fixed — switched to 12-hour with a meridiem (h:mm a) across all locales, so it now reads e.g. "March 31, 2026 11:59 PM" (localized AM/PM per locale) instead of the invalid 24h + meridiem combo.

apps/evm/src/pages/PrimeLeaderboard/EndOfCycle/index.tsx
```diff
@@ -1,18 +1,65 @@
...
+  hours: number;
+  minutes: number;
+  seconds: number;
+  completed: boolean;
+}
+
+export const EndOfCycle: React.FC<EndOfCycleProps> = ({ endDate, className }) => {
+  const { t, Trans } = useTranslation();
+
+  const deadline = endDate ? t('primeLeaderboard.endOfCycle.deadline', { date: endDate }) : '';
```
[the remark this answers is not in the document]
cuzz-venus: endDate is always provided in the product (the page derives it, and it'll ultimately come from the API), so the blank-deadline case won't occur — the optional path is just a defensive  fallback. I'll tighten the type when wiring it to the API.

apps/evm/src/pages/PrimeLeaderboard/EndOfCycle/index.tsx
```diff
@@ -1,18 +1,65 @@
...
+        </p>
+      ) : (
+        <p className="text-center text-b1r text-light-grey">
+          <Trans
+            i18nKey="primeLeaderboard.endOfCycle.helper"
+            values={{ deadline }}
+            components={{
+              bold: <span className="text-b1s text-white" />,
+              summaryLink: <span className="text-blue underline" />,
+            }}
```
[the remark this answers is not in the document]
cuzz-venus: Not a navigation link — "See last cycle's Prime summary" opens the last-cycle summary modal (PRD §6.5), not a page, so an <a> / router Link would be the wrong semantic here. It's an intentional placeholder rendered as inline text for now (kept as a <span> so the copy can wrap mid-phrase to match the design). It'll be wired up as an accessible trigger in the summary-modal ticket.

cuzz-venus:
@greptile review again

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -1247,25 +1246,36 @@
...
+      "days": "Days",
+      "deadline": "{{date, MMMM d, yyyy h:mm a}}",
+      "helper": "Stay in the top 500 until <bold>{{deadline}}</bold> to receive Prime rewards during the next cycle. Check the <summaryLink>See last cycle's Prime summary</summaryLink> now to view your ranking and stay competitive.",
```
therealemjy: The content in the Figma document is incorrect (I left a comment about this but it has not been updated):


```suggestion
      "helper": "Stay in the top 500 until <bold>{{deadline}}</bold> to receive Prime rewards during the next cycle. <summaryLink>See last cycle's Prime summary</summaryLink>.",
```

apps/evm/src/pages/PrimeLeaderboard/EndOfCycle/Timer/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+    { label: t('primeLeaderboard.endOfCycle.hours'), value: hours },
+    { label: t('primeLeaderboard.endOfCycle.mins'), value: minutes },
+    { label: t('primeLeaderboard.endOfCycle.sec'), value: seconds },
+  ];
+
+  return (
+    <div className="flex w-full flex-col items-center">
+      <div className="flex w-full items-center justify-between text-center text-b2s text-light-grey">
+        {segments.map(segment => (
+          <p key={segment.label} className="w-9">
```
therealemjy, line 28: You'll need to add a `key` prop (perhaps using the `label` property of each segment) to each element for React to differentiate them when rendering the DOM.

apps/evm/src/pages/PrimeLeaderboard/EndOfCycle/Timer/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  days: number;
+  hours: number;
+  minutes: number;
+  seconds: number;
+}
+
+export const Timer: React.FC<TimerProps> = ({ days, hours, minutes, seconds }) => {
+  const { t } = useTranslation();
+
+  const pad = (value: number) => String(value).padStart(2, '0');
```
therealemjy: This utility function should be inside its own file and directory (inside the Timer directory).

apps/evm/src/pages/PrimeLeaderboard/EndOfCycle/index.tsx
```diff
@@ -1,18 +1,65 @@
 import { cn } from '@venusprotocol/ui';
+import ReactCountdown from 'react-countdown';
 
 import { Card } from 'components';
+import { useTranslation } from 'libs/translations';
+
+import { Timer } from './Timer';
 
 export interface EndOfCycleProps {
+  endDate?: Date;
```
therealemjy: I think `endDate` should be required, so that we don't end up with an odd state where the timer displays 00:00:00:00 and the description underneath is incomplete ("...in the top 500 until **(empty)** to receive").

If an end date hasn't been fetched from the page yet, then this page can either decide to show a spinner or simply not display this component.

apps/evm/src/libs/translations/translations/en.json
```diff
@@ -1247,25 +1246,36 @@
...
+      "helperEnded": "Stay in the top 500 until {{deadline}} to receive Prime rewards during the next cycle.",
+      "hours": "Hours",
+      "mins": "Mins",
+      "sec": "Sec",
+      "title": "End of cycle"
+    },
+    "tablesRefreshNote": "Refreshed hourly · Last refresh: {{distance}} ago",
```
therealemjy: The codebase has a built-in way of displaying distances in time:
```suggestion
    "tablesRefreshNote": "Refreshed hourly · Last refresh: {{date, distanceToNow}} ago",
```

You might have to check how it handles past dates, but the idea is that the date will be converted into a distance in time from now.

.gitignore
```diff
@@ -46,3 +46,7 @@ Thumbs.db
 
 # AI development artifacts
 .ui-develop
+
+# local file
+*.local
+*.local.*
```
therealemjy: I think these can be removed, I imagine they are created by the LLM you use locally but as far as I know we don't specify in our AGENTS.md file that it should create these directories when working.

apps/evm/src/libs/translations/translations/en.json
cuzz-venus: Good suggestion! uodate!

apps/evm/src/pages/PrimeLeaderboard/EndOfCycle/Timer/index.tsx
cuzz-venus: Nice catch. already have the key prop.

apps/evm/src/pages/PrimeLeaderboard/EndOfCycle/Timer/index.tsx
cuzz-venus: Good point!

apps/evm/src/pages/PrimeLeaderboard/EndOfCycle/index.tsx
cuzz-venus: Yes. have fixed this one, and add spinner for this

apps/evm/src/libs/translations/translations/en.json
cuzz-venus: Update!

.gitignore
cuzz-venus: Yes, used for save PRD.md, but no need now, I have remove this one

apps/evm/src/pages/PrimeLeaderboard/EndOfCycle/index.tsx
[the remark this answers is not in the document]
therealemjy: @cuzz-venus Your point is fair, though I agree with greptile that the CTA will need to use a `button` tag (and to open the last cycle's modal when clicked).

therealemjy:
The PR looks good to me. The only work left on it is for the "See last cycle's summary" button to open the last cycle's summary modal.

apps/evm/src/pages/PrimeLeaderboard/EndOfCycle/index.tsx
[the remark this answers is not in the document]
cuzz-venus: Hey @therealemjy, you are right, actually I have updated the span to the button at https://github.com/VenusProtocol/venus-protocol-interface/pull/5630/changes#diff-4df7a2b008795f28a1f81865b97c60a1efa765a95984d98cd6adb22af9878030R55

forgot to update here. but now I have updated in this branch :)

therealemjy (approved):

PR #5625, opened 2026-06-10
feat: add geo-block

therealemjy:
## Jira ticket(s)

VPD-1352

## Changes

- add geo-block
- list NVDAB, TSLAB and SPCB on testnet


apps/evm/src/clients/api/queries/useGetPools/getPools/index.ts
```diff
@@ -85,8 +90,50 @@ export const getPools = async ({
...
+  if (vaiPriceMantissaResult.status === 'rejected') {
+    throw vaiPriceMantissaResult.reason;
+  }
+
+  if (vaiRepayRateMantissaResult.status === 'rejected') {
+    throw vaiRepayRateMantissaResult.reason;
+  }
+
+  if (ipLocationResult.status === 'rejected') {
+    logError(ipLocationResult.reason.data?.exception);
```
cuzz-venus: Out of curiosity — since the IP check is done from the browser, if stops a user from blocking the request in DevTools? In that case ipCountryCode would be undefined and all restrictions would be skipped.

therealemjy: That's correct yes.
The only way we could really prevent users from being able to interact with restricted markets is by adding a blocker at the contract level. Any frontend blocker can be removed by modifying the client.

therealemjy:
@greptile

apps/evm/src/clients/api/queries/useGetPools/getPools/index.ts
cuzz-venus: Thanks! make sense!

therealemjy:
@greptile your first assumption is incorrect. The `isGated` and `isRestricted` properties of each asset within pools will be `true` if freeapi.com is unavailable, which is an acceptable drawback.
It is true that assets will be ungated and unrestricted while whiling for freeapi.com to respond, but that's an acceptable drawback too.

therealemjy:
@greptile

therealemjy:
@greptile

cuzz-venus (approved):
LGTM

therealemjy:
@greptile

apps/evm/src/clients/api/queries/useGetPools/index.ts
```diff
@@ -102,4 +108,33 @@ export const useGetPools = (input?: TrimmedInput, options?: Options) => {
...
+      pools: applyCountryCodeToPools({
+        countryCode: getIpLocationData?.countryCode,
+        pools: poolsQuery.data.pools,
+        tokenMetadataMapping: poolsQuery.data.tokenMetadataMapping,
+      }),
+    };
+  }, [getIpLocationData?.countryCode, poolsQuery.data]);
+
+  return {
+    ...poolsQuery,
```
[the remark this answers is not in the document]
therealemjy: This is accepted yes.

therealemjy:
@greptile

PR #5639, opened 2026-06-16
chore: upgrade Vite to v8 and enable React Compiler

therealemjy:
## Jira ticket(s)

VPD-1406

## Changes

- upgrade Vite to v8 and enable React Compiler


cuzz-venus (approved):
PR looks good to me.

PR #5641, opened 2026-06-17
feat: support abi & api

cuzz-venus:
## Jira ticket(s)

[VPD-1337](https://venus-labs.atlassian.net/browse/VPD-1337)

## Changes

- support contract abi
- support api 

[VPD-1337]: https://venus-labs.atlassian.net/browse/VPD-1337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/evm/src/clients/api/queries/getPrimeCurrentCycle/index.ts
```diff
@@ -0,0 +1,89 @@
...
+          startsAt: new Date(payload.cycle.startsAt),
+          endsAt: new Date(payload.cycle.endsAt),
+        }
+      : null,
+    pendingPool: payload.pendingPool
+      ? {
+          ...payload.pendingPool,
+          computedAt: new Date(payload.pendingPool.computedAt),
+        }
+      : null,
```
therealemjy: Let's not use `null` in the frontend repo, but rather `undefined`. `null` has many many issues: https://craftbettersoftware.com/p/stop-using-null-its-a-bad-practice

apps/evm/src/clients/api/queries/getPrimeCurrentCycle/index.ts
```diff
@@ -0,0 +1,89 @@
+import { VError } from 'libs/errors';
+import type { ChainId } from 'types';
+import { restService } from 'utilities';
+
+export interface PrimeCurrentCycle {
+  cycleIndex: number;
+  status: string;
```
therealemjy: This should be an enum so we can list all the possible statuses.

apps/evm/src/clients/api/queries/getPrimeCurrentCycle/index.ts
```diff
@@ -0,0 +1,89 @@
...
+  endsAt: Date;
+  anchorBlockNum: string | null;
+  mintLimitUsed: number;
+}
+
+export interface PrimePendingRewardPool {
+  blockNumber: string;
+  computedAt: Date;
+  primeHolderCount: number;
+  totalPendingUsdCents: string;
```
therealemjy: You don't need to add "Usd" into the name (the repo considers that dollars are the the only on-crypto currency used).

apps/evm/src/clients/api/queries/getPrimeCycles/index.ts
```diff
@@ -0,0 +1,79 @@
...
+import { restService } from 'utilities';
+
+export interface PrimeFinalizedCycle {
+  cycleIndex: number;
+  startsAt: Date;
+  endsAt: Date;
+  mintLimitUsed: number;
+  totalRewardPoolUsdCents: string | null;
+  finalizedAt: Date | null;
+}
```
therealemjy: The type for defining cycles is defined in multiple places.

I recommend creating a shared type that can live in side the root `types` file (apps/evm/src/types/index.ts) and that contains all the optional properties to define the various states of a cycle (past, current, etc..).

apps/evm/src/clients/api/queries/getPrimeCurrentCycle/index.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export interface GetPrimeCurrentCycleOutput {
+  cycle: PrimeCurrentCycle | null;
+  pendingPool: PrimePendingRewardPool | null;
+}
+
+interface PrimeCurrentCycleResponse {
+  cycleIndex: number;
+  status: string;
```
therealemjy: Same thing here, let's use an enum.

apps/evm/src/clients/api/queries/getPrimeCurrentCycle/index.ts
```diff
@@ -0,0 +1,89 @@
+import { VError } from 'libs/errors';
+import type { ChainId } from 'types';
+import { restService } from 'utilities';
+
+export interface PrimeCurrentCycle {
+  cycleIndex: number;
+  status: string;
+  startsAt: Date;
+  endsAt: Date;
+  anchorBlockNum: string | null;
```
therealemjy: No `null`

apps/evm/src/clients/api/queries/getPrimeCycles/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  return {
+    page: payload.page,
+    limit: payload.limit,
+    total: payload.total,
+    cycles: payload.result.map(cycle => ({
+      ...cycle,
+      startsAt: new Date(cycle.startsAt),
+      endsAt: new Date(cycle.endsAt),
+      finalizedAt: cycle.finalizedAt ? new Date(cycle.finalizedAt) : null,
```
therealemjy: No `null`

apps/evm/src/clients/api/queries/getPrimeEffectiveStake/useGetPrimeEffectiveStake.ts
```diff
@@ -0,0 +1,53 @@
...
+  return useQuery({
+    queryKey: [FunctionKey.GET_PRIME_EFFECTIVE_STAKE, { accountAddress, chainId }],
+    queryFn: () =>
+      callOrThrow({ accountAddress, primeLeaderboardContractAddress }, params =>
+        getPrimeEffectiveStake({ publicClient, ...params }),
+      ),
+    ...options,
+    enabled:
+      (options?.enabled === undefined || options?.enabled) &&
+      isPrimeLeaderboardEnabled &&
```
therealemjy: I'd say you don't need to check that the feature is enabled from the hook directly: we anyway only enable the page if the Prime leaderboard feature is enabled, so any code underneath can just assume it will only run if it is allowed to.

apps/evm/src/clients/api/queries/getPrimeLeaderboard/useGetPrimeLeaderboard.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const useGetPrimeLeaderboard = (input: TrimmedInput = {}, options?: Partial<Options>) => {
+  const { chainId } = useChainId();
+  const params = { ...input, chainId };
+
+  return useQuery({
+    queryKey: [FunctionKey.GET_PRIME_LEADERBOARD, params],
+    queryFn: () => getPrimeLeaderboard(params),
+    placeholderData: keepPreviousData,
+    ...options,
```
therealemjy, line 34: We might want to refresh this query quite frequently by default. I'd pass a `refetchInterval` property generated via the `generatePseudoRandomRefetchInterval` function.

apps/evm/src/clients/api/queries/getPrimeOnChainPendingRewards/index.ts
therealemjy, line 1: Rather than creating a separate query hook, I think we should update the existing `useGetPendingRewards` hook so that it uses Prime V2 (it currently uses Prime V1): https://github.com/VenusProtocol/venus-protocol-interface/blob/68b31bb938894abe416a6938564e7198825354d3/apps/evm/src/clients/api/queries/getPendingRewards

apps/evm/src/clients/api/queries/getPrimePastCycle/index.ts
```diff
@@ -0,0 +1,102 @@
...
+import type { Address } from 'viem';
+
+export interface PrimePastCycle {
+  cycleIndex: number;
+  status: string;
+  startsAt: Date;
+  endsAt: Date;
+  mintLimitUsed: number;
+  totalRewardPoolUsdCents: string | null;
+  finalizedAt: Date | null;
```
therealemjy: No `null`

apps/evm/src/clients/api/queries/getPrimePastCycle/index.ts
```diff
@@ -0,0 +1,102 @@
...
+  finalTotalStakedMantissa: string;
+}
+
+export interface GetPrimePastCycleInput {
+  chainId: ChainId;
+  cycleIndex: number | 'latest';
+}
+
+export interface GetPrimePastCycleOutput {
+  cycle: PrimePastCycle | null;
```
therealemjy: No `null`

apps/evm/src/clients/api/queries/getPrimePastCycle/index.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  return {
+    ...cycle,
+    startsAt: new Date(cycle.startsAt),
+    endsAt: new Date(cycle.endsAt),
+    finalizedAt: cycle.finalizedAt ? new Date(cycle.finalizedAt) : null,
+  };
+};
+
+export const getPrimePastCycle = async ({
```
therealemjy: For simplicity and to follow the same naming logic as `getPrimeUserCycleRewards`, I'd name this function `getPrimeCycle` (and propagate the change onto the related types and hook).

apps/evm/src/clients/api/queries/getPrimeUserCycleRewards/useGetPrimeUserCycleRewards.ts
```diff
@@ -0,0 +1,45 @@
...
+    queryKey: [
+      FunctionKey.GET_PRIME_USER_CYCLE_REWARDS,
+      { chainId, cycleIndex, accountAddress: accountAddress as Address },
+    ],
+    queryFn: () =>
+      callOrThrow({ accountAddress }, params =>
+        getPrimeUserCycleRewards({ chainId, cycleIndex, ...params }),
+      ),
+    ...options,
+    enabled: (options?.enabled === undefined || options?.enabled) && !!accountAddress,
```
therealemjy, line 48: Since we are fetching past data, we don't need to refetch it every 5 minutes (which is what React Query will do by default). You can pass the following options to prevent the query from being refetched this session"
```tsx
    refetchOnMount: false,
    refetchOnReconnect: false,
    refetchOnWindowFocus: false,
    staleTime: Number.POSITIVE_INFINITY,
    gcTime: Number.POSITIVE_INFINITY,
```

apps/evm/src/clients/api/queries/getPrimePastCycle/useGetPrimePastCycle.ts
```diff
@@ -0,0 +1,27 @@
...
+>;
+
+export const useGetPrimePastCycle = ({ cycleIndex }: TrimmedInput, options?: Partial<Options>) => {
+  const { chainId } = useChainId();
+  const params = { cycleIndex, chainId };
+
+  return useQuery({
+    queryKey: [FunctionKey.GET_PRIME_PAST_CYCLE, params],
+    queryFn: () => getPrimePastCycle(params),
+    ...options,
```
therealemjy: Since we are fetching past data, we don't need to refetch it every 5 minutes (which is what React Query will do by default). You can pass the following options to prevent the query from being refetched this session"
```tsx
    refetchOnMount: false,
    refetchOnReconnect: false,
    refetchOnWindowFocus: false,
    staleTime: Number.POSITIVE_INFINITY,
    gcTime: Number.POSITIVE_INFINITY,
```

apps/evm/src/containers/PrimeRank/EligibilityStatus/index.tsx
therealemjy: To follow the codebase practice of adding a suffix with the unit, I'd rename this: `TOP_500_GAP_THRESHOLD_XVS_TOKENS`.

apps/evm/src/clients/api/queries/getPrimeCurrentCycle/index.ts
cuzz-venus: Done - make sense use undefined instead of null

apps/evm/src/clients/api/queries/getPrimeCurrentCycle/index.ts
cuzz-venus: Done — added a PrimeCycleStatus enum (active / finalized, the values the BE returns)

apps/evm/src/clients/api/queries/getPrimeCurrentCycle/index.ts
cuzz-venus: Done — domain fields drop Usd;

apps/evm/src/clients/api/queries/getPrimeCycles/index.ts
cuzz-venus: Done — added a shared PrimeCycle type in types/index.ts with the common fields required and the state-specific ones `optional (status?, anchorBlockNum?, totalRewardPoolCents?, finalizedAt?),` and replaced the three local types

apps/evm/src/clients/api/queries/getPrimeCurrentCycle/index.ts
cuzz-venus: Done — same enum applied to the response status.

apps/evm/src/clients/api/queries/getPrimeCurrentCycle/index.ts
cuzz-venus: update!

apps/evm/src/clients/api/queries/getPrimeCycles/index.ts
cuzz-venus: update!

apps/evm/src/clients/api/queries/getPrimeEffectiveStake/useGetPrimeEffectiveStake.ts
cuzz-venus: Done — removed the useIsFeatureEnabled check; it now only gates on accountAddress.

apps/evm/src/clients/api/queries/getPrimeLeaderboard/useGetPrimeLeaderboard.ts
cuzz-venus: Done — added refetchInterval via generatePseudoRandomRefetchInterval('fast').

apps/evm/src/clients/api/queries/getPrimeOnChainPendingRewards/index.ts
cuzz-venus: Done — migrated `useGetPendingRewards` to Prime V2 and removed the separate `getPrimeOnChainPendingRewards` hook.

I deliberately kept the exact same call as V1 — `getPendingRewards` (+ `paused`) via `simulateContract`, just pointed at the PrimeV2 contract/ABI — so the behavior stays identical to what was already QA'd: it accrues then returns the up-to-date pending. The V2 return shape matches V1 (vToken / rewardToken / amount), so the downstream formatting is unchanged.

apps/evm/src/clients/api/queries/getPrimePastCycle/index.ts
cuzz-venus: Done

apps/evm/src/clients/api/queries/getPrimePastCycle/index.ts
cuzz-venus: Done

apps/evm/src/clients/api/queries/getPrimePastCycle/index.ts
cuzz-venus: Done — renamed and propagated to the types, hook, query key (GET_PRIME_CYCLE) and exports

apps/evm/src/clients/api/queries/getPrimePastCycle/useGetPrimePastCycle.ts
cuzz-venus: Done — added refetchOnMount/Reconnect/WindowFocus: false and staleTime/gcTime: Infinity so it isn't refetched during the session.

apps/evm/src/containers/PrimeRank/EligibilityStatus/index.tsx
cuzz-venus: Good point

cuzz-venus:
@greptile review again

therealemjy (approved):

cuzz-venus:
@greptile review again

cuzz-venus:
@greptile review again

cuzz-venus:
@greptile review again

cuzz-venus:
@greptile review again

PR #5647, opened 2026-06-17
fix: display swap quote missing error

therealemjy:
## Changes

- display "No swap found" error on Repay with collateral form


cuzz-venus (approved):
LGTM

PR #5649, opened 2026-06-17
fix: prevent Trade page from fetching unnecessary swap quote

therealemjy:
## Changes

- prevent Trade page from fetching unnecessary swap quote


cuzz-venus (approved):
LGTM

PR #5676, opened 2026-07-07
refactor: slippage tolerance form

therealemjy:
## Changes

- refactor slippage tolerance form to use React Hook form and only allow values of at least 0.01%


cuzz-venus (approved):
looks good to me

PR #5681, opened 2026-07-10
feat: only show Prime banner on XVS vault when user is connected

therealemjy:
## Changes

- only show Prime banner on XVS vault when user is connected


cuzz-venus (approved):
LGTM

PR #5682, opened 2026-07-10
refactor: move market operation form to containers

therealemjy:
## Changes

- move market operation form to containers
- fix translation keys of some components


cuzz-venus (approved):
LGTM

PR #5690, opened 2026-07-16
fix: improve modal stacking of AccountModal

therealemjy:
## Changes

- prevent language selector from making its parent scrollable on desktop


cuzz-venus (approved):
LGTM

PR #5700, opened 2026-07-19
ci: add AI command

therealemjy:
## Changes

- add a yarn ai command for launching Pi’s interactive AI client with repository-defined skills
- introduce a `sync-translations` skill that updates locales from the English source while preserving interpolation and markup


cuzz-venus (approved):
LGTM

PR #5702, opened 2026-07-20
feat: update LiquidityHub object to support token distributions

therealemjy:
## Jira ticket(s)

VPD-1642

## Changes

- add token distribution support to Liquidity Hubs and their mock data
- make `Apy` and `ApyBreakdown` generic across assets and Liquidity Hubs, with shared APY calculations and an asset adapter
- display distribution-inclusive APYs in Liquidity Hub forms and tables, including consistent table sorting


cuzz-venus (approved):
LGTM

PR #5704, opened 2026-07-20
ci: simplify AGENTS.md

therealemjy:
## Changes

This PR is based on the findings from [this paper](https://arxiv.org/abs/2602.11988), which show that context files do not provide a great codebase overview to LLMs and increase the thinking time as well as the cost of answering to prompts.

The `AGENTS.md` is kept as a short list of essential rules to follow when writing code for this repository, rather than a complete overview of the codebase.


cuzz-venus (approved):
LGTM

PR #5728, opened 2026-07-23
fix: heap memory allocation issue

therealemjy:
## Changes

- fix heap memory allocation issue
- use slim versions of Venus contract packages as we only need addresses from them


cuzz-venus (approved):
LGTM

PR #5732, opened 2026-07-25
refactor: start refactoring Emotion CSS components

therealemjy:
## Changes

This PR serves as a proof of concept to refactor components using Emotion CSS and Style Components using LLMs so that they use Tailwind CSS instead. A Playwright config is added for LLMs to render the components and ensure the UI is exactly the same after each refactor. Other components will be refactored in subsequent PRs.

- add Playwright config to evm app
- add stories to components missing them
- refactor `Table` component to use Tailwind CSS classes instead of Emotion CSS


cuzz-venus (approved):
LGTM

PR #5759, opened 2026-08-01
fix: various fixes

therealemjy:
## Changes

 - Fixed E-Mode group sorting so blocked groups are correctly pushed after enableable groups
 - Added unit tests covering E-Mode sorting behavior
 - Updated pending rewards query key to include isolated pool comptroller addresses and fixed the XVS vault pool count query enablement logic
 - Removed duplicate API exports
 - Updated the format script to use the workspace Biome dependency and added Biome as a dev dependency

### DIFF
--- a/.changeset/early-results-talk.md
+++ b/.changeset/early-results-talk.md
@@ -1,0 +1,5 @@
+---
+"@venusprotocol/evm": patch
+---
+
+various fixes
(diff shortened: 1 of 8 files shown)


therealemjy:
@greptile

cuzz-venus (approved):
LGTM

PR #5764, opened 2026-08-04
feat: fallback to default chain ID

therealemjy:
## Changes

- fallback to default chain ID when URL param is incorrect


cuzz-venus (approved):
LGTM

PR #5772, opened 2026-08-06
feat: add category for markets

cuzz-venus:
## Jira ticket(s)

[VPD-1695](https://venus-labs.atlassian.net/browse/VPD-1695)

## Changes

- support select dropdown
- support category filter

[VPD-1695]: https://venus-labs.atlassian.net/browse/VPD-1695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/evm/src/containers/MarketTable/useControls/index.tsx
```diff
@@ -12,7 +13,15 @@ export const useControls = ({
...
   const [searchValue, onSearchValueChange] = useState('');
+  const [selectedCategories, onSelectedCategoriesChange] = useState<string[]>([]);
   const [userChainSettings] = useUserChainSettings();
+  const { chainId } = useChainId();
+
+  const [prevChainId, setPrevChainId] = useState(chainId);
+  if (chainId !== prevChainId) {
+    setPrevChainId(chainId);
+    onSelectedCategoriesChange([]);
+  }
```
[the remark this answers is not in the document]
cuzz-venus: make sense

apps/evm/src/types/index.ts
```diff
@@ -233,12 +234,19 @@ export interface PoolVai {
...
+  order: number;
+}
+
 export interface Pool {
   comptrollerAddress: Address;
   name: string;
   isIsolated: boolean;
   assets: Asset[];
   eModeGroups: EModeGroup[];
+  categories?: MarketCategory[];
```
therealemjy: I think you can simplify the implementation by not adding this property to the `Pool` object type.
The `categories` property of a pool is basically an array of all the categories its markets are in. Since the markets are using plain strings as their categories, I'd work with that.

On the frontend, the only time you really need the full list of categories is in the `MarketTable` component. It can infer the list of categories by simply going through all markets and generating an array of strings. The `MarketCategory` type can then be removed completely (you can differentiate and sort the categories alphabetically by their names directly).

apps/evm/src/containers/Controls/index.tsx
```diff
@@ -33,14 +40,31 @@ export const Controls: React.FC<ControlsProps> = ({
...
+            placeholder={searchInputPlaceholder}
+            leftIconSrc="magnifier"
+            className="sm:grow @2xl:max-w-75"
+          />
+
+          {!!categories?.length && onSelectedCategoriesChange && (
+            <MultiSelect
+              options={[...categories]
+                .sort((a, b) => a.order - b.order)
+                .map(category => ({ value: category.tag, label: category.label }))}
```
therealemjy: I think this could be made a bit clearer by sorting the categories beforehand:
```tsx
// Outside the return block:
const sortedCategories = [...(categories ?? [])]
  .sort((a, b) => a.order - b.order)
  .map(category => ({ value: category.tag, label: category.label }))
```

```suggestion
// Inside the return block
          {!!sortedCategories.length && onSelectedCategoriesChange && (
            <MultiSelect
              options={sortedCategories}
```

therealemjy (commented):
I think this feature should be placed under a feature flag, as I imagine

apps/evm/src/types/index.ts
cuzz-venus: Done — Pool.categories and ApiPool.categories are gone, and MarketTable now derives the list by walking the markets, as you suggested.

One deviation: instead of working with the plain category string, I switched to market.marketCategory, which the API was already returning and we simply weren't using — it's the same { tag, label, order } object. That keeps MarketCategory around, but it means we don't lose the display names (ETH-derived, bStocks) or the backend's intended ordering (Others last), which sorting plain tags alphabetically would have.

apps/evm/src/containers/Controls/index.tsx
cuzz-venus: good catch! addressed!

apps/evm/src/types/index.ts
therealemjy: That makes sense.

therealemjy (approved):

PR #5790, opened 2026-08-21
feat: add support for vceBTC and Solv venue

therealemjy:
## Changes

- add support for vceBTC and Solv venue


cuzz-venus (approved):
LGTM

PR #5793, opened 2026-08-25
feat: rename the Venus vault Active state to Supply

cuzz-venus:
## Jira ticket(s)

VPD-1896

## Changes

- rename the Venus vault `Active` state to `Supply`: unpaused Venus vaults now report the existing `Deposit` state, whose label already reads "Supply", and `VaultStatus.Active` is removed
- point both Supply badges at a single translation key, since they diverged in `zh-Hans` (`供应` vs `存款`)
- map the legacy `?status=active` filter link onto the Supply state and rewrite it in the URL


cuzz-venus:
@greptile review again

PR #5800, opened 2026-09-01
chore: remove fixed-rate vault banner

cuzz-venus:
## Jira ticket(s)

VPD-1942

## Changes

- Removed the "USDT Fixed-Term Vault Now Live on BNB Chain. Target ≈4.2% APR." promotional banner from the Markets page, now that the Ceffu fixed-rate vault has entered its lock period.
- Reverted the layout scaffolding the banner required, so the Markets page renders through the shared `Layout` wrapper again, and dropped the now-unused `doNotShowFixedRateVaultsAdBanner` user setting and `layout.header.dashboardAdBanner.*` translations.


cuzz-venus:
@greptile review again

PR #5804, opened 2026-09-02
feat: replace the hub contract row with the operator address

cuzz-venus:
## Jira ticket(s)

VPD-1964

## Changes

- Replaced the "Hub contract" row in the Liquidity Hub info section with an "Operator address" row, since the hub contract address is always the vhToken address already shown right below it.
- The operator address is read on-chain per hub and links to the chain explorer, like the other address rows.


cuzz-venus:
@greptile review again

PR #5814, opened 2026-09-14
fix: skip API calls and hide UI for features the connected chain does not support

cuzz-venus:
## Changes

Two features were being requested on chains that do not support them, producing failing API calls.

**Liquidity Hub**

- `useGetLiquidityHubs` now only runs when the Liquidity Hub feature is enabled on the connected chain, so `/liquidity-hub/hubs` is no longer called (and no longer 400s) on chains that do not support it.
- The guard lives in the hook rather than at the call sites, so it cannot be bypassed by a caller passing `enabled: true`. This also covers the four call sites that had no chain guard of their own: `useGetPendingRewards`, `useGetAccountTransactionHistory`, `LiquidityHubForm`, and the `LiquidityHubs` page.
- Removed the now-redundant call-site `enabled` flags on the dashboard.

**Transaction history**

- Restricted the `transactionHistory` feature flag to BNB Chain. It previously also included Ethereum and Unichain, where `/transactions` returns 500.
- This hides the dashboard Transactions tab and stops the API call on those chains, since both the tab and the query were already gated behind this flag.
- Visiting a stale `?tab=transactions` URL on an unsupported chain is safe: `useTabs` falls back to the first tab and rewrites the param.


cuzz-venus:
@greptile review again

